Issue #102, opened 2026-05-22
mcp(http): validate bearer at initialize, not just at tool dispatch


jiashuoz:
## Observed

`https://mcp.e2a.dev/mcp` accepts arbitrary non-empty `Authorization: Bearer …` values during MCP `initialize` and `tools/list`. Surfaced during the independent review of PR #101:

> with the bogus header `Authorization: Bearer e2a_TESTVALUE` the server happily completed `initialize` (HTTP 200, `Mcp-Session-Id` issued) **and** returned the full `tools/list` catalog. Only the *no-header* case correctly returns 401.

So an unauthenticated probe can:
1. Open an MCP session and obtain a `Mcp-Session-Id`
2. List the full tool catalog with schemas
3. Cannot actually call any tool (the backend rejects on the first real dispatch)

## Why this happens

The MCP HTTP server is pure transport. It forwards the bearer to the e2a backend only at tool-call time. `initialize` + `tools/list` are served entirely by the in-process MCP SDK against the per-session `E2AClient` — neither requires a backend round trip, so the bearer's validity is never checked at session open.

## Impact

Low. The tool catalog is already public (same schemas ship on npm), so this leaks no real secrets. But:

- Session resources are allocated for unauthenticated callers — a trivial way to grief the `maxSessions` cap (default 500) without ever auth'ing.
- The 401 / WWW-Authenticate flow that MCP clients depend on for the OAuth re-auth dance never fires for invalid-but-present bearers, so they get a misleadingly "successful" connection followed by 401s only on first real tool call.

## Suggested fix

Validate the bearer once at `initialize` time, before allocating session state. A single `GET /api/v1/agents` (or a cheaper `GET /api/v1/info`-equivalent) against the backend with the bearer would do it — if it returns 401, the MCP server returns 401 to the client immediately with the proper WWW-Authenticate challenge. If it returns 200, proceed with session creation.

Code shape (~30 LOC in `mcp/src/http-server.ts` near the `isInitializeRequest` block at [http-server.ts:197-238](https://github.com/Mnexa-AI/e2a/blob/main/mcp/src/http-server.ts#L197-L238)):

\`\`\`ts
if (!entry && isInitializeRequest(req.body)) {
  // Validate the bearer eagerly so unauthenticated callers get
  // 401 + WWW-Authenticate at session-open instead of allocating
  // a session and surfacing the failure only on first tool call.
[...]

PR #104, opened 2026-05-22
fix(mcp): validate bearer during initialize

Baijack-star:
## Summary

Hi, I am an AI agent helping with the MCP auth/session issue described in #102.

This PR validates the bearer during MCP `initialize` before session allocation, so a bogus non-empty bearer no longer receives a successful initialize response or an `Mcp-Session-Id`.

Changes:
- reuse a `WWW-Authenticate` bearer challenge helper across MCP auth failures
- detect 401-style SDK/backend errors during session client construction
- return `401` + `WWW-Authenticate` for invalid bearers during `initialize`
- keep non-auth `listAgents()` failures best-effort so transient backend hiccups do not block initialize
- add a regression proving invalid bearer initialize does not allocate a session or issue `Mcp-Session-Id`
- update the existing agent-email test to reflect that `listAgents()` now validates the bearer while still avoiding client reconstruction when `agentEmail` is already set

Addresses #102.

## Verification

The local environment routes HTTP through a proxy by default, so MCP loopback tests need `NO_PROXY=127.0.0.1,localhost`.

- `npm_config_cache=/tmp/npm-cache-e2a-102 npm --workspace sdks/typescript run build`
- `NO_PROXY=127.0.0.1,localhost npm_config_cache=/tmp/npm-cache-e2a-102 npm --workspace mcp test -- tests/http.test.ts`
- `NO_PROXY=127.0.0.1,localhost npm_config_cache=/tmp/npm-cache-e2a-102 npm --workspace mcp test`
- `npm_config_cache=/tmp/npm-cache-e2a-102 npm --workspace mcp run build`
- `git diff --check`

Note: the first plain `npm install` attempt failed because the local user npm cache had root-owned files. I reran with `npm_config_cache=/tmp/npm-cache-e2a-102`; no lockfile changes are included.

### DIFF
--- a/mcp/src/http-server.ts
+++ b/mcp/src/http-server.ts
@@ -163,6 +163,17 @@ function resourceMetadataURL(req: Request, opts: HttpServerOptions): string {
   return `${base}/.well-known/oauth-protected-resource`;
 }
 
+function bearerChallenge(req: Request, opts: HttpServerOptions): string {
+  return `Bearer realm="e2a", resource_metadata="${resourceMetadataURL(req, opts)}"`;
+}
+
+class InvalidBearerError extends Error {
+  constructor() {
+    super("invalid bearer token");
+    this.name = "InvalidBearerError";
+  }
+}
+
 async function handleClientRequest(
   req: Request,
   res: Response,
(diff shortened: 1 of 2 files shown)


jiashuoz (commented):
Thanks for the fix — code looks good. Two follow-ups before merge:

1. I've approved CI on the branch; let's confirm it goes green.

2. Could you tighten the `WWW-Authenticate` assertion in the new test? The fix adds `, error="invalid_token"` on the invalid-bearer path (per RFC 6750 §3.1), but the test only matches `/Bearer realm="e2a"/` — same regex as the missing-bearer test. MCP clients may key off the `error` parameter for the re-auth prompt, so it's worth asserting it explicitly:

```ts
expect(res.headers.get("www-authenticate")).toMatch(
  /Bearer realm="e2a".*error="invalid_token"/,
);
```

Also, a small note for the PR description: this change makes `listAgents()` run on every `initialize` even when `E2A_AGENT_EMAIL` is preset (previously zero round-trip in that path). That's intentional and required for the fix — just worth calling out so operators aren't surprised by the added per-session call.

Issue #92, opened 2026-05-19
MCP HTTP server: v0.2 follow-up polish


jiashuoz:
Tracking the items deliberately deferred during v0.2 PR 1 ([#91](https://github.com/Mnexa-AI/e2a/pull/91)) and the second `/review` pass. None are blockers for going live; most either resolve naturally when v0.3 OAuth ships or are operational nits worth doing before the integration page gets real traffic.

Grouped by where they should land.

## Resolved by v0.3 (OAuth) — no separate work needed

- [ ] **Cross-Bearer session access on `GET`/`DELETE /mcp`.** Today a valid Bearer + a known session id is sufficient to subscribe to that session's SSE notifications or terminate it. Session ids are 128-bit UUIDs so this is unguessable in practice, but the proper fix is to bind a session to the OAuth `client_id` at consent time and reject cross-client access. Implementation lives naturally inside v0.3 PR B.
- [ ] **CORS `origin: \"*\"`.** Open for v0.2 because LLM-spawned clients are non-browser. Once we have a real list of remote MCP hosts (Claude Code's UA pattern, Cursor, Cline, etc.), tighten to an allowlist in [mcp/src/http-server.ts:52](mcp/src/http-server.ts#L52).

## Operational polish (any time before launch traffic)

- [ ] **Per-IP rate limiting at Caddy.** The Go backend already rate-limits per agent (60 sends/min). Add a coarse per-IP limit at the Caddy layer in `e2a-ops` to absorb abusive clients before they hit the Go backend's auth path. Cheap insurance.
- [ ] **Structured JSON logging.** Currently the bin writes plain text to stderr (e.g. `e2a-mcp-http listening on :3000 …`). Switch to one-line JSON per event so GCE Cloud Logging parses cleanly. Same change in [mcp/src/bin/http.ts:71-75](mcp/src/bin/http.ts#L71-L75) and the shutdown lines.
- [ ] **\`E2A_BASE_URL\` validation.** [mcp/src/bin/http.ts](mcp/src/bin/http.ts) accepts any string today. A typo like \`htps://e2a.dev\` would surface as a tool-call failure at runtime, not at boot. Validate via \`new URL()\` in \`loadConfig\` and fail with \`ConfigError\` like the other vars.

## Low-severity nits

[...]

PR #198, opened 2026-06-07
fix(mcp): use ASCII hyphens in "no session" error for log-pipeline safety

James-Z-Zhang00:
## Summary

  The JSON-RPC error returned on a `/mcp` request with no session id used a
  U+2014 em-dash (`"no session — send initialize first"`), which can be mangled
  by ASCII-only log pipelines. Replace it with two ASCII hyphens (`--`).

  String-only change — no API shape, endpoint, or contract change.

  Refs #92 (low-severity nits).

  ## Operational risk

  None. Single-character change to the wording of an existing error message; no
  user data, side effects, auth, billing, or external integrations touched.
  Nothing to roll back beyond reverting the commit.

  ## Test plan

  - [x] `npm run build --workspace @e2a/mcp-server` — compiles clean
  - [x] `npm test --workspace @e2a/mcp-server` — 118/118 pass

### DIFF
--- a/mcp/src/http-server.ts
+++ b/mcp/src/http-server.ts
@@ -262,7 +262,7 @@ async function handleClientRequest(
     return;
   }
   if (!entry) {
-    res.status(400).json(jsonRpcError(req.body, -32000, "no session — send initialize first"));
+    res.status(400).json(jsonRpcError(req.body, -32000, "no session -- send initialize first"));
     return;
   }
 


jiashuoz (approved):
Approved. Correct, minimal, well-scoped change. Line 265 was the only `jsonRpcError` call site carrying a non-ASCII character — the other three are already plain ASCII, so this makes the JSON-RPC protocol-error layer fully ASCII-consistent. No contract/behavior change. (Note for #92 follow-up: a few thrown Error/Zod messages in tools/agents.ts, tools/domains.ts, tools/attachments.ts still carry em-dashes if full ASCII hygiene is the eventual goal — out of scope here.)

PR #200, opened 2026-06-08
feat(mcp): emit structured JSON logs from the HTTP server bin (#92)

James-Z-Zhang00:
## Summary

  Closes one item from #92 (MCP HTTP server v0.2 polish): the `e2a-mcp-http` bin emitted operational events as plain text to stderr (`e2a-mcp-http
  listening on :3000 …`, `received SIGTERM, draining...`, etc.). On GCE these are captured by Cloud Logging, which stores plain text as an opaque
  `textPayload` — not filterable by field, and multi-line content (e.g. a fatal stack trace) is split into separate fragmented entries.

  This switches every operational line to a single-line JSON object via a small `logJson(severity, event, message, fields)` helper. Cloud Logging
  parses these into structured `jsonPayload` and honors the `severity` (log level) and `message` (console summary) special fields — so logs stay
  human-skimmable *and* become queryable (filter on `severity`, `event`, `port`, `error`, …). Scope is the HTTP server bin only.

  Example, before → after:

  e2a-mcp-http listening on :3000 (base https://e2a.dev, hosts mcp.e2a.dev)
  ```json
  {"severity":"INFO","event":"listening","message":"e2a-mcp-http listening on
  :3000","port":3000,"baseUrl":"https://e2a.dev","allowedHosts":["mcp.e2a.dev"]}
```

 ## Operational risk

  Low. No data, auth, billing, or external integrations touched — output format only.

  - Log format change: anything that grep'd the old plain-text lines (dashboards, alerts, saved queries) will need updating to match the JSON
  fields. No code consumes these lines; they're operator-facing only.
  - Scope: HTTP server bin (mcp/src/bin/http.ts) only. The stdio server (index.ts/config.ts) is intentionally left as plain text — it runs locally
  inside a user's MCP host and is read by a human at a terminal, where JSON would be a regression.
  - Severity mapping: e2a_base_url_deprecated → WARNING; config_error / drain_timeout / shutdown_error / fatal → ERROR; listening / draining → INFO.
  - Rollback: revert the commit; no migrations or persisted state involved.

 ## Test plan

  - [x] npm run build --workspace @e2a/mcp-server — clean (no tsc errors)
  - [x] npm test --workspace @e2a/mcp-server — 119 passing (8 files)
  - [x] New logJson unit test: asserts a single line (trailing newline, no embedded newlines) parsing to {severity, event, message, ...fields}
  - [x] Tightened the E2A_BASE_URL deprecation test to assert the emitted line parses as JSON with severity:"WARNING" + the event name

### DIFF
--- a/mcp/src/bin/http.ts
+++ b/mcp/src/bin/http.ts
@@ -13,6 +13,26 @@ interface BinConfig {
 
 class ConfigError extends Error {}
 
+type LogSeverity = "INFO" | "WARNING" | "ERROR";
+
+// logJson emits one operational event as a single-line JSON object on
+// stderr. GCE Cloud Logging parses a single-line JSON payload into
+// structured `jsonPayload` fields and honors two special keys: `severity`
+// (sets the entry's log level) and `message` (the human-readable summary
+// shown in the console). Keeping each event on one line also means
+// multi-line content like a stack trace in `error` stays a single log
+// entry instead of being split into several fragmented ones. The result
+// is both human-skimmable (via `message`) and queryable (filter on
+// `severity`, `event`, or any structured field).
+function logJson(
+  severity: LogSeverity,
+  event: string,
+  message: string,
+  fields: Record<string, unknown> = {},
+): void {
+  process.stderr.write(`${JSON.stringify({ severity, event, message, ...fields })}\n`);
+}
+
 function parsePositiveInt(name: string, raw: string, def: number): number {
   if (raw === "") return def;
   const n = Number(raw);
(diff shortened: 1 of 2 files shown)


jiashuoz (approved):
Approved. Clean, well-scoped conversion of the HTTP server bin to structured Cloud Logging JSON.

Highlights:
- Gets the subtle thing right: all logs (incl. INFO `listening`) go to stderr, but each carries a structured `severity` key, so Cloud Logging surfaces them at the correct level instead of defaulting stderr to ERROR.
- `fatal` puts `err.stack` into the `message` field — JSON.stringify escapes newlines, keeping a multi-line stack as one log entry.
- Stable `event` discriminators make the log queryable independent of the human message.
- Conversion is complete within the bin (only stderr write left is inside logJson); startHttpServer emits no logs of its own. Plain-text writes in index.ts/config.ts belong to the stdio binary and are correctly out of scope.
- Tests pin the load-bearing invariants: single-line, no embedded newlines, exact JSON shape, plus the deprecation path.

One non-blocking nit: `{ severity, event, message, ...fields }` spreads fields last, so a future caller passing fields.severity/message/event would silently override the fixed keys. Latent only — no current caller hits it. Worth a guard if the helper spreads to more sites later.

PR #206, opened 2026-06-13
fix(mcp): let create_agent target a verified custom domain

jiashuoz:
## Problem

The MCP `create_agent` tool only exposed `slug`, which the backend forces onto the shared domain (`agents.e2a.dev`, `internal/agent/api.go:851-862`). So through MCP you could `register_domain` → `verify_domain` a custom domain, but then had **no MCP path to create an inbox on it** — the tool dead-ends on its own happy path. Binding `support@agentdrive.run` (a verified custom domain) required dropping to the raw REST API with a CLI key.

This is a pure MCP-layer divergence: the REST contract `RegisterAgentRequest` and the TS SDK's `client.registerAgent` **already accept `email`** for custom-domain agents (`sdks/typescript/src/v1/generated/types.ts`, `client.ts:134-143`). Only the MCP tool's zod schema had drifted by omitting it. **No API or SDK change is needed.**

## Change

- Add optional `email` alongside `slug`; make `slug` optional. `email` = a full address on a custom domain already registered + verified for the account.
- Enforce **exactly-one-of `slug`/`email`** in the handler with a directive error. This is superset-safe: every call it lets through behaves identically on the API; it only rejects the ambiguous both/neither case earlier, with a clearer message than the API's bare `400 "email is required"` / silent slug-precedence.
- Retitle/redescribe the tool — it no longer advertises itself as shared-domain-only, which is the framing that hid the capability.

## Tests

`mcp/tests/tools.test.ts`: custom-domain `email` forwarding, plus both-given and neither-given rejection. Full suite: **122 passed**. `tsc` build clean.

## Not included (tracked separately)

This fixes the capability gap. Broader items surfaced in the same review — the "via e2a" sending model, MCP-vs-REST credential split, `verify_domain` flagging SPF that custom domains don't need, and a contract test asserting MCP request bodies validate against the API schema (the durable anti-drift guard) — are left for follow-ups.

### DIFF
--- a/mcp/src/tools/agents.ts
+++ b/mcp/src/tools/agents.ts
@@ -57,20 +57,31 @@ export function registerAgentTools(server: McpServer, client: E2AClient): void {
   server.registerTool(
     "create_agent",
     {
-      title: "Create a new agent inbox on the shared domain",
+      title: "Create a new agent inbox (shared or custom domain)",
       description:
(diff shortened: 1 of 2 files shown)


jiashuoz:
Superseded by #253 (merged). The §6a MCP re-curation reshaped `create_agent` to take a full `email` on any verified owned domain (or the shared domain) and removed `slug`/`agent_mode`/`webhook_url`. This PR's goal — create_agent targeting a verified custom domain — is already on `main`, and its diff patches the now-removed slug/agent_mode code (would conflict with the rewritten agents.ts). Closing as stale.

PR #208, opened 2026-06-14
feat(api): v1 contract on Huma+chi — full additive /v1 surface (Slice 1)

jiashuoz:
Implements the **additive `/v1` contract** of the [e2a API v1 redesign](docs/design/api-v1-redesign.md) (Slice 1) on **Huma + chi**, with the OpenAPI 3.1 spec generated from typed handlers (drift-proof), a standardized error envelope, cursor pagination, `X-Request-Id`, and per-write idempotency. Mounted as a **strangler** in front of the legacy `gorilla/mux` surface: chi owns `/v1` and falls back to `/api/v1` for anything not yet ported — so this PR is **non-breaking** (legacy `/api/v1` is untouched and fully functional).

## What's here — the entire bearer surface on `/v1` (~34 operations)
- **meta**: `GET /v1/info`
- **agents**: list · get · create (domain-ownership guard + limits→402) · update · delete · test-send
- **messages**: list (cursor + filters) · get · **send · reply · forward** (idempotent)
- **conversations**: list · get
- **domains**: list · get · register · update · delete · verify (DNS probe, 412)
- **webhooks**: create · list · get · update (cooldown→409) · delete · rotate-secret · test · deliveries
- **events**: list · get · redeliver (idempotent)
- **account**: limits · export (GDPR) · delete (`?confirm=DELETE`)

## How the risky bits were done safely
The live email/HITL/idempotency path was **not** re-implemented. The send orchestration was *extracted* from `agent.API` into HTTP-free cores — `DeliverOutbound`, `HoldForApprovalCore`, `SendTestCore`, plus `ExportUserDataCore`/`DeleteUserDataCore` — which **both** the legacy handlers and `/v1` now call, so the live path exists exactly once. Idempotency is a reusable Huma helper (`runIdempotent`) replacing the legacy `capturingWriter`. Security-sensitive validators (SSRF, event allowlist, IDN, recipient parsing, references, self-alias stripping) are reused from their canonical homes, not duplicated.

## Verification
- `go test ./internal/httpapi/` green (per-endpoint positive + negative paths over the real chi+Huma transport).
- **Full `go test ./internal/agent/` green on Postgres** after each core extraction — the existing send/HITL/self-send/idempotency/user-data-rights suites prove zero behavior change.
- **Live-path e2e** (real binary + Mailpit): `POST /v1/send` → 200 → email delivered; **idempotency replay** (same key+body) → cached response, **no second email**; same key + different body → 422; clean server logs.
- CI green on the additive surface.

## Deliberately a follow-up PR (not here)
The **destructive 1Z cutover** is split out to keep this large additive change reviewable and non-breaking: removing the legacy `/api/v1` bearer routes, retargeting the existing `agent` test suite from `/api/v1`→`/v1`, the cross-cutting rate-limit + `RateLimit-*` headers, the contract-drift CI gate, SDK regen, and the host/config cutover.


jiashuoz:
## Independent + adversarial review — dispositions

Ran two independent read-only review passes (a quality/fidelity reviewer and an adversarial "try to break it" reviewer). They **converged on a single real blocker**, now fixed in `ec3455b`.

**BLOCKER (fixed)** — `runIdempotent` orphaned the idempotency key when `json.Marshal` of the response failed *after* the side effect committed (neither Complete nor Release), risking a 5-min DOS and a double-send after the stale window. Fix: always `Complete` (lock) the key on the success path, with a `{}` fallback body on marshal failure; never `Release` post-commit (that would permit re-execution). Regression test added (`TestIdempotentMarshalFailureStillCompletes`).

**Verified as non-issues:**
- *test-send doesn't meter usage* — matches the legacy `handleSendTestEmail` (it never called `RecordAndCheck` either); Slice 1 preserves behavior. The send-cap pre-check (`CheckMessageSend`) is applied.
- *missing unowned-domain create test* — already covered by `TestCreateAgentUnregisteredDomain`.
- *unused index in `assertAgentsOwned`* — `i` is used (`agents[i].ID`).

**Confirmed safe by the adversarial pass (REFUTED exploit classes):** tenant isolation (every per-agent op gates on `resolveOwnedAgent`; inbound reply/forward + webhook `agent_ids` filters are ownership-checked), CRLF/SSRF injection (subject/recipient/conversation_id + `ValidateWebhookURL` on both create and the effective PATCH state), strangler route collision (Huma routes registered before the legacy `NotFound` fallback), HITL double-notify on retry (held 202 is cached → replay, no re-hold), and fail-open/nil-deref (all Deps nil-guarded, no 200-on-error).

PR #227, opened 2026-06-18
feat(sdk/ts): namespaced E2AClient over generated /v1 base; retire legacy surface (3.0)

jiashuoz:
## Slice 8b — TypeScript consumer port (SDK)

Replaces the hand-written-on-types `/api/v1` SDK with a **namespaced ergonomic layer** that wraps the generated `oag/` `/v1` client (composition, not inheritance). Bumps `@e2a/sdk` to **3.0.0** (SemVer major; breaking surface change).

### What's in
- **`E2AClient` with resource sub-clients**: `agents`, `messages`, `conversations`, `domains`, `events`, `webhooks`, `account` (+ `account.suppressions`), plus `info()` and `listen(address)`. Per-agent methods take an explicit `address` (no inference).
- **Typed error hierarchy** (`E2AError` + subclasses) mapped from `ApiException` via the response envelope (`code`/`message`/`requestId`/`status`/`retryable`).
- **Retry + idempotency** (from 8b-1) and **AutoPager** wired into every write/list.
- **Webhook verification**: `constructEvent(rawBody, header, secret)` → typed `WebhookEvent` (verify + parse in one call), plus multi-secret verification for rotation/multi-endpoint handlers.
- **WS**: repoint `/api/v1/agents/{email}/ws` → `/v1/agents/{address}/ws`; `client.listen(address)` returns the hybrid `WSStream`.
- **Retire** `api.ts` + `inbound-email.ts` (legacy shared-secret/MIME path) and the legacy `generated/types` re-export; public types now come from the `oag` models. Drop the unused `mailparser` dep.

### Bug found & fixed (with regression test)
The generated layer unconditionally sets an **empty** `Idempotency-Key` header on `send`/`reply`/`forward`/`approve` when the caller passes none — which made the retry layer treat the key as "caller-supplied" and skip minting. Fixed to mint on present-but-empty; regression test added in `retry.test.ts`. Caught by the new `client.test.ts`.

### Verification
- `tsc --noEmit` clean; `npm test --workspace @e2a/sdk` → **69 passing**; `npm run build` clean.
- **Over-the-wire e2e** against a freshly-built `./bin/e2a` (fresh Postgres + Mailpit): 12/12 — auth→`E2AAuthError`, `info()`, agents create/list/get, `messages.list` cursor encoding, **idempotent send replay produced exactly one outbound message**, webhook create→`constructEvent` round-trip, envelope→typed-error mapping. Server log clean (no 5xx/panic). Real-server finding folded in: `agents.get` of an unknown address returns **403** (existence-hiding) → `E2APermissionError`, not 404.

### Deferred (tracked)
- **Rich `InboundEmail`** ergonomic object (typed accessors / object-level `.reply()`) depends on typed event payloads (S2) — deferred to a post-S2 fast-follow. `constructEvent` + raw `messages.get()` cover webhook + REST today.
- **cli/mcp repoint**: `@e2a/cli` and `@e2a/mcp-server` still consume the retired surface and **will be red on this PR's `cli`/`mcp` CI jobs** — repointed in the immediate follow-up PR (per agreed sequencing).
- Root README cross-SDK section + Python SDK (8c) updated when the Python port lands.


jiashuoz:
## Review disposition (independent + adversarial, run in parallel)

**Security-critical paths — all REFUTED (fail-closed):** webhook signature verification (empty/malformed/non-hex/length-collision headers, empty-secret-array, replay, verify-before-parse), idempotency mint-once/reuse (incl. the present-but-empty generated stub), retry bounds/abort/deadline, and error-envelope mapping (string/null/array bodies) were all probed and hold.

**Findings fixed in `e602598` (each with a regression test):**
- _should-fix_ — README `send` example used the wire field `html_body`; corrected to `htmlBody`.
- _should-fix_ — `webhooks.deliveries()` surfaced the response `next_cursor` on a cursorless endpoint, which would re-fetch the same page and trip the cycle guard. Now single-page by contract. Test added.
- _low (converged)_ — pagination cycle guard only caught 1–2-step cycles; multi-step cycles (A-B-C-A) now caught via a bounded seen-set. Test added.
- _nice-to-have_ — `WSListener` default baseUrl aligned to `https://api.e2a.dev`.
- _coverage_ — added tests for the connection-error path (→ `E2AConnectionError`) and account/suppressions passthroughs.

**Deferred (with reason):**
- WSStream unbounded buffer + un-encoded `?token=` — **pre-existing** (this PR only touched doc/path lines in `ws.ts`); tracked separately.
- `reject`/`updateLabels` can't take a caller idempotency key — acceptable (auto-mint still applies); noted.

Unit suite now 73 passing; tsc + build clean; over-the-wire e2e against a real `./bin/e2a` remains 12/12.

PR #235, opened 2026-06-18
Retire the legacy /api/v1 surface + flatten the SDK oag/ subpath

jiashuoz:
Finishes the v1 cutover (Slice 8e): deletes the legacy `/api/v1` strangler now that every consumer is on `/v1`, and un-nests the generated SDK base. Three commits — server retire, SDK flatten, docs.

## 1. Delete the legacy `/api/v1` surface (`feat(api)!`)
Consumer-verified — nothing in the repo calls any of these:
- **Deleted (covered by `/v1` or unused):** legacy WS alias `/api/v1/agents/{email}/ws` (real WS is `/v1/agents/{address}/ws`); label-update `PATCH`; flat outbound detail `GET /api/v1/messages/{id}`; account-wide `GET /api/v1/pending` (web + CLI already build the cross-agent pending view from `GET /v1/agents` + per-agent message lists).
- **Dropped (no `/v1` equivalent, no consumer):** account-level `/api/v1/users/me/signing-secrets` (superseded by per-webhook `whsec_` / `rotate-secret`); bulk `POST /api/v1/webhooks/{id}/redeliver-since` (`/v1` has per-event redeliver). Shared replay helpers (`loadEventForReplay`/`insertReplayDelivery`) preserved — `/v1` uses them.
- **Moved to `/v1`:** HITL email magic-link pages → `/v1/approve`, `/v1/reject` (raw token-gated HTML via the chi root's fallback to the mux); notifier link + form action emit the new path.

**Gap the deletion exposed (fixed here):** the legacy mux handlers were the *only* code emitting the RFC 6750 `WWW-Authenticate` challenge on 401. The `/v1` surface now advertises the Bearer scheme from one place — an `authChallenge` middleware reusing `agent.API`'s challenge builder. Its `challengeWriter` forwards `http.Hijacker` so the `/v1` WebSocket upgrade still works (caught + regression-tested).

## 2. Flatten the SDK `oag/` subpath (`refactor(sdk)`)
`src/v1/oag/` → `src/v1/generated/` in both SDKs (reusing the slot freed by deleting dead legacy codegen). Generate scripts / `.openapi-generator-ignore` / hand-written imports updated so regen stays reproducible. Deleted the dead `openapi-typescript` output, the Python `datamodel-codegen` module, and the swag→TS npm scripts. `make swagger`/`web/public/openapi.yaml` kept (the dashboard's `scalar.html` renders it) but unwired from `make generate`. The regen also picked up the #231 MessageView trust fields the base was missing.

## Verification
- Full Go suite green (one pre-existing, unrelated flake: `TestSubscriberStore_RecordAttemptFailureClimbsToFailed`, fails identically on `main`).
- Contract suite green — real server, **real WebSocket dials through the full chi + middleware stack** (exercises the Hijack fix) + real auth.
- New over-the-wire tests: every retired route → 404; `/v1` 401 advertises Bearer; `challengeWriter` Hijack + 401-only header unit tests.
- `spec-check` + `generate-sdk-check` clean; SDK/CLI/MCP build + tests green.
- `api/openapi.yaml` unchanged (no deleted route was a Huma operation).


jiashuoz:
**Post-review status (both review passes triaged):**
- Adversarial review caught a real break CI missed — `internal/e2e` (behind the `integration` tag) called the dropped `redeliver-since`; removed those 4 tests, full e2e suite now green. Fixed 2 cosmetic nits.
- **Signing-secrets:** confirmed as a deliberate deprecation (management API dropped; per-user signing stays live in the relay with deployment-secret fallback). Documented in design §13.3.
- **Required pre-deploy follow-up (separate PR):** migrate `tests/e2e-prod` (~243 `/api/v1` refs) to `/v1` — it works against prod via the legacy fallback today but breaks once this deploys. Tracked in design §13.3.

PR #239, opened 2026-06-19
docs(design): reconcile v1 doc with as-built (drop `from`, defer `reply_to`, F4 shipped)

jiashuoz:
Reconciles `docs/design/api-v1-redesign.md` with the shipped `/v1` surface, prompted by reviewing whether dropping the send `from` param made sense.

## What changed in the doc (doc-only PR)

- **Decision 3 (authoritative body-shape spec):** added an as-built note — the proposed `from`/`reply_to` outbound body fields did **not** ship.
  - **`from` dropped** (PR #238): the sender is structurally the path agent (`POST /v1/agents/{address}/messages`) and the server hardcodes `From = agent.address`, so a body `from` is redundant at best and an address-spoofing vector at worst. Confirmed by the #238 adversarial review. `from` survives only as a **response** field (observed sender on message views).
  - **`reply_to` deferred, not rejected:** it sets the Reply-To header, is independent of the sender, and is genuinely useful — an open gap to add to send/reply/forward later.
- **Status banner:** F4 (uniform `AutoPager`) marked **shipped** (`conversations.list` now returns an `AutoPager`, PR #238); recorded the `from`/`reply_to` outcome.
- **Goals, §6a MCP target table (`send_message`/`reply_to_message` rows), §13 change inventory:** brought in line with the above.

## Verified consistency
`from`/`reply_to` are absent from `SendEmailRequest`/`ReplyRequest`/`ForwardRequest` in `api/openapi.yaml` and both SDK generated bases; the only remaining `from`/`reply_to` are response-side fields on `Message`/`MessageSummaryView`/`MessageView`.

### DIFF
--- a/docs/design/api-v1-redesign.md
+++ b/docs/design/api-v1-redesign.md
@@ -19,13 +19,16 @@
 > management API were **dropped**; `send` ships **nested** at
 > `POST /v1/agents/{address}/messages` (no top-level `/v1/send`); HITL ships as
 > the two routes **`…/approve` + `…/reject`** (not a single `approval`
-> sub-resource); and decisions 4/9/10 **shipped**. **Still pending:** the MCP
-> tool-surface re-curation to the §6a target (the SDK *transport* was repointed
-> onto `/v1`, but the tool names/args in `mcp/src/tools/` remain the pre-redesign
-> set — tracked as a separate round) and F4 (uniform `AutoPager` —
-> `conversations.list` still returns a plain array). The event name
-> `email.rejected` shipped as-is (decision 9's `email.approval_rejected` rename
-> did not land).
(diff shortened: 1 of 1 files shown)


jiashuoz:
Redundant. The reconcile here (`from` dropped, `reply_to` deferred, F4/AutoPager shipped) is already on `main` via #237/#238/#250. It also still asserts claims that are now false post-#253 ("MCP tool-surface re-curation still pending", "email.approval_rejected rename did not land") — both shipped. Merging would add nothing and re-introduce stale framing. Closing as stale. (Note: `main`'s intro banner + §13 note still carry those stale 'pending' lines — tracked as a separate doc-cleanup, not this PR.)

PR #240, opened 2026-06-19
GA launch-review blockers: freeze a consistent, forward-compatible /v1

jiashuoz:
Final pre-GA hardening from the launch review. Every finding was **adversarially verified against the code before fixing** — one alleged blocker (a dual-identifier URL mismatch) was *refuted* on second look and dropped.

### Contract / wire format (the irreversible stuff)
- **Webhook envelope discriminator unified on `type`** (was `event`). ⭐ This was a genuine end-to-end break no prior review caught: the server delivered `{"event": …}` but the SDK's own `construct_event`/`constructEvent` helper *requires* `type` and rejects anything else — so the recommended one-call webhook helper **could not parse a single real delivery**. Now a delivery body is the same shape as a `GET /v1/events` object. Adds a wire-key contract test.
- **List endpoints → shared `Page[T]{items, next_cursor}` envelope** (`agents`, `domains`, `webhooks`, `suppressions`) instead of bespoke `{agents:[]}` wrappers — so pagination can be added later without a breaking shape change. `next_cursor` is null at launch.
- **`rotate-secret` honors `Idempotency-Key`** — a retried rotate replays the first secret instead of minting a second and silently dropping the one the caller already stored.

### SDK correctness
- **TS retry gating per-operation**: only reads, HTTP-idempotent writes (except account deletion), and server-deduped POSTs (send/reply/forward/approve) are retried — no more double-create / re-fired delete on a transient failure. Matches the Python SDK.
- **Python tolerates unknown enum values** (strip `*_validate_enum` validators in `generate-oag.sh`) so a new server enum value doesn't crash deployed clients. Matches TS passthrough.

### Security
- **Webhook SSRF**: re-validate the resolved IP at dial time (`net.Dialer.Control`), closing the DNS-rebinding window between registration validation and delivery. Prod-gated so loopback CI/dev still works.

### Docs / release
- README, `docs/api.md`, and the `adk-cloud-webhook` example use the real 3.0 SDK surface (`construct_event` + namespaced `client.messages.*`); example pin → `e2a>=3`.
- `VERSION` 0.2.0 → **1.0.0** to match the `/v1` GA line and `openapi info.version`.

### Tests / CI
- **Per-package coverage gate** (`.testcoverage.yml` + `make cover`/`cover-check` + a Postgres-backed `go-coverage` CI job) for `webhook`/`webhookpub`/`httpapi`. Floors set just under current (60/60/73) to lock coverage and prevent regression.
- Fixed a **stale webhook test** that hard-coded `max_attempts=5` after migration 027 bumped the column default to 8 (it was red on `main` locally; CI's `go` job has no DB so it never ran).

Regenerated `api/openapi.yaml` and both SDK bases (verified deterministic).

## Verification
- TS SDK: 84/84 · Python SDK: 131 pass / 14 skipped (contract) · Go `httpapi`/`webhookpub`/`webhook`/`agent` green under `-p 1` · spec golden + generate-sdk drift gates pass · coverage gate passes with teeth (verified it fails on an impossible floor).


jiashuoz:
## Independent review pass — results

Ran 4 parallel read-only reviewers (Go/API, security, SDK/codegen, adversarial) against this diff. Three approved their domains; the adversarial + security reviewers caught **2 real blockers the original pass missed**, now fixed in a09a55c:

- **B1 (blocker):** web dashboard read old list keys (`body.webhooks` / `data.domains` / `agentsResp.agents`) from `/v1/*` → pages would render empty under the new `items` envelope. Web test mocks were masking it. Fixed + mocks updated.
- **B2 (blocker):** `internal/e2e` still read the old webhook envelope key (`env.Event` compile error / `Envelope["event"]` nil) after the `event→type` rename. Fixed. (CI's `go` job runs without `-tags integration`, so it never compiled these.)
- **M1 (major):** rotate-secret retry parity — Python now retries+keys it (`_write_idempotent`) to match TS, since the endpoint is server-deduped. Comment corrected.

### Verified SAFE by reviewers (no action)
SSRF dial guard (Control fires per resolved IP, IPv4-mapped/CGNAT caught, pool-reuse not a bypass, redirects blocked); rotate-secret idempotency secret-caching (within existing plaintext-at-rest posture); Python enum-strip (only `*_validate_enum` removed, models compile, idempotent); TS retry gate (Idempotency-Key emitted on exactly the 5 server-deduped POSTs); pagination AutoPagers (no infinite loop on null cursor); codegen determinism.

### Non-blocking follow-ups (not gating this PR)
1. **SSRF predicate — NAT64/6to4 (LOW):** `64:ff9b::/96` and `2002::/16` embed an IPv4 that isn't re-checked; passes `IsDisallowedWebhookIP`. Exploitable only on a host actually running a NAT64/6to4 translator to internal v4 (not default in standard VPCs). Add: extract + re-check the embedded v4.
2. **Self-host LAN webhooks (LOW):** the prod SSRF guard blocks RFC-1918 targets with no opt-out. Pre-existing at registration (`ValidateWebhookURL` already blocked private IPs); the dial-time block just enforces it at send. If on-prem LAN webhook delivery is a supported self-host topology, add a config escape hatch.
3. **rotate-secret idempotency cache TTL (LOW):** the new secret is cached in `idempotency_keys` (plaintext, 24h) on replay — same posture as the existing plaintext `webhooks.signing_secret`, but consider a shorter TTL or omitting the secret from the cached body.
4. **Coverage floors are thin (~2.5pp margin)** on webhook/webhookpub and DB-dependent — minor future-flake risk; ratchet up as tests grow.

PR #268, opened 2026-06-22
feat: agent protection config sub-resource (beta) — closes #13/#21

jiashuoz:
Relocates the per-agent screening config from 12 flat fields on `AgentView`/`UpdateAgentRequest` to an account-scoped **`/v1/agents/{email}/protection`** sub-resource (beta). Design: `docs/design/2026-06-22-agent-protection-config.md`.

## Why
- **#13 (threshold leak):** `AgentView` exposed raw scan thresholds to agent-scoped creds → an injected agent could calibrate evasion. Now the whole resource is account-only, and `AgentView` slims to identity+status.
- **#21 (MCP unconfigurable):** `update_agent` advertised retired `hitl_enabled`/`hitl_mode` (silent no-op). Now a dedicated `get_protection`/`update_protection` MCP pair (admin tier).

## Shape
```
GET /v1/agents/{email}/protection   (account scope, beta)
PUT /v1/agents/{email}/protection   (account scope, full replace, beta)
```
Nested `inbound`/`outbound` × `{gate{policy,allowlist,action}, scan{sensitivity}}` + shared `holds`. Trust gate is the monotonic `open|allowlist|domain` ladder (no `verified_only`). Scan is a semantic `off|low|medium|high` sensitivity; the handler derives the engine's float thresholds (engine untouched). PUT requires all 3 top-level keys (D3).

## Decisions
- D1: `outbound.gate.scope` containment **deferred** (additive later).
- D2: `updateAgent` reduced to **name-only**.
- D3: PUT **required top-level keys**, default-filled leaves.
- O4: additive `migration 045` (`*_scan_sensitivity` columns); engine reads derived floats.

## Slices
1. store layer + migration 045 (DB-backed tests)
2-3. handler + view + AgentView slim + name-only PATCH + spec regen
4. SDK regen (TS+Py) + ergonomic `getProtection`/`replaceProtection`
5. MCP `get_protection`/`update_protection` (admin tier) + name-only `update_agent`
6. contract scenarios + design reconcile

## Verification
- Full Go suite green (unit + integration + e2e + contract).
- TS SDK (97) + MCP (134) tests green; spec + SDK + MCP-tier drift gates green.
- **Over-the-wire e2e** against the real contract-server binary + Postgres: GET defaults, PUT 200, GET confirms persistence, missing-key → 422, `AgentView` carries no thresholds (#13), and the DB shows `high` → scan on / 0.3 / 0.8 (derived engine columns).
- Agent-scope 403 on `/protection` covered through the real chi+Huma HTTP stack (scope_test) — not re-run against the standalone binary (no agent key minted in the manual run).

## Deferred (beta umbrella, all additive)
Web dashboard UI (stabilization order: API/SDK/MCP first), `gate.require_auth` (verified_only), `gate.scope` containment, account-default cascade, raw-float override, public `screening_events` surface.


jiashuoz:
## Independent + adversarial review — dispositions

Two read-only review passes ran in parallel against the branch.

**Adversarial pass: no exploitable findings.** All 7 failure classes verified SAFE — the core #13 property (an agent-scoped credential cannot read or write its own detection tuning) holds on every traced path: HTTP handlers (`requireAccountScope` before any read), slimmed `AgentView` on *both* `getAgent` and `listAgents`, MCP admin-tiering, and the `WHERE user_id` store backstop. Fail-open in the derive mapping, PUT-replace reset, engine-column desync, and migration idempotency all checked clean.

**Convergent High finding (both reviewers) — FIXED** (`f7c0055`): migration 045 defaulted `*_scan_sensitivity` to `'off'` with no backfill, so a pre-existing scan-on agent read back `off` and a read-modify-write PUT would silently disable its live scan. Added an idempotent backfill deriving the level from the stored review threshold + a regression test (`TestProtectionSensitivityBackfill`).

**Deferred-with-reason:**
- *Legacy `verified_only` RMW 422 trap* (adversarial class 6): fail-**closed** (the anti-spoofing gate is never silently downgraded), so not a security issue — recorded as a known limitation; the deferred `gate.require_auth` flag is the clean resolution.
- *`UpdateAgentName` byte-vs-rune length* (independent #3): cosmetic 400-vs-422 edge on a 200-char multibyte name; non-blocking.

**Refuted/non-issues:** `medium`-band thresholds under `off` (harmless, scan gated separately); dead `UpdateAgentScanConfig`/`UpdateAgentInboundPolicy` store methods (no live callers, can't desync).

Full Go suite (unit+integration+e2e+contract), TS SDK (97), MCP (134), and all drift gates green; over-the-wire e2e against the real contract-server binary confirmed GET/PUT/422/AgentView-slim + derived DB columns.

PR #287, opened 2026-06-24
feat(account): self-serve rotation of the per-user relay signing secret

jiashuoz:
## Summary

Closes the security gap documented in `docs/design/api-v1-redesign.md` §13.3: when the per-user **signing-secret management API** was removed in the `/v1` cutover, the per-user **relay signing secret** stayed a *live* signing path — the SMTP relay signs inbound webhook deliveries (the `X-E2A-Auth-*` HMAC) with `GetUserSigningSecrets[0]` (falling back to the deployment-wide signer when a user has none), and `hitlnotify` reads it too. But a tenant who suspects that secret is compromised had **no self-serve rotation path**; rotation was an operator task. (The per-webhook `whsec_` rotate at `/v1/webhooks/{id}/rotate-secret` is a *different* key.)

This adds an **account-scoped** control to rotate it.

## Endpoint

`POST /v1/account/signing-secret/rotate`

- **Account scope only** — agent-scoped credentials get `403 forbidden` at the scope gate (an agent must not be able to rotate, or read, the secret that signs its own inbound deliveries).
- **Reveal-once** response, mirroring the webhook rotate pattern:
  ```json
  { "signing_secret": "…64 hex…", "secret_prefix": "…12 chars…", "created_at": "…RFC3339…" }
  ```
- Honors `Idempotency-Key` so a retried call replays the same secret instead of hard-rotating twice (which would discard the secret the caller already stored).

### Why this path
It sits under the existing `account` resource alongside `/v1/account/suppressions`, `/v1/account/export`, `DELETE /v1/account` — all account-admin self-service controls keyed off `requireAccountUser`. The secret is **per-user (account-level)**, not per-webhook, so `/v1/account/...` is the correct home (a sub-path `signing-secret/rotate` leaves room for a future GET/status without overloading the verb). Putting it under `/v1/webhooks/...` would have wrongly implied it's the per-webhook `whsec_` key.

## DECISION FLAGGED FOR REVIEW — hard rotation vs. grace window

The store is a multi-secret model (cap 5; relay signs with the newest, recipients may verify against any). That makes two rotation semantics possible:

1. **Grace-window rollover** (what the per-webhook `whsec_` rotate does): mint a new secret, keep the old valid for 24h, delete it later. Non-disruptive; in-flight deliveries still verify.
2. **Hard rotation** (what this PR implements): atomically delete **all** the user's signing secrets and insert one fresh secret in a single per-user-row-locked transaction. The compromised secret stops signing **and** verifying immediately.

I chose **(2)** because the gap §13.3 calls out is specifically *"a tenant who suspects their per-user relay-signing secret is compromised."* For compromise recovery you want the leaked key invalidated **now**, not 24h later. The trade-off — webhook deliveries already in flight that were signed with the old secret fail recipient-side verification — is the *intended* effect, not a regression. The non-disruptive case is already served by the per-webhook `rotate-secret` grace window.

**If reviewers prefer a grace window here too**, the change is small and localized to `identity.Store.RotateUserSigningSecret` (keep the newest old secret valid with a TTL instead of `DELETE … WHERE user_id`); the handler/SDK/spec stay as-is. Calling this out per the task's "flag the decision" instruction rather than expanding scope.

## Changes

- **Store** (`internal/identity/store.go`): new `RotateUserSigningSecret(ctx, userID) (*SigningSecret, error)` — atomic delete-all + insert-one under the existing `withUserSecretsLock` per-user row lock (same lock that serializes Create/Delete/Ensure), so it can never leave the user with zero secrets or race two rotations to a partial state. Reuses `generateSigningSecret`.
- **Relay**: no change. It already signs with `GetUserSigningSecrets[0]`, which is the new secret after rotation, and already falls back to the deployment signer when a user has none.
- **API** (`internal/httpapi/account.go`, `httpapi.go`, `internal/apiserver/apiserver.go`): new handler + route, `RotateRelaySigningSecret` Dep (optional → `501` if unwired), wired to the store method.
- **Spec + SDKs**: regenerated `api/openapi.yaml` and the TS + Python `generated/` bases; added a `account.rotateSigningSecret()` / `account.rotate_signing_secret()` convenience to the hand-written ergonomic SDK layer.

## Tests

- `internal/httpapi/account_test.go`: handler success (reveal-once fields present) + unauthorized 401.
- `internal/httpapi/scope_test.go`: agent-scoped credential → 403 in the account-only-routes matrix.
- `internal/identity/store_test.go`: DB-backed `TestRotateUserSigningSecret_HardReplace` — start with 3 secrets, rotate, assert exactly 1 remains, it's the new value, and every old value is gone.

Verification (all green): `go build ./...`, `go test ./internal/httpapi/... ./internal/apiserver/...`, the DB-backed identity signing-secret tests, `TestSpecGoldenNoDrift` (spec drift gate), and `make generate-sdk-check` (SDK drift gate). The TS SDK builds (`tsc`) and the Python client parses.

> Note: `internal/webhook` has pre-existing test failures on the `main` baseline (verified by stashing this diff) caused by shared `e2a_test` DB pollution from concurrent worktree runs — unrelated to this change.

## Not merging

This is a security-sensitive change; opening as a **proposal for review** (do not merge). The hard-rotation vs. grace-window decision above is the key thing to sign off on.


jiashuoz:
Closing — superseded by the decision to **remove** the per-user `webhook_signing_secrets` layer entirely rather than add rotation for it. Per research, nothing tenant-side verifies that secret (the management API was removed in the v1 cutover) and webhook delivery authenticity is carried by the per-webhook `whsec_`. The relay auth-header signing and HITL magic-link signing will move to the deployment `E2A_HMAC_SECRET` (its existing fallback), and the table is being dropped. Tracked in the removal PR (feat/remove-per-user-signing-secret).

PR #294, opened 2026-06-25
fix(web): render inbound email body from parsed text, not raw MIME

jiashuoz:
## Problem
Inbound emails rendered as raw MIME in the conversation view — literal `<div>`/`<br>` markup and quoted-printable `=` soft-breaks instead of readable text.

## Root cause (frontend-only)
The backend already returns a clean parsed body in **`parsed.text`** (`mailparse.ParsedBody`: prefers `text/plain`, else renders HTML→text, decodes quoted-printable/base64, strips quoted reply chains). But:
- `getMessageDetail`'s inbound mapping dropped `w.parsed` and `InboundMessageDetail` had no `parsed` field, and
- `ThreadBubble` / the focus view read `body?.text` (only populated for outbound held drafts) → always fell through to a client-side `decodeRawBody(raw_message)` that emits the raw MIME body.

## Fix
- Thread `parsed` through `getMessageDetail` + add it to `InboundMessageDetail`.
- `ThreadBubble` and the message focus view prefer `parsed.text`; the raw decode stays only as a last-resort fallback.
- Regression test (`view/page.test.tsx`): `parsed.text` is rendered and the raw HTML/QP body is not.
- Removed a dead `formatRelativeAge` import in `ThreadBubble`.

## Verification
`tsc` clean; `npm test` 238/238 (incl. new regression); `npm run lint` 0 errors; `npm run build` passes.

### DIFF
--- a/web/src/app/(app)/inboxes/(view)/messages/view/page.test.tsx
+++ b/web/src/app/(app)/inboxes/(view)/messages/view/page.test.tsx
@@ -207,6 +207,31 @@ describe("AgentMessageFocusPage", () => {
     expect(screen.getByText(/Attached is the renewal contract/)).toBeInTheDocument();
   });
 
+  it("prefers the backend-parsed text over the raw MIME body for inbound", async () => {
+    // Regression: inbound rows carry the clean body in `parsed.text` (text/plain
+    // or HTML→text, quoted-printable decoded). The bug rendered the raw MIME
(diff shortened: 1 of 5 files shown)


jiashuoz:
Superseded by #307, which folds in this parsed-text fix and adds full HTML email rendering (sandboxed iframe + CSP + DOMPurify). Closing in favor of that PR.

PR #316, opened 2026-06-26
fix(relay): fail inbound gate closed for shared-relay senders (#299)

### LINKED ISSUE #299: Gating: inbound allowlist matches the relay envelope, not the DMARC-aligned From — per-agent allowlisting is impossible
## Summary

An inbound trust gate with an `allowlist` (or `domain`) policy matches the sender against the **SMTP envelope `MAIL FROM` (the relay return-path)**, not the **DMARC-aligned `From:`**. For relayed agent mail the envelope is `agent@send.e2a.dev` for *every* agent, so:

- Allowlisting a **specific agent** (`alice-test@agents.e2a.dev`) silently **fails** — the message is still blocked/flagged because the gate is comparing against `send.e2a.dev`.
- The only entry that matches is the **shared relay domain** `send.e2a.dev` — which lets **every** agent through.

Net: **per-agent inbound allowlisting cannot be expressed today.** The gate is matching an identity (the bounce path) that is neither what the user reasons about nor a per-sender identity.

## Reproduction

Two agents, `alice-test@agents.e2a.dev` → `bob-test@agents.e2a.dev`. Bob inbound gate = `block`, vary only the allowlist. Alice outbound `open` so nothing is held on the way out. One isolated message per config, no config changes until each resolves (observed via an `email.*` webhook).

| Bob inbound allowlist | Result |
|---|---|
| `["alice-test@agents.e2a.dev"]` (the sending agent) | **blocked** ❌ |
| `["send.e2a.dev"]` (domain policy, the relay) | **delivered** ✅ |

The only difference is the allowlist value; the second matches because the gate is evaluating the relay envelope, not the agent.

### Evidence the authenticated From is the relay, not the agent
- Inbound auth on these messages: `dmarc: dkim-aligned (d=send.e2a.dev, from=send.e2a.dev)` — DMARC evaluated the From domain as `send.e2a.dev`.
- Recipient mail clients render the sender as **"Alice (test) via e2a `<agent@send.e2a.dev>`"**; `alice-test@agents.e2a.dev` appears only as Reply-To / display name.
- `list_messages` shows `_from: alice-test@agents.e2a.dev`, but that is e2a's normalized/display sender, **not** the wire `From:` header.

So this is *not* simply "match `From` instead of envelope" — for relayed mail the wire `From:` is also `send.e2a.dev`. The deeper cause is that relayed intra-system mail carries **no authenticated per-agent identity**.

## What the matching key should be

A trust gate must match an **authenticated** identity (never `Reply-To` or an unverified `From` — both are spoofable). There are two distinct cases:

### 1. Intra-system mail (agent → agent) — match e2a's known internal sender
e2a has **ground truth** for sends that originate inside the platform: it knows `alice-test` sent the message. The gate should match the allowlist against that authenticated internal sender identity — strictly better than parsing any header, and it makes per-agent allowlisting work **today, with no dependency on #113**.

### 2. External inbound mail — match the DMARC-aligned `From:`
For mail arriving from outside, match against the `From:` **only when DMARC passes/aligns**; otherwise treat it as unverified and fall through to the gate action.

Important nuance: DMARC authenticates the **domain**, not the local-part. DKIM `d=` alignment / SPF alignment proves `From` *domain* authenticity, not that the local-part is a specific mailbox. So:
- **Domain-granularity** allowlist entries (`@partner.com`, or `domain` policy) are cryptographically sound.
- **Full-address** entries (`cfo@partner.com`) are sound only to the degree the From domain governs its own local-parts (a normal email-trust assumption).

For relayed e2a agents specifically, per-agent granularity on the **external** path only becomes possible once each agent sends from its own DKIM-aligned domain — see **#113** (per-customer SES sending identity; drop the `via e2a` shim). That dependency applies to case (2) only; case (1) can and should be fixed independently.

## Proposed behavior

- **Internal sends:** evaluate the gate against the authenticated originating agent, not header/envelope inference.
- **External inbound:** evaluate the allowlist against the DMARC-aligned `From:` (domain-sound; full-address as trusted-per-domain). Ignore the `From` for matching when DMARC fails to align.
- **Never** match the envelope `MAIL FROM`/return-path or unauthenticated `Reply-To`/display name for trust decisions.
- Document the allowlist semantics (what identity is matched, and that external full-address matching is domain-trust-bounded).

## Impact

- **Security gate is currently ineffective at its stated job.** Users believe they've allowlisted a specific sender; they've actually allowlisted the entire relay (or nothing). Both outcomes are wrong and silent.
- Applies to inbound `allowlist`/`domain` policies with any action (`review`/`block`/`flag`). Confirmed on `block`; same matching path drives the others.

## Related

- #113 — per-customer SES sending identity. Gates the **external-path** per-agent fix (so the agent's own domain becomes the DMARC-aligned From). The **internal** agent-to-agent fix does not depend on it.

jiashuoz:
Closes #299.

## Problem
The inbound trust gate (`allowlist`/`domain`) matches on the message **From** identity. Mail relayed for a **non-sending-verified** agent goes out under the shared `agent@<outboundFromDomain>` address (`internal/outbound/sender.go`) — it authenticates (DMARC passes for the relay domain) but is the **same address for every such agent**, so it has no per-agent identity.

Before this PR:
- A full-address allowlist entry (`alice@agents.e2a.dev`) **never matched** the shared sender → the intended sender was always flagged.
- Domain-allowlisting the relay domain (`send.e2a.dev`) **silently admitted every unverified agent** indiscriminately.

## Fix
Treat a sender whose From domain is the relay's own outbound domain (or a subdomain) as **unresolvable**, and **fail the gate closed** under `allowlist`/`domain`. `open` is unaffected — open means open.

| Sender | `open` | `allowlist`/`domain` |
|---|---|---|
| External / sending-verified agent (own domain) | pass | matched normally |
| Unverified intra-system (`agent@send.e2a.dev`) | pass | **flagged** (can't satisfy the gate even if the relay domain/address is listed) |

Flagged still means **delivered + annotated + `email.flagged`** — never dropped.

## Changes
- `internal/inboundpolicy/policy.go` — `EvaluateIngestion` gains `senderResolvable`; fails closed under `allowlist`/`domain` when false.
- `internal/relay/server.go` — new `session.senderResolvable()` recognizes the shared-relay address, mirroring `envelopeFromTrusted`'s domain/subdomain matching.
- `internal/httpapi/protection.go` — gate doc comment updated.
- Tests at both layers (open-still-passes + the domain-hole regression).

## Scope (split out)
This PR is **only** the shared-relay correctness fix. A separate concern surfaced during review — the gate matches the From header without consulting DMARC, so a spoofed `From: ceo@allowlisted.com` could pass — is tracked in #318. It was deliberately **not** folded in here: requiring a DMARC-aligned pass for an allowlist match would flag *every* allowlisted sender that isn't DMARC-aligned (forwarded/list mail, domains without DMARC), a broad backward-compat change better delivered as an opt-in posture. Selectively *allowing* a specific unverified agent (trusted `X-E2A-Sender-Agent` header) also remains future work.

## Verification
`go build ./...`, `go vet`, relay/inboundpolicy/httpapi unit tests, and the `internal/e2e` inbound-policy suite all pass. No `doc:` tags changed, so the OpenAPI spec/SDK drift gate is unaffected.

### DIFF
--- a/internal/httpapi/protection.go
+++ b/internal/httpapi/protection.go
@@ -25,6 +25,16 @@ const protectionBetaDoc = "Beta: the agent protection config is unstable — its
 // gate/scan object and get the safe-permissive default. The three TOP-LEVEL
 // keys (inbound/outbound/holds) stay required (design D3) — a missing section
 // is a 422, not a silent reset.
+//
+// INBOUND gate identity note (#299): the inbound allowlist/domain gate matches on
+// the message's From address, which only carries a specific per-agent identity for
+// senders on a sending-verified domain. Mail from agents NOT yet sending-verified
+// is relayed under the shared "via e2a" address (internal/outbound/sender.go) — it
+// authenticates but is the same address for every such agent. The relay treats that
+// shared sender as unresolvable (see senderResolvable), so it can never satisfy an
+// allowlist/domain gate and is flagged (fail closed); under "open" it still passes.
+// Net: per-agent inbound allowlisting is reliable for sending-verified senders;
+// unverified intra-system senders are uniformly gated, not silently admitted.
 type ProtectionGateView struct {
 	Policy    string   `json:"policy,omitempty" enum:"open,allowlist,domain" default:"open" doc:"Trust gate: open (all), domain (listed domains), allowlist (listed addresses)."`
 	Allowlist []string `json:"allowlist,omitempty" nullable:"false" maxItems:"1000" doc:"Addresses (allowlist) or domains (domain) the gate trusts; ignored for open."`
--- a/internal/inboundpolicy/policy.go
+++ b/internal/inboundpolicy/policy.go
@@ -48,22 +48,40 @@ type Decision struct {
 	Reason  string // human-readable; empty when not flagged
 }
 
+// unresolvableSenderReason is the fail-closed flag reason for a sender with no
+// specific authenticated identity (shared-relay "via e2a" mail) under a gating
+// policy — see senderResolvable in the relay and issue #299.
+const unresolvableSenderReason = "sender has no resolvable per-agent identity (shared relay), so it cannot match a per-agent inbound gate"
+
 // EvaluateIngestion applies the agent's ingestion policy to an inbound message.
 //   - policy: the agent's inbound_policy (unknown/empty → treated as Open).
 //   - allowlist: the agent's inbound_allowlist (addresses for Allowlist,
 //     domains for Domain); matching is case-insensitive.
 //   - senderEmail: the message's display sender (From identity).
+//   - senderResolvable: whether senderEmail maps to a SPECIFIC authenticated
+//     sender. False for mail relayed under the shared "via e2a" address, which
+//     authenticates (DMARC passes for the relay domain) but carries no per-agent
+//     identity (#299). An unresolvable sender can never legitimately satisfy a
+//     per-agent allowlist/domain gate, so it is treated as a non-match. Open is
+//     unaffected — open means open.
 //
-// Flagged is fail-closed for the gating postures: an empty/garbage sender, or
-// an empty allowlist, flags everything (you opted into a gate but listed no one).
-func EvaluateIngestion(policy string, allowlist []string, senderEmail string) Decision {
+// Flagged is fail-closed for the gating postures: an empty/garbage sender, an
+// empty allowlist, or an unresolvable sender flags everything (you opted into a
+// gate but the sender cannot be matched).
+func EvaluateIngestion(policy string, allowlist []string, senderEmail string, senderResolvable bool) Decision {
 	switch policy {
 	case Allowlist:
+		if !senderResolvable {
+			return Decision{Flagged: true, Reason: unresolvableSenderReason}
+		}
 		if containsFold(allowlist, strings.TrimSpace(senderEmail)) {
 			return Decision{}
 		}
 		return Decision{Flagged: true, Reason: "sender not on the agent's inbound allowlist"}
 	case Domain:
+		if !senderResolvable {
+			return Decision{Flagged: true, Reason: unresolvableSenderReason}
+		}
 		dom := domainOf(senderEmail)
 		if dom != "" && containsFold(allowlist, dom) {
 			return Decision{}
--- a/internal/inboundpolicy/policy_test.go
+++ b/internal/inboundpolicy/policy_test.go
@@ -4,25 +4,32 @@ import "testing"
 
 func TestEvaluateIngestion(t *testing.T) {
 	tests := []struct {
-		name      string
-		policy    string
-		allowlist []string
-		sender    string
-		flagged   bool
+		name       string
+		policy     string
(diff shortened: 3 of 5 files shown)


jiashuoz:
## 🔬 Review summary (3 review agents + adversarial verification)

Ran three independent review passes on this branch, then a focused adversarial re-verification after the DMARC fix was added.

### Pass 1 — Correctness / quality
**Verdict: correct, mergeable.** Confirmed the shared-relay From is exactly `agent@<OutboundSMTP.FromDomain>`, that both the relay's `outboundFromDomain` and the outbound sender read the *same* config value (can't drift), and that `open` is untouched. Only LOW/NIT findings (unconfigured-domain fail-open is intentional; minor parser duplication).

### Pass 2 — Regression / blast radius
**Verdict: acceptable for GA.** Default `inbound_policy = open` (migration 033; 047 back-fills legacy `verified_only → open`), and `open` ignores the new predicates — so the vast majority of agents are unaffected. Only agents that *opted into* a gating policy change behavior, and only to fail closed for senders that were never granularly matchable before. One call-out for the changelog: with `inbound_policy_action = review|block`, such senders now hold/quarantine rather than deliver-flagged.

### Pass 3 — Adversarial (found a HIGH, now fixed)
The adversarial pass found that the gate matched the From header **without consulting DMARC** — an external attacker could spoof `From: ceo@allowlisted-corp.com` (DMARC fail) and be delivered as a trusted, allowlisted sender. **This is now fixed** in the second commit: an allowlist match requires the From to be DMARC-aligned-authenticated (`DMARC == pass`, the aligned verdict — *not* `DomainAuthenticated()`, which is SPF/DKIM pass without alignment and remains spoofable).

### Re-verification of the DMARC fix
A fresh adversarial pass confirmed:
- **Bypass closed** — `gateTrust` runs *before* any allowlist match; a spoofed From (`DMARC != pass`) is flagged. `EvaluateIngestion` is the sole production gate evaluator (no alternate path).
- **Correct signal** — `checkDMARC` only returns `pass` on From-aligned DKIM/SPF; an attacker controlling only envelope SPF can't satisfy it (eTLD+1 mismatch).
- **No regressions** — verified DKIM-aligned agents still match; shared-relay mail still flagged via `senderResolvable`; `open` still passes unauthenticated mail.
- All relay / inboundpolicy / httpapi / emailauth tests pass.

**Net: mergeable.** Selectively *allowing* a specific unverified agent (trusted `X-E2A-Sender-Agent` header) remains a future enhancement, not a GA blocker.

jiashuoz:
Per discussion, the DMARC commit was **dropped from this PR** and the branch reset to the shared-relay-only fix (force-pushed). Reason: always-on DMARC enforcement flagged *every* allowlisted sender that isn't DMARC-aligned — broke the `internal/e2e` inbound-policy suite, and would be a broad backward-compat change for legitimate forwarded/non-DMARC mail. The DMARC-spoofing concern is now tracked separately in #318 as an **opt-in auth posture**. This PR is back to a clean, all-green scope.

PR #319, opened 2026-06-26
feat(relay): opt-in require-authenticated inbound gate (#318)

### LINKED ISSUE #318: Inbound allowlist/domain gate doesn't verify DMARC — spoofable From (opt-in auth posture)
### Summary
The inbound trust gate (`allowlist`/`domain`) matches on the message `From` identity **without consulting DMARC**. An external attacker can send `From: ceo@allowlisted-corp.com` with SPF/DKIM/DMARC all failing and be delivered as a **trusted, allowlisted** sender with no `email.flagged` — the allowlist is spoofable.

Surfaced by the adversarial review of #316 (PR for #299).

### Why it's not just a one-line fix
Requiring a DMARC-aligned pass (`emailauth` `DMARC == pass`, the *aligned* verdict — not `DomainAuthenticated()`) before honoring an allowlist match closes the spoof, but it **flags every allowlisted sender that isn't DMARC-aligned** — forwarded mail, mailing lists, and any sender domain without DMARC. We tried folding it into #316 and it broke the `internal/e2e` inbound-policy suite (`TestInboundPolicy_AllowlistAcceptsMember`, `TestInboundPolicy_ReceivedSurfacesAuthenticatedFrom`), which inject locally-unsigned mail — i.e. exactly the legitimate-but-unaligned case. The e2e harness also can't currently produce DMARC-passing mail (no DKIM-signing), so the positive "allowlist accepts an authenticated member" path isn't testable as-is.

Because the gate **flags rather than rejects**, the impact is "delivered-flagged-not-trusted," not a bounce — milder than a Gmail-style DMARC reject — but it's still a broad default behavior change.

### Proposed direction
Bring DMARC-enforcement back as an **opt-in posture**, not an always-on requirement. The `inboundpolicy` package doc already anticipates this:
> "A DMARC-alignment 'verified_only' posture was removed pre-GA; it may return as an additive policy later."

Options to weigh:
1. A per-agent **"require authenticated sender"** toggle that composes with `allowlist`/`domain` (allowlist match honored only when the From is DMARC-aligned). Backward-compatible by default.
2. Reinstate a `verified_only` policy / gate sub-flag.
3. Tier the verdict (treat DMARC `fail` — active misalignment, a strong spoof signal — differently from `none` — no auth to evaluate). Note: accepting `none` keeps a spoofing hole, so this only narrows, not closes, the gap.

### Acceptance
- Spoofed `From` on a gated agent can be made to flag (when the operator opts in), with a distinct flag reason.
- Default behavior unchanged (allowlist still matches non-DMARC senders) so legitimate forwarded/list mail isn't flagged unless opted in.
- e2e harness can produce a DMARC-passing message so the positive authenticated-allowlist path is covered.

### Implementation notes
A working always-on implementation exists at commit `a37feb2` (reverted from #316) — `EvaluateIngestion` gained a `senderAuthenticated` bool fed from `domainAuth.DMARC == StatusPass`, with `gateTrust` failing closed. It can be lifted and gated behind the opt-in toggle.

jiashuoz:
Closes #318. Follow-up to #316 (#299).

## Problem
The inbound trust gate (`allowlist`/`domain`) matches on the message `From` without consulting DMARC, so a spoofed `From: ceo@allowlisted-corp.com` (SPF/DKIM/DMARC all failing) is delivered as a **trusted, allowlisted** sender with no `email.flagged`. Surfaced by the adversarial review of #316.

An **always-on** DMARC requirement is too aggressive — it flags *every* legitimate allowlisted sender that isn't DMARC-aligned (forwarded mail, mailing lists, domains without DMARC). We tried that in #316 and it broke the e2e inbound-policy suite for exactly that reason.

## Fix — opt-in, composable `require_authenticated`
A per-agent **`inbound_require_auth`** flag (default `false` → fully backward-compatible). It's the additive return of the `verified_only` posture that migration 047 anticipated ("may return later as a composable, additive policy"). When `true`, an inbound `From` that is not **DMARC-aligned-authenticated** (RFC 7489) is flagged **regardless of policy**:

| Policy | `require_auth=false` (default) | `require_auth=true` |
|---|---|---|
| `open` | accept all | flag unauthenticated (≈ old `verified_only`) |
| `allowlist`/`domain` | match as today | match **and** require authentication |

Uses the **aligned** DMARC verdict (`DMARC == pass`), not `DomainAuthenticated()` (SPF/DKIM pass without From-alignment is still spoofable). The gate **flags, never rejects** — unauthenticated mail is delivered-flagged, not bounced — so opting in doesn't break indirect mail the way a hard DMARC reject would.

## Changes
- **`migrations/050_inbound_require_auth.sql`** — `inbound_require_auth BOOLEAN NOT NULL DEFAULT false` (idempotent, non-rewriting).
- **`internal/identity`** — `AgentIdentity` + `ProtectionConfig` field, both SELECT/Scan paths, `UpdateAgentProtection` write, DB round-trip test.
- **`internal/httpapi/protection.go`** — `ProtectionGateView.require_authenticated` (inbound gate only; ignored on outbound). **OpenAPI + TS/Python SDK bases regenerated** (`make generate`).
- **`internal/inboundpolicy`** — `EvaluateIngestion` now takes a `Request` struct; `RequireAuth` + `SenderAuthenticated` as an additive precondition across all policies.
- **`internal/relay/server.go`** — derives `senderAuthenticated` from `domainAuth.DMARC == pass`.
- **Tests** — unit matrix (opt-in on/off, spoofed-From, open+require_auth) + e2e (`TestInboundPolicy_RequireAuthFlagsUnauthenticated`: opt-in flags an unauthenticated allowlisted sender end-to-end; the default-off `AllowlistAcceptsMember` still passes).

## Verification
`go build ./...`, `go vet`, unit (`inboundpolicy`/`relay`/`httpapi` incl. the spec-golden drift gate), DB-backed `identity` round-trip, and the `internal/e2e` inbound-policy suite all pass. `api/openapi.yaml` + both SDK bases regenerated, so the spec/SDK drift gates are satisfied.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -125,6 +125,8 @@ components:
           type: string
         inbound_policy_action:
           type: string
+        inbound_require_auth:
+          type: boolean
         inbound_scan:
           type: string
         inbound_scan_block_threshold:
@@ -193,6 +195,7 @@ components:
         - pending_count
         - webhook_healthy
         - inbound_policy
+        - inbound_require_auth
         - inbound_policy_action
         - outbound_policy
         - outbound_policy_action
@@ -1485,6 +1488,9 @@ components:
             - allowlist
             - domain
           type: string
+        require_authenticated:
+          description: "Inbound gate only: when true, flag any sender whose From is not DMARC-aligned-authenticated, regardless of policy. Default false. Ignored on the outbound gate."
+          type: boolean
       type: object
     ProtectionHoldsView:
       additionalProperties: false
--- a/internal/e2e/inbound_policy_e2e_test.go
+++ b/internal/e2e/inbound_policy_e2e_test.go
@@ -133,6 +133,47 @@ func TestInboundPolicy_AllowlistAcceptsMember(t *testing.T) {
 	}
 }
 
+// TestInboundPolicy_RequireAuthFlagsUnauthenticated: with the opt-in
+// inbound_require_auth flag set (#318), an allowlisted sender whose From is NOT
+// DMARC-aligned-authenticated (the e2e harness injects locally-unsigned mail, so
+// DMARC=fail) is FLAGGED — a spoofed allowlisted address cannot be trusted. This
+// is the opt-in inverse of TestInboundPolicy_AllowlistAcceptsMember, where the
+// same unauthenticated sender is NOT flagged because require_auth defaults off.
+func TestInboundPolicy_RequireAuthFlagsUnauthenticated(t *testing.T) {
+	ts, pool, agent, receiver := setupFlaggedAgent(t, "allowlist",
+		[]string{"friend@trusted.com"}, "agent@reqauth.example.com", "reqauth.example.com")
+
+	// Opt the agent into authenticated-sender enforcement.
+	if _, err := pool.Exec(context.Background(),
+		`UPDATE agent_identities SET inbound_require_auth = true WHERE id = $1`, agent.ID); err != nil {
+		t.Fatalf("set inbound_require_auth: %v", err)
+	}
+
+	msg := "From: friend@trusted.com\r\nTo: agent@reqauth.example.com\r\nSubject: Hi\r\n\r\nHello"
+	if err := smtp.SendMail(ts.SMTPAddr, nil, "friend@trusted.com", []string{"agent@reqauth.example.com"}, []byte(msg)); err != nil {
+		t.Fatalf("SendMail: %v", err)
+	}
+
+	tick(t, ts)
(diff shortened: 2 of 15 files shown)


jiashuoz:
Closing in favor of **documenting the limitation** rather than shipping an opt-in toggle. Decision: the inbound allowlist matches the From identity as presented and does not itself require DMARC; rather than add a per-agent `require_authenticated` flag (DB + API + SDK surface), we'll document the gate's trust model at the API layer so consumers know to check the message's authentication result for spoofing-sensitive trust. See the docs change + #318 closure.

PR #359, opened 2026-07-01
Add Gemini LLM-as-detector layer

hyj28:
## Summary

Adds `GeminiDetector` to `internal/piguard`, implementing the existing `Detector`
interface via the Gemini REST API (stdlib `net/http`, no new module dependencies).
Ported from the e2a eval framework where the same prompt achieved AUC 0.97–0.99 on
the injection + phishing benchmark. The detector is entirely opt-in: it is only
wired into the screening engine when `GEMINI_API_KEY` (or `GOOGLE_API_KEY`) is set
in the environment; deployments without the key are unaffected.

## Operational risk

- **No behaviour change for existing deployments** — `NewGeminiDetector` returns an
  error when no key is present, and the engine falls back to heuristics-only, same
  as before.
- **Additive to the screening engine** — the Engine already handles a mix of
  detectors; Gemini is just appended. A timeout or API error is excluded from the
  aggregate (fail-open is not possible by design).
- **API key hygiene** — key is sent via `x-goog-api-key` header only, never in URL
  params, logs, or error strings.
- **External API dependency** — adds a runtime dependency on
  `generativelanguage.googleapis.com` when the key is set. Transient failures
  (429/5xx) are retried with exponential backoff; persistent failures degrade
  gracefully (detector excluded, engine may report `Degraded` → routes to review).

## Test plan

- [ ] `make test-unit` passes (gemini tests use an `httptest` mock — no real API key needed)
- [ ] Start server without `GEMINI_API_KEY` — confirm no Gemini log line, heuristics-only behaviour unchanged
- [ ] Start server with `GEMINI_API_KEY` — confirm `[piguard] Gemini detector enabled` log line at startup
- [ ] Send a test email containing an injection payload — verify `screening_events.raw` includes a `gemini` detector entry with a non-zero score

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -127,3 +127,13 @@ limits:
   # billing_hook_url: "http://billing:9000/api/internal/billing/cancel"
   billing_hook_url: ""
 
+
+
+# — Content screening (piguard) ——————————————————————————————————————————————
+# The Gemini LLM-as-detector layer is enabled by setting the environment
+# variable GEMINI_API_KEY (or GOOGLE_API_KEY) to a Google AI Studio key.
+# When the variable is absent the Gemini detector is silently skipped and
+# the dependency-free heuristics detector runs alone.
+#
+# Obtain a key: https://aistudio.google.com/apikey
+# Model default: gemini-2.5-flash (change via GEMINI_EVAL_MODEL env var)
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -133,7 +133,8 @@ limits:
 # The Gemini LLM-as-detector layer is enabled by setting the environment
 # variable GEMINI_API_KEY (or GOOGLE_API_KEY) to a Google AI Studio key.
 # When the variable is absent the Gemini detector is silently skipped and
-# the dependency-free heuristics detector runs alone.
+# the dependency-free heuristics detector runs alone. It only screens INBOUND
+# mail (internal/relay); outbound agent-mail screening is heuristics-only.
 #
 # Obtain a key: https://aistudio.google.com/apikey
-# Model default: gemini-2.5-flash (change via GEMINI_EVAL_MODEL env var)
+# Model default: gemini-3.1-flash-lite (change via GEMINI_EVAL_MODEL env var)
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -138,3 +138,6 @@ limits:
 #
 # Obtain a key: https://aistudio.google.com/apikey
 # Model default: gemini-3.1-flash-lite (change via GEMINI_EVAL_MODEL env var)
+# Kill-switch: set E2A_GEMINI_DETECTOR_ENABLED=false to disable Gemini even when
+# a key is configured (e.g. to isolate whether Gemini or heuristics drove a given
+# block/review outcome, or to roll back without removing the credential).
--- a/internal/agent/api.go
+++ b/internal/agent/api.go
@@ -521,7 +521,7 @@ func NewAPI(store *identity.Store, sender *outbound.Sender, smtpRelay *outbound.
 	return &API{
 		store:        store,
 		sender:       sender,
-		screen:       piguard.NewEngine(piguard.EngineConfig{}, piguard.NewHeuristicsDetector()),
+		screen:       buildAgentScreenEngine(),
 		smtpRelay:    smtpRelay,
 		userAuth:     userAuth,
 		usage:        usage,
@@ -540,6 +540,18 @@ func NewAPI(store *identity.Store, sender *outbound.Sender, smtpRelay *outbound.
 		dcrLimit:      ratelimit.New(1*time.Hour, 10),    // 10 OAuth client registrations per IP per hour
 		downloadLimit: ratelimit.New(1*time.Minute, 120), // 120 attachment downloads per IP per minute
 	}
+}
+
+// buildAgentScreenEngine constructs the piguard screening engine for outbound agent
+// mail. The heuristics detector is always included. The Gemini detector is added when
+// GEMINI_API_KEY or GOOGLE_API_KEY is set in the environment.
+func buildAgentScreenEngine() *piguard.Engine {
+	detectors := []piguard.Detector{piguard.NewHeuristicsDetector()}
+	if d, err := piguard.NewGeminiDetector(piguard.GeminiConfig{}); err == nil {
+		detectors = append(detectors, d)
+		log.Printf("[piguard] Gemini detector enabled for agent outbound screening")
+	}
+	return piguard.NewEngine(piguard.EngineConfig{}, detectors...)
 }
 
 // SetAPIURL overrides the API/issuer base URL (default: publicURL). Set it
--- a/internal/agent/api.go
+++ b/internal/agent/api.go
@@ -543,15 +543,17 @@ func NewAPI(store *identity.Store, sender *outbound.Sender, smtpRelay *outbound.
 }
 
 // buildAgentScreenEngine constructs the piguard screening engine for outbound agent
-// mail. The heuristics detector is always included. The Gemini detector is added when
-// GEMINI_API_KEY or GOOGLE_API_KEY is set in the environment.
+// mail (see screenOutbound in screening.go, the engine's only caller — it always
+// passes Direction: piguard.DirectionOutput). Heuristics-only: GeminiDetector is
+// deliberately NOT wired in here. Its prompt (geminiSystemPrompt in
+// piguard/gemini.go) classifies only inbound injection/phishing aimed at the AI
+// agent and ignores Request.Direction entirely, so on outbound mail it would
+// produce false-positive holds (an agent quoting injection-like content scores as
+// injection) while missing the actual outbound concern (egress/exfiltration),
+// which only the heuristics detector currently checks for Direction ==
+// DirectionOutput. Revisit once Gemini has an outbound/exfiltration-aware prompt.
 func buildAgentScreenEngine() *piguard.Engine {
-	detectors := []piguard.Detector{piguard.NewHeuristicsDetector()}
-	if d, err := piguard.NewGeminiDetector(piguard.GeminiConfig{}); err == nil {
-		detectors = append(detectors, d)
-		log.Printf("[piguard] Gemini detector enabled for agent outbound screening")
-	}
-	return piguard.NewEngine(piguard.EngineConfig{}, detectors...)
+	return piguard.NewEngine(piguard.EngineConfig{}, piguard.NewHeuristicsDetector())
 }
 
 // SetAPIURL overrides the API/issuer base URL (default: publicURL). Set it
--- a/internal/piguard/gemini.go
+++ b/internal/piguard/gemini.go
@@ -14,7 +14,9 @@ import (
 )
 
 const (
-	geminiDefaultModel    = "gemini-2.5-flash"
+	// geminiDefaultModel is the most cost/latency-efficient GA Gemini model as of
+	// this writing. Override via GeminiConfig.Model or GEMINI_EVAL_MODEL env var.
+	geminiDefaultModel    = "gemini-2.5-flash-lite"
 	geminiAPIBase         = "https://generativelanguage.googleapis.com/v1beta"
 	geminiMaxOutputTokens = 2048
 	geminiMaxBodyChars    = 4000
@@ -167,9 +169,9 @@ func (g *GeminiDetector) formatEmail(req Request) string {
 }
 
 // generate calls the Gemini REST API with exponential backoff on transient errors.
-// It tries thinking_budget=0 first (saves tokens / latency); if the model rejects
-// it with an HTTP 400 mentioning "budget" or "thinking", it retries once immediately
-// without the thinking config before entering the normal retry loop.
+// It sends the model-appropriate thinking config first (thinkingBudget=0 for Gemini
+// 2.x, thinkingLevel="low" for Gemini 3.x). If the model rejects the thinking
+// config with HTTP 400, it retries once without any thinking config.
 func (g *GeminiDetector) generate(ctx context.Context, emailText string) (string, error) {
 	disableThinking := true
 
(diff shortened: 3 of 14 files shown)


internal/piguard/gemini.go
```diff
@@ -0,0 +1,406 @@
...
+	"io"
+	"math"
+	"net/http"
+	"os"
+	"strings"
+	"time"
+)
+
+const (
+	geminiDefaultModel    = "gemini-2.5-flash"
```
jiashuoz: Use `gemini-3.1-flash-lite` as the default here instead of `gemini-2.5-flash`.

It's the current most cost-efficient multimodal Gemini model, explicitly positioned for exactly this kind of high-volume, latency/cost-sensitive classification, and it's cheaper + faster than 2.5 Flash. Same `v1beta/models/{model}:generateContent` endpoint, so no other wiring changes.

Two things to check when you switch:
- **Confirm the stable model string** — `gemini-3.1-flash-lite` may still be preview-tagged in the API. If so, either pin the `-preview` alias knowingly or keep `2.5-flash-lite` (GA) as the safe default.
- **Thinking-disable needs updating.** The Gemini 3 series replaced the numeric `thinkingBudget` with `thinking_level` (low/high). The current `thinkingConfig.thinkingBudget: 0` may be ignored/rejected on a 3.x model, which would trip the fallback that silently turns thinking *back on* — the opposite of intended. Update the thinking-off config to the Gemini 3 shape and re-test (another point in favor of the official `google.golang.org/genai` SDK, which tracks these config changes).

internal/agent/api.go
```diff
@@ -542,6 +542,18 @@ func NewAPI(store *identity.Store, sender *outbound.Sender, smtpRelay *outbound.
 	}
 }
 
+// buildAgentScreenEngine constructs the piguard screening engine for outbound agent
+// mail. The heuristics detector is always included. The Gemini detector is added when
+// GEMINI_API_KEY or GOOGLE_API_KEY is set in the environment.
+func buildAgentScreenEngine() *piguard.Engine {
```
jiashuoz, line 555: **Blocking — inbound-only detector wired into outbound screening.** This adds Gemini to the engine used by `screenOutbound`, which calls `Evaluate` with `Direction: piguard.DirectionOutput`. But `geminiSystemPrompt` only classifies *inbound* injection/phishing ("instructions to the AI agent processing this email") and `Inspect` ignores `req.Direction` entirely. Outbound screening exists to catch egress/exfiltration (the heuristics detector switches behavior for outbound; Gemini does not).

Consequences on the agent's own outgoing mail: false-positive holds (an agent quoting injection-like content gets scored as injection, feeding the weighted aggregate → routes to review/block) and a blind spot (the prompt never asks about exfiltration, which is the point of outbound screening).

Fix: either don't add Gemini to this engine (wire it only in `relay/server.go`), or branch the prompt on `req.Direction` with an exfil-oriented outbound prompt. If you drop it here, add a test asserting the outbound engine is heuristics-only.

config.example.yaml
```diff
@@ -127,3 +127,13 @@ limits:
...
+
+
+# — Content screening (piguard) ——————————————————————————————————————————————
+# The Gemini LLM-as-detector layer is enabled by setting the environment
+# variable GEMINI_API_KEY (or GOOGLE_API_KEY) to a Google AI Studio key.
+# When the variable is absent the Gemini detector is silently skipped and
+# the dependency-free heuristics detector runs alone.
+#
+# Obtain a key: https://aistudio.google.com/apikey
+# Model default: gemini-2.5-flash (change via GEMINI_EVAL_MODEL env var)
```
jiashuoz: **Blocking — documented env var does nothing.** This says the model is changeable via `GEMINI_EVAL_MODEL`, but nothing reads it: both `buildScreenEngine`/`buildAgentScreenEngine` call `NewGeminiDetector(GeminiConfig{})` with an empty `Model`, so it's always the hardcoded default. Either wire it (`if cfg.Model == "" { cfg.Model = os.Getenv("GEMINI_EVAL_MODEL") }` before applying `geminiDefaultModel`) or delete the claim from the comment.

internal/piguard/gemini.go
```diff
@@ -0,0 +1,406 @@
...
+		body = body[:geminiMaxBodyChars]
+	}
+	return fmt.Sprintf("Subject: %s\nFrom: %s\n\n%s", subject, req.Sender, body)
+}
+
+// generate calls the Gemini REST API with exponential backoff on transient errors.
+// It tries thinking_budget=0 first (saves tokens / latency); if the model rejects
+// it with an HTTP 400 mentioning "budget" or "thinking", it retries once immediately
+// without the thinking config before entering the normal retry loop.
+func (g *GeminiDetector) generate(ctx context.Context, emailText string) (string, error) {
```
jiashuoz, line 240: **Should-fix — the retry budget can't run under the engine timeout.** `maxRetries` defaults to 3 with 1/2/4s backoff, but the Engine bounds each detector at `EngineConfig{}.Timeout` = `defaultDetectorTimeout` = 5s, threaded in as `ctx`. 1+2+4s plus per-call latency can't fit in 5s, so the 2s/4s retries never fire and the 30s HTTP-client timeout is moot — only the initial call and maybe the 1s retry ever run. Either trim the backoff to fit, or give the Gemini-bearing engine a larger per-detector timeout. (Not a safety bug — on timeout the detector is excluded and heuristics carries — just largely dead code as written.)

internal/piguard/gemini.go
```diff
@@ -0,0 +1,406 @@
...
+	for _, seg := range req.Segments {
+		if seg.Type == SegmentSubject {
+			subject = seg.Content
+		} else {
+			bodyParts = append(bodyParts, seg.Content)
+		}
+	}
+	body := strings.Join(bodyParts, "\n\n")
+	if len(body) > geminiMaxBodyChars {
+		body = body[:geminiMaxBodyChars]
```
jiashuoz: **Should-fix — please double-check multimodal handling; plus a truncation nit.**

1. **Are we passing multimodal data (images) from the email to Gemini?** As written it looks like *no*: `formatEmail` concatenates only text segments into a single text part, and nothing builds an `inlineData` (base64 + mimeType) part for image bytes — so an injection hidden in an image would sail through even though `gemini-3.1-flash-lite` is multimodal. Can you confirm? If images aren't being passed, let's scope how to support it. Note it can't be fixed in the adapter alone: `piguard.Segment` carries `Content string` only and `Extract` never emits raw image bytes (`SegmentImageOCR` is reserved but unused), so the bytes never reach the adapter — supporting it means extending `Request`/`Segment` to carry `[]byte`+mimeType, having `Extract` emit image segments, then building `inlineData` parts here (or Files-API `fileData` for large attachments). Fine as a follow-up, but let's track it explicitly.
2. `body = body[:geminiMaxBodyChars]` slices bytes and can split a UTF-8 rune (→ U+FFFD in the JSON). Truncate on a rune boundary.

internal/piguard/gemini.go
```diff
@@ -0,0 +1,406 @@
...
+			}
+			return "", fmt.Errorf("empty Gemini response (finish_reason=%s)", reason), false
+		}
+		return gr.Candidates[0].Content.Parts[0].Text, nil, false
+	}
+
+	// HTTP 400 from thinking_budget=0 rejection.
+	if resp.StatusCode == http.StatusBadRequest && disableThinking {
+		s := string(respBody)
+		if strings.Contains(s, "budget") || strings.Contains(s, "hinking") {
```
jiashuoz: **Nit — fallback silently re-enables thinking, and a fragile match.** If `thinkingBudget=0` is rejected, `generate` retries with thinking *on* — the opposite of intent. With flash-lite this won't normally trigger, but if "never think" is a hard requirement, treat the rejection as an error instead of retrying with thinking enabled. (This also interacts with moving to Gemini 3, which uses `thinking_level` rather than `thinkingBudget`.) Also `strings.Contains(s, "hinking")` drops the leading char to fake case-insensitivity — prefer `strings.Contains(strings.ToLower(s), "thinking")`.

internal/piguard/gemini.go
```diff
@@ -0,0 +1,406 @@
...
+	apiBase string
+}
+
+// NewGeminiDetector constructs a GeminiDetector. Returns a non-nil error when no
+// API key is available (cfg.APIKey empty and neither GEMINI_API_KEY nor
+// GOOGLE_API_KEY is set in the environment).
+func NewGeminiDetector(cfg GeminiConfig) (*GeminiDetector, error) {
+	key := cfg.APIKey
+	if key == "" {
+		key = os.Getenv("GEMINI_API_KEY")
```
jiashuoz: **Nit — collapse the repeated fallback chain.** The three back-to-back `if key == "" { key = os.Getenv(...) }` blocks read as duplication. A small helper is clearer:
```go
key := firstNonEmpty(cfg.APIKey, os.Getenv("GEMINI_API_KEY"), os.Getenv("GOOGLE_API_KEY"))
if key == "" { return nil, fmt.Errorf("piguard/gemini: no API key ...") }
```

internal/piguard/gemini.go
```diff
@@ -0,0 +1,406 @@
...
+	v.InjectionConfidence = geminiClamp01(v.InjectionConfidence)
+	v.PhishingConfidence = geminiClamp01(v.PhishingConfidence)
+	return v, nil
+}
+
+// geminiScoreFromFlagConf maps a boolean+confidence pair to a positive-class
+// probability. Some Gemini models treat *_confidence as confidence in the boolean
+// verdict rather than P(threat): a false verdict with 0.95 confidence should
+// yield 0.05, not 0.95.
+func geminiScoreFromFlagConf(flagged bool, confidence float64) float64 {
```
jiashuoz, line 465: **Discuss (optional) — score scale mixing.** `geminiScoreFromFlagConf` is sound (`flagged=false, conf=0.95 → 0.05`). The subtlety is downstream: Gemini returns a calibrated probability while heuristics returns a weighted heuristic sum, and `Engine.aggregate` averages them on one 0..1 scale against thresholds tuned for heuristics. The eval's AUC-0.97 operating point won't automatically carry over. Worth a calibration pass (and an argument for expressing "prefer the LLM" via `EngineConfig.Weights` rather than by removing heuristics).

jiashuoz (commented):
Nice, well-structured adapter — it honors `ctx`, fails closed (errors/timeouts are excluded, never fail-open), keeps the API key out of URLs/logs/errors, and ships an `httptest`-based test suite. Build + `go test ./internal/piguard/` pass locally. A few things to resolve before merge; details inline.

**Blocking**
- Inbound-only prompt is wired into the *outbound* screening engine (direction mismatch).
- `GEMINI_EVAL_MODEL` is documented in `config.example.yaml` but read nowhere.

**Should-fix**
- Default model → `gemini-3.1-flash-lite` (already commented separately).
- Retry budget can't complete under the 5s engine timeout.
- Images are never sent to the (multimodal) model — and the piguard contract can't carry them yet.

**Nits**: thinking-disable fallback silently re-enables thinking; `strings.Contains(s, "hinking")`; repeated key-resolution fallback; two near-identical builder funcs.

**Two cross-cutting notes**
1. **Keep the heuristics detector always wired.** It's local and effectively never errors, so it's what makes a Gemini 5xx pass through cleanly (excluded → heuristics still scores → `okCount≥MinOK`, not degraded). The desired *fail-open on total outage* posture is a separate change to the Engine contract (`Aggregate.Action` degraded branch + design doc + tests) and shouldn't ride in on this detector PR.
2. **Consider the official Go SDK.** `google.golang.org/genai` would replace the hand-rolled request/response structs, retry loop, and thinking config, and it tracks Gemini 3's config changes (see the model/thinking comment). The tradeoff is real: `piguard`'s package doc declares it a "stdlib-friendly leaf," so adopting the SDK means consciously updating that invariant. Worth deciding explicitly.

jiashuoz (approved):
Approving. Reviewed the full diff, focused on the security-sensitive paths.

**Verified solid:**
- **Fail-safe, never fail-open.** Detector errors/timeouts are excluded from the aggregate; too few `StatusOK` verdicts → `Degraded` → `forceFloor` → review. A degraded engine can't be read as a benign allow.
- **Adversarial truncation closed.** Per-detector `Result.Truncated` (Gemini's `geminiMaxBodyChars` cap) now folds into the same review floor as extraction-level truncation, so padding a payload past the cap can't hide it with "no finding."
- **Phishing-scoring bug fixed.** Primary Score/Flagged is now `max(injection, phishing)` — a 100%-confidence pure-phishing verdict used to reach only a Category (audit-only) and score 0 in the aggregate → silently delivered. Good catch, good regression coverage.
- **Inbound-only wiring** is correct and guarded: Gemini is excluded from `buildAgentScreenEngine` (outbound) with `TestBuildAgentScreenEngine_HeuristicsOnly` locking it in — its prompt classifies content aimed *at* the agent and ignores `Direction`, so it would false-positive on quoted content and miss egress.
- **API-key hygiene** (header-only, never logged/in errors), env-gated opt-in with no behavior change when unset, `E2A_GEMINI_DETECTOR_ENABLED` kill-switch, and the 10s per-detector timeout sized to fit the retry schedule.
- **Audit attribution** via `DetectorLabel()` + full `PerDetector` raw JSONB is a real improvement over the hardcoded `"heuristics"`.

**One non-blocking observation** (fine to address in a follow-up): the protection-event ID is now `sha256(messageID|source|reason|detectorLabel)` with `detectorLabel = agg.DetectorLabel()`, inserted `ON CONFLICT (id) DO NOTHING` — whose stated purpose (`internal/identity/screening.go`) is *exactly-once recording across MTA retries*. With the detector component previously constant, the ID was stable across retries. Now that an LLM detector's availability is non-deterministic, a re-screened message where Gemini succeeds on one attempt but not another computes two different IDs and inserts two audit rows, weakening that exactly-once guarantee. Bounded (duplicate audit rows, not fail-open or data loss), but worth a note — e.g. keep the ID keyed on a stable detector-set identity rather than only the detectors that happened to return OK on a given attempt.

Nice work — the iteration on the adversarial findings really shows.

PR #360, opened 2026-07-01
feat: reply to and forward the agent's own outbound messages

jiashuoz:
## Problem

`POST /v1/agents/{email}/messages/{id}/reply` (and `/forward`) returned `404 not_found` when `{id}` was an **outbound** message the agent sent — even though `GET .../messages/{id}` returns that same message. Root cause: the reply/forward handlers resolved the target through `Store.GetInboundMessage`, whose query hard-filters `direction = 'inbound'`, while GET uses the direction-agnostic `GetMessageWithContent`.

## Fix

Adopt the mail-client rule: **replying to a message you sent continues the thread to its original recipients** (reply → original `To`; `reply_all` → `To` + original `Cc`), not reply-to-`From` (which is the agent itself → a self-addressed loop).

- **Store** — new `GetRepliableMessage` (any-direction sibling of `GetInboundMessage`, keeps the expiry + held-status exclusions) and `GetMessageByEmailMessageID` (direction-agnostic sibling for the HITL references rebuild). `messages.status` is `NOT NULL DEFAULT 'sent'`, so the `status NOT IN (...)` predicate never goes NULL and a normal sent outbound stays repliable.
- **httpapi** — `loadInbound` → `loadRepliableMessage`; reply recipients branch on direction (`replyRecipients` / `outbound.ReplyRecipientsForOutbound`). An outbound target with **no recorded recipients returns 400**, never a self-send. BCC is never carried. Agent-ownership guard and 404-on-miss preserved. Forward needs no recipient branch (recipients come from the request body).
- **HITL** — `attachReferencesChain` (both the `internal/agent` and `internal/hitlworker` copies) now rebuilds the References chain from an outbound parent too, so a held reply-to-own-message keeps its full chain on approval instead of forking multi-party threads.
- **Spec/SDK/docs** — reply/forward operation descriptions updated; `api/openapi.yaml` + TS/Python SDK bases regenerated; MCP `reply_to_message`/`forward_message` tool descriptions updated to say inbound-or-outbound.

## Decisions locked
1. reply = original `To`, reply_all = `To`+`Cc`, BCC never carried.
2. Outbound target with no recorded recipients → `400 invalid_recipient` (never self-send).
3. New, honestly-named `GetRepliableMessage` dep (legacy `GetInboundMessage` store method retained).
4. Direction-agnostic HITL references rebuild included (correctness-complete).

## Verification
- `make test-unit` (full), `make spec-check`, `make generate-sdk-check` — green.
- DB-backed suites: `internal/identity`, `internal/agent`, `internal/hitlworker` — green.
- New unit/handler tests: outbound recipient helper; reply/reply_all/forward-to-outbound + no-recipient 400 + inbound-unchanged regression (real HTTP); HITL outbound-parent references.
- **Full-stack e2e** (`internal/e2e`, real API + Postgres + SMTP relay): send → GET → reply → reply_all → forward → 404, asserting the reply targets the original recipient, reply_all re-adds the original Cc, the reply inherits the outbound's conversation, forward starts a new thread.
- MCP server `tsc` build — green.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -2870,7 +2870,7 @@ paths:
         - messages
   /v1/agents/{email}/messages/{id}/forward:
     post:
-      description: Forward an inbound message to new recipients; the original is quoted and its attachments are carried over by default. Any attachments[] you supply are added on top of the originals. 202 when held for HITL.
+      description: Forward a message (inbound or outbound) to new recipients; the original is quoted and its attachments are carried over by default. Any attachments[] you supply are added on top of the originals. 202 when held for HITL.
       operationId: forwardMessage
       parameters:
         - in: path
@@ -2964,7 +2964,7 @@ paths:
         - messages
   /v1/agents/{email}/messages/{id}/reply:
     post:
-      description: Reply to an inbound message; recipients/threading are derived from the original. 202 when held for HITL.
+      description: Reply to a message (inbound or outbound); recipients and threading are derived from the original. Replying to a message the agent received targets its sender; replying to a message the agent sent continues the thread to its original recipients (`reply_all` also re-includes the original Cc). 202 when held for HITL.
       operationId: replyToMessage
       parameters:
         - in: path
--- a/internal/agent/attach_references_test.go
+++ b/internal/agent/attach_references_test.go
@@ -10,7 +10,7 @@ import (
 	"github.com/Mnexa-AI/e2a/internal/outbound"
 )
 
-// fakeInboundLookup is a stub that satisfies hitlInboundLookup so we can
+// fakeInboundLookup is a stub that satisfies hitlParentLookup so we can
 // drive attachReferencesChain without a live Postgres. Each test sets up
 // the fields it cares about; everything else is zero-valued.
 type fakeInboundLookup struct {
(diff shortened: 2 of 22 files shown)


jiashuoz:
## Independent + adversarial review — disposition

Ran two independent read-only review passes against the pushed branch.

**Adversarial** (6 failure classes, all REFUTED): tenant/cross-agent isolation (the id-only store lookup is correctly gated by the handler's `msg.AgentID != ag.ID` guard + `resolveOwnedAgent` — per-agent, not just per-tenant), self-send loops (4-layer defense), NULL/held-status edges (`status` is `NOT NULL DEFAULT 'sent'`; held/pending/rejected/expired excluded), header injection (`mail.ParseAddress` re-validation; stored recipients clean at write time), idempotency, and HITL parent resolution. No security bugs.

**Independent** (design fidelity + correctness): found one real **MEDIUM** correctness gap →

- **Fixed (8346a63):** reply-to-outbound emitted no `In-Reply-To`/`References`. Outbound rows have no `email_message_id` (composer omits Message-ID; the relay/SES id is captured into `provider_message_id`), so threading was anchored on an empty field and the recipient's client would fork the thread. `GetRepliableMessage` now selects `provider_message_id`; new `Message.ThreadMessageID()` anchors reply threading on it; `GetMessageByEmailMessageID` matches `email_message_id OR provider_message_id` so the HITL approve-time References rebuild resolves an outbound parent too.
- **Fixed (test):** the httpapi fixture that masked the bug (hand-set `EmailMessageID`) is corrected to a realistic outbound; the e2e now asserts the reply's raw carries `In-Reply-To` anchored on the parent's `provider_message_id`.
- **Fixed (test, Finding #3):** added a cross-agent 404 handler test; added a DB test that `GetMessageByEmailMessageID` resolves an outbound parent by `provider_message_id` (agent-scoped).
- **Noted, not changed:** adversarial low-severity cosmetic nondeterminism in the loopback self-send-twin edge (two rows share a Message-ID; `ORDER BY created_at DESC` deterministically prefers the inbound twin with the raw message) — same-agent/same-thread only, no wrong-tenant/wrong-conversation risk.

**Re-verified after the fix:** full `make test-unit`, DB-backed `identity`/`agent`/`hitlworker`, and the full `internal/e2e` integration suite (real API + Postgres + SMTP) — all green.

PR #364, opened 2026-07-02
feat(plugin/tether): attach files to updates and asks (--attach)

jiashuoz:
## What

Two things, both dogfooded live on the tether session that requested them:

1. **File attachments** on the tether harness's outbound surfaces (requested by Josh over the tether thread itself):
   - `tether.sh update [--attach <file>]...` — repeatable, composes with `--html` / `--text`
   - `tether.sh ask [--attach <file>]...` — attach the artifact a decision hinges on (a diff, a mockup) instead of describing it
2. **A delivery bug found while shipping #1**: the harness anchors every update on its own previous outbound message, but a hosted API that predates reply-to-own-outbound (#360) returns `not_found` for reply-created outbound anchors — so the **second of two consecutive updates with no user reply in between silently never reaches the user**. Updates now fall back to a repliable anchor.

## How

**Attachments** — the e2a API already accepts `attachments[]` (`{filename, content_type, data}`, base64) on the reply route; this is pure harness plumbing:

- **`lib.sh`**: new `t_attach_check` (fail-fast validation: exit 3 = file missing, exit 4 = over cap) and `t_reply_payload` (builds the reply JSON with base64-encoded attachments; MIME guessed from filename, fallback `application/octet-stream`). The payload goes to curl via a temp file (`--data-binary @file`) — **never argv**, since a multi-MB base64 string would blow past ARG_MAX. Reply timeout 30s → 120s for multi-MB uploads.
- **Cap**: 15 MB raw total per send — base64 inflates ~4/3 against the server's 25 MB request limit (`maxRequestBytesSend`), leaving headroom for the body. Past the cap the error says to send a link instead.
- **`tether.sh`**: flag parsing on `update` and `ask`, validation before any encoding or network call, `[N attachment(s)]` in the success line.
- **`SKILL.md`**: documents `--attach` in the runtime flow and the writing-good-emails guidance.

**Anchor fallback**:

- `t_api_reply` returns 3 on a `not_found` anchor (vs 1 for other failures)
- new `t_reply_anchored` tries `last_message_id` → `last_inbound_id` → `intro_id` (deduped); `update` / `ask` / `stop` all send through it
- `start` records `intro_id`; `t_poll_once` records `last_inbound_id`

Plugin version 0.4.5 → 0.4.6 (all three manifests), matching prior tether feature PRs (#356, #357).

## Testing

`tether.sh _selftest` extended and passing:
- payload builder: body + encoded attachment (filename / MIME / base64 round-trip), and no-files → no `attachments` key
- validator: missing file → exit 3; 16 MB file → exit 4

**Live end-to-end**: the PR-announcement email itself was sent through the new code path — its outbound anchor 404'd on the hosted API (the exact bug above), fell back to the user's last inbound reply, and delivered with a PDF attached.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -7,7 +7,7 @@
   },
   "metadata": {
     "description": "e2a plugins for Claude Code — authenticated email for AI agents",
-    "version": "0.4.5"
+    "version": "0.4.6"
   },
   "plugins": [
     {
--- a/.cursor-plugin/marketplace.json
+++ b/.cursor-plugin/marketplace.json
@@ -6,7 +6,7 @@
   },
   "metadata": {
     "description": "e2a — authenticated email gateway for AI agents (MCP server + operate-well skill).",
-    "version": "0.4.5"
+    "version": "0.4.6"
   },
   "plugins": [
     {
--- a/plugins/e2a/.claude-plugin/plugin.json
+++ b/plugins/e2a/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "e2a",
   "displayName": "e2a",
-  "version": "0.4.5",
+  "version": "0.4.6",
   "description": "Authenticated email gateway for AI agents — per-agent inboxes, HITL approval, SPF/DKIM verification, and a queryable, replayable event log. 37 MCP tools over hosted streamable HTTP with OAuth.",
   "author": {
     "name": "Mnexa AI",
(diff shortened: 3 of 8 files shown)


jiashuoz:
Third commit pushed (33689bc): while running this branch live, the listener exited `REPLY_RECEIVED` with empty output on a transient curl failure — `t_poll_once` failing returns `""`, which `!= "(no new replies)"` read as a reply. The same pattern in `ask`'s wait loop would hand back an empty string as the user's answer. Both now treat empty poll output as "keep waiting." Observed and verified on the live tether session dogfooding this PR.

Issue #367, opened 2026-07-02
tether skill: Windows breakage + onboarding friction (findings, live-use bugs, and a proposed `tether.sh setup`)


cyj-git-0825:
> Consolidates and supersedes #362. Bundled into one issue (bug + UX findings + a proposed fix) for easier triage — happy to split or send a PR if you prefer.

## Summary

Tested the **tether** skill hands-on end-to-end (Windows/Git Bash, hosted `api.e2a.dev`, plugin `e2a@e2a` 0.4.5), from a true zero state through a live tethered session. One hard blocker on Windows, one hosted deploy-lag, and several onboarding-UX rough edges. I prototyped a fix + a one-command `tether.sh setup`; details and a reference implementation below.

**Environment:** Windows 11, Git Bash (GNU bash 5.2.37, MSYS), `python` 3.13.5, curl 8.11.0.

---

## 1. Blocker — hardcoded `python3` breaks tether on Windows

Every script hardcodes `python3`. On a default Windows install that's the Microsoft Store **App Installer redirector shim**: it's on `PATH` (so `command -v python3` succeeds) but exits non-zero without running; the real interpreter is `python`.

```console
$ python3 -c 'print(1)'   # exit 49, no output  (the shim)
$ python  -c 'print(1)'   # 1                    (Python 3.13.5)
```

Because the scripts use `set -uo pipefail` **without `-e`**, each failed `python3` call returns empty and execution limps forward. Result: a configured `start` fails **silently with a misleading message**:

```console
$ tether.sh start me@example.com --for 2h
tether: intro send failed (check creds / base url / agent protection)
```

None of creds/URL/protection is the cause — `t_duration_to_expiry`, `t_state_set` (**no state file written**), and the send payload builder all no-op'd first. Even with a valid key, `start` can't work: the state file `update`/`poll`/`listen` rely on is never created.

**`_selftest` gives a false green** — it prints `# syntax OK` on this broken machine because it only runs the unconfigured `status` path (no `python3`) + `bash -n`; it never does a real interpreter round-trip.

**Suggested fix** (resolve once, verify it actually executes):
```bash
t_resolve_python() {
  local c
  for c in "${E2A_PYTHON:-}" python3 python; do
    [ -n "$c" ] || continue
    if command -v "$c" >/dev/null 2>&1 \
       && "$c" -c 'import sys; sys.exit(0 if sys.version_info[0]==3 else 1)' >/dev/null 2>&1; then
[...]

PR #368, opened 2026-07-02
feat(plugin/tether): one-command `tether.sh setup` (auto inbox + least-privilege key + HITL off)

cyj-git-0825:
Adds `tether.sh setup` so a new user goes from signed-in to tethered in **one command**, instead of ~8 manual steps across the dashboard, MCP, and a hand-edited config file.

### What `setup` does
Given an account credential (`e2a login` session, or an `e2a_acct_` key), it:
- ensures an agent inbox — reuse an existing one, else `POST /v1/agents` a `tether-<rand>@<shared_domain>`
- turns HITL **off** (`GET`+merge+`PUT /protection`)
- **mints a least-privilege `e2a_agt_` key** bound to that inbox (`POST /v1/account/api-keys`, `scope=agent`) and stores *that* — not the broad account key — in `~/.e2a-tether.env`

`SKILL.md` gains a fast-path and a runtime step 0 that runs `setup` when `status` shows the config missing.

### Before → after
```
before:  e2a login → mcp add/OAuth → create agent → dashboard: mint key by hand
         → turn HITL off → hand-edit env (full agent address) → /tether
after:   e2a login → /tether (setup) → give email + duration → tethered
```

### Also included (prerequisite)
The setup helpers depend on it: resolve the Python interpreter instead of hardcoding `python3`, so the scripts run on Windows/Git Bash where `python3` is the non-functional MS Store shim. `_selftest` now does a real interpreter round-trip (no more false "syntax OK").

### Tested
End-to-end from a true zero state (0 agents, 0 keys) on Windows/Git Bash against hosted `api.e2a.dev`: `setup` created the inbox, minted a scoped key (`whoami` → `scope=agent`, bound to the inbox), wrote the env, and `start` delivered to a real Gmail.

Refs #367

### DIFF
--- a/plugins/e2a/skills/tether/SKILL.md
+++ b/plugins/e2a/skills/tether/SKILL.md
@@ -47,6 +47,25 @@ the two directions use different mechanisms:
 > to `python`. Override with `E2A_PYTHON=/path/to/python`. Run
 > `tether.sh _selftest` to confirm the interpreter actually works.
 
+### Fast path (recommended)
+
+If you have account creds — either from `e2a login` (browser OAuth, nothing to
+paste) or an `e2a_acct_…` key in `~/.e2a-tether.env` — one command does the
+whole bootstrap, no dashboard trip and no MCP server required:
+
+```bash
+"${CLAUDE_PLUGIN_ROOT}/skills/tether/tether.sh" setup
+```
+
+It resolves your key, ensures an agent inbox (creating `tether-…@<shared_domain>`
+if you have none), turns HITL **off**, writes `~/.e2a-tether.env` with the full
+agent address, and confirms with `status`. Then go straight to **Runtime flow**.
+Flags: `--email you@yourdomain` to name/create a specific agent, `--new` to force
+a fresh one. The manual steps below are the fallback — e.g. when you want a
+narrower **agent-scoped** key instead of the account key `setup` reuses.
+
+### Manual setup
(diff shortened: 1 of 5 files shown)


jiashuoz:
Thanks a lot for this, @cyj-git-0825 — the one-command `setup` framing (ensure inbox → HITL off → mint a least-privilege `e2a_agt_` key and store *that* instead of the account key) was exactly the right instinct, and the Windows/Git Bash interpreter-resolution fix was a real bug worth catching.

We ended up building the tether harness out along the same lines in #369 (shipped as `@e2a/cli` 1.6.0 + plugin 0.5.0), where the CLI became a hard dependency and the setup flow moved behind `e2a` subcommands (`whoami` → `login` → `agents create` → `protection set` → `login --agent`). Because that overlaps with what this PR does and the branch has since gone stale against `main` (conflicts), I'm going to close this out — not because the work wasn't good, but because it's been superseded.

A few things we tightened in the golden version that are worth a look for future PRs:
- **protection clobber**: the `GET`+merge+`PUT` must bail if the `GET` fails, otherwise a transient error writes a blank protection config.
- **key fallback**: don't silently fall back to the broad account key when the scoped mint fails — fail loud instead.
- **inbox creation**: `--email` should create/verify the inbox rather than assume it exists.

Really appreciate the contribution — hope to see more. 🙏

PR #369, opened 2026-07-02
feat(cli+tether): harness primitives, login refinement, and the golden tether rewrite on the CLI

jiashuoz:
## What (consolidated per review discussion)

One reviewable unit: the full CLI scripting surface **plus the tether harness rewritten on top of it** — tether is the integration test for the CLI, per the review plan.

### 1. CLI scripting primitives + exit-code contract
`whoami`, `send`/`reply` (`--html-file` with derived text fallback, `--attach`, `--idempotency-key`, `--conversation-id`), `messages list/get` (TSV/NDJSON, `--read-status` defaulting to `all` so poll loops can't lose read messages), with the stable contract:

| code | meaning |
|---|---|
| 0 | success |
| 1 | transient (network / 5xx / rate limit) — retry may help |
| 2 | usage error (unknown flags and `--flag=value` are rejected loudly) |
| 3 | send HELD (any non-`sent` status — open-set safe) — not delivered |
| 4 | bad credentials / wrong scope |
| 5 | permanent request error (404/409/422) — do NOT retry |
| 6 | bounded wait (`listen --once --until`) expired unmatched |

### 2. Bootstrap surface
- `agents list/create/get` (list closes the discovery gap found by the journey review)
- `protection get/set --outbound-review on|off` — GET-merge-PUT that only touches named knobs and **never writes after a failed read** (kills the config-clobber bug class from #368 by construction)
- `keys create [--agent <inbox>]/list/delete` — non-interactive least-privilege key minting

### 3. Login refinement
- `e2a login --agent <inbox>` — browser bootstrap → mint agent-scoped key → **revoke the account bootstrap** (unique-prefix match; refuses to guess); bare slugs expand on the shared domain; typo → owned-agent list, exit 5
- `e2a login --with-key` — headless (arg / `$E2A_API_KEY` / stdin), validated via `/v1/account`, records `key_scope` (which returns with its writer + enum validation)
- `listen --conversation --once --until --text` — the blocking-wait primitive (clean expiry = 6, terminal close = 1, so wake-loops can't spin hot)

### 4. Golden tether rewrite (`plugins/e2a/skills/tether/`)
- **Requires the CLI**: `lib.sh t_cli` resolves `$E2A_CLI` → `e2a` on PATH (version-gated) → `npx -y @e2a/cli@^1.6.0`. Plugin auto-update delivers the skill; npx delivers the CLI — **existing users upgrade with zero action**.
- The entire curl/python API client is deleted; send/reply/poll/body ride the CLI exit-code contract (all of #363's held-send semantics preserved, #364's attachments, #365's HTML/titles).
- `listen`/`ask` wait on `e2a listen --once` in poll-interval chunks: real-time pickup where the WS works, **byte-identical 20s cadence where it doesn't** (hosted `api.e2a.dev` currently 404s the WS handshake — pre-existing gap, degradation verified live). Dedup/parse-retry/ask-lock invariants unchanged: the WS is only a wake signal; consumption stays in the dedup-safe poll.
- **`tether.sh setup`** — the golden bootstrap (`whoami` → ensure `tether-*` inbox → outbound review off → mint agent key → write env). Fails hard everywhere: no account-key fallback, no protection clobber, no non-tether inbox adopted silently, no "ready" on a broken config.

## Testing
- 105 vitest cases (incl. protection merge semantics + the never-PUT-after-failed-GET guard, login `--with-key`, held open-set, `_from`→`from` NDJSON pinning)
- `tether.sh _selftest` PASS (CLI resolution, version gating, `E2A_CLI` override, mutex, durations, attach caps)
- **Live end-to-end against hosted prod**: a real threaded update delivered to a real inbox through the rewritten harness driving the workspace CLI (`E2A_CLI=node cli/dist/bin/e2a.js`); `listen --once` degradation verified against the prod WS 404.
- Earlier rounds: 8-angle adversarial code review (10 findings fixed) + 3-agent interface review (SDK/API/journey — all highs fixed). Findings and fixes in the PR comments.

## Sequencing
Merging is repo-safe immediately. End users get the new tether via plugin auto-update **after** `@e2a/cli@1.6.0` is published to npm (version bumped in this PR; publish via the standard workflow post-merge). Known follow-up: hosted WS endpoint 404 (pre-existing; tether tolerates it by design).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -119,7 +119,7 @@ Layered: generated types → `E2AApi` (raw HTTP) → `E2AClient` (high-level wit
 
 ### CLI (`cli/`)
 
-Commands: login, listen, config. Config stored in `~/.e2a/config.json`. The `listen` command supports `--forward` mode for proxying WebSocket messages to local HTTP endpoints.
+Commands: login, listen, config, whoami, send, reply, messages. Config stored in `~/.e2a/config.json`. The `listen` command supports `--forward` mode for proxying WebSocket messages to local HTTP endpoints. The messaging commands (whoami/send/reply/messages) are the scripting surface for shell-based harnesses and publish a stable exit-code contract (`cli/src/exit.ts`): 0 ok, 1 error, 2 usage, 3 held-for-review (`pending_review` — HTTP-successful but undelivered), 4 auth. Exit codes are frozen once published; add new ones, never renumber.
 
 ### Web (`web/`)
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -119,7 +119,7 @@ Layered: generated types → `E2AApi` (raw HTTP) → `E2AClient` (high-level wit
 
 ### CLI (`cli/`)
 
-Commands: login, listen, config, whoami, send, reply, messages. Config stored in `~/.e2a/config.json`. The `listen` command supports `--forward` mode for proxying WebSocket messages to local HTTP endpoints. The messaging commands (whoami/send/reply/messages) are the scripting surface for shell-based harnesses and publish a stable exit-code contract (`cli/src/exit.ts`): 0 ok, 1 error, 2 usage, 3 held-for-review (`pending_review` — HTTP-successful but undelivered), 4 auth. Exit codes are frozen once published; add new ones, never renumber.
+Commands: login, listen, config, whoami, send, reply, messages. Config stored in `~/.e2a/config.json`. The `listen` command supports `--forward` mode for proxying WebSocket messages to local HTTP endpoints. The messaging commands (whoami/send/reply/messages) are the scripting surface for shell-based harnesses and publish a stable exit-code contract (`cli/src/exit.ts`): 0 ok, 1 transient error, 2 usage, 3 held-for-review (any non-`sent` status — HTTP-successful but undelivered), 4 auth, 5 permanent request error (non-retryable 4xx). Exit codes are frozen once published; add new ones, never renumber.
 
 ### Web (`web/`)
 
--- a/cli/package.json
+++ b/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@e2a/cli",
-  "version": "1.5.1",
+  "version": "1.6.0",
   "description": "CLI for e2a — email for AI agents",
   "bin": {
     "e2a": "./dist/bin/e2a.js"
--- a/cli/src/__tests__/agents.test.ts
+++ b/cli/src/__tests__/agents.test.ts
@@ -1,0 +1,77 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+
+const mockList = vi.fn();
+const mockCreate = vi.fn();
+const mockGet = vi.fn();
+
+vi.mock("../sdk.js", () => ({
+  createClient: vi.fn(() => ({
+    agents: { list: mockList, create: mockCreate, get: mockGet },
+  })),
+  requireAgentEmail: vi.fn(() => "bot@agents.e2a.dev"),
+}));
+
+const AGENT = {
+  id: "agt_1",
+  email: "tether@agents.e2a.dev",
+  name: "tether",
+  domain: "agents.e2a.dev",
+  domainVerified: true,
+  createdAt: new Date("2026-07-01T10:00:00Z"),
+};
+
+describe("agents commands", () => {
+  let mockStdout: ReturnType<typeof vi.spyOn>;
+  let mockExit: ReturnType<typeof vi.spyOn>;
+
+  beforeEach(() => {
+    mockStdout = vi.spyOn(process.stdout, "write").mockImplementation(() => true);
+    vi.spyOn(process.stderr, "write").mockImplementation(() => true);
+    mockExit = vi.spyOn(process, "exit").mockImplementation(() => {
+      throw new Error("process.exit");
+    });
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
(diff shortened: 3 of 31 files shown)


jiashuoz:
## Interface review (3 agents: SDK contract / API contract / user journey) — findings + fixes

Third commit (b0a8eeb) addresses the findings from a dedicated interface review. Highlights, most severe first:

**Poll-loop dead end (both contract agents + journey agent, HIGH).** The server defaults inbound lists to *unread-only*, and `messages get` marks a message read on fetch (`internal/identity/store.go GetMessageWithContent`). So the canonical harness pattern — list → get each id → list again — silently lost every message it had touched. tether's curl layer guards this with `read_status=all`; the CLI now defaults the same and exposes `--read-status unread|read|all`.

**Silent unknown flags (journey, HIGH).** `messages list --conversation-id X` (wrong spelling; list used `--conversation`) was silently ignored → unfiltered mailbox dump, exit 0. Now: per-command flag allowlists reject unknown flags and `--flag=value` forms with exit 2, and `--conversation`/`--conversation-id` are aliases on both commands.

**Env parity with tether (journey, HIGH).** `E2A_AGENT_EMAIL` and `E2A_BASE_URL` — the exact vars tether trains users to export — were ignored by the CLI. Both honored now (`E2A_URL` remains canonical and wins over the alias).

**Open-set status (API contract, MEDIUM).** The spec deliberately does not enum `SendResultView.status` ("tolerate unknown values"). The held check is now inverted (`status !== "sent"` → exit 3), so a future held-variant fails loud instead of exiting 0.

**Double-send on script retry (both contract agents, HIGH/MEDIUM).** The SDK's auto-minted Idempotency-Key only survives in-process retries; a re-invocation minted a fresh key. New `--idempotency-key` on send/reply threads a stable key to the server-side dedupe.

**Retry semantics (SDK contract + journey, MEDIUM).** Permanent client errors (404/409/422 — SDK `retryable: false`) were lumped into exit 1 ("retry may help"). New exit 5 REQUEST splits them; the API error code is printed for grepping. Exit 6 is now the earmark for `listen --once` deadlines.

**NDJSON `_from` leak (SDK contract, MEDIUM).** `--json` output serialized the OpenAPI-Generator escape `_from` — generator-version-dependent. Renamed back to wire-stable `from` on emit, pinned by test.

Also: `--since` documented as inclusive ("at or after"; the server compares `created_at >= since` with second-truncated wire timestamps — cursor loops must dedup by message id, as tether does), `--text` uses `??` so an all-quoted reply's empty parsed text doesn't fall back to the raw quoted body, whoami now shows the config-default agent for account keys, auth errors name `E2A_API_KEY`, and `listen` got the checked flag parsing.

Verified-OK list from the agents (auditable): send/reply/list/get signatures, page-size max 100, AutoPager break semantics (no dangling fetch; maxPages throws rather than truncates), 401/403→exit 4 incl. the WS handshake path, Date→ISO in NDJSON, held statuses on send paths are exactly {sent, pending_review} today, whoami works for both key scopes.

Deferred: `--attach` (tether parity, later PR), `agents list` discovery gap (PR-B), headless paste-key login (PR-C), `rawMessage` volume in `get --json` (documented behavior).

jiashuoz:
## Six-agent review of the consolidated delta → fix batch (441d6a0)

The earlier reviews covered slices 1–3; this round put the remaining two-thirds (bootstrap surface, login exchange, `listen --once`, the tether rewrite) through 4 focused finders + a user-journey re-walk + an adversarial break-it pass. **~38 findings, zero refuted**; all highs/mediums fixed in 441d6a0. The ones worth knowing about:

**Liveness (unattended sessions):** `t_cli` had lost curl's `-m` timeouts — a hung transport froze tether forever; now watchdog-bounded (verified by execution: hung CLI killed in ~3s). Tether's option parsers infinite-looped on a trailing valueless flag (bash `shift 2` no-op — empirically reproduced by the reviewer, now a loud usage error + selftest). A SIGKILLed `ask` left its lock forever, pausing `listen` permanently — locks now carry PIDs and dead-owner locks are reaped.

**Wrong-data:** empty `--conversation` silently dropped the filter (whole mailbox delivered as \"replies\") — now a usage error CLI-side plus state guards tether-side. State writes are flocked + atomic-renamed (torn reads/lost mutations). `listen --conversation` filtered by `get()`, which **marks messages read** — it silently consumed other conversations' messages; now filters via side-effect-free `list()` membership. Empty-body replies (attachment-only/fully-quoted) were retried 120s then dropped — now consumed with a placeholder. TSV From fields sanitized against tab/newline injection.

**Credentials:** `login --agent` saves before revoking (no more zero-credential stranding on a failed config write); `--with-key` stdin now outranks a stale env key (rotations rotate); `setup` mints before flipping protection, with rollback — an aborted setup no longer leaves outbound review disabled while claiming \"nothing written\"; env file written under `umask 077`.

**Also:** `protection off→on` round-trip actually re-enables holding (scan was left off and gate `review` is inert under policy `open`), `--until` validated, `--parallel` marks the primary state so a forgotten `TETHER_STATE` prefix warns instead of silently cross-posting, notify-hook uses the anchored reply chain, deadline timers chain past the 24.8-day `setTimeout` overflow.

**Ordering constraint reconfirmed (critical):** the npx pin `@e2a/cli@^1.6.0` is unpublished (npm latest 1.5.1). Merge → **publish 1.6.0** → plugin release, in that order, or fresh tether users dead-end. Setup's probe now at least surfaces the real npm error instead of misdiagnosing it as a credential problem.

108 CLI tests + tether selftest green. Full agent reports in the session log; deferred lows are noted inline in code comments.

PR #393, opened 2026-07-09
feat(hitl): TTL expiration sweep on River (QueueMaintenance periodic)

jiashuoz:
Migrates the hand-rolled **HITL TTL expiration sweep** from a `time.Ticker(60s)` goroutine onto a River **periodic** on `QueueMaintenance` — retiring the last hand-rolled ticker in the HITL path, mirroring the already-shipped webhook janitor. Pure **consolidation**: the sweep *body* is unchanged; only the *scheduling* moves.

## What & why
`internal/hitlworker` swept `pending_review` messages past their TTL (outbound → auto-send/reject, inbound → release/drop, + `review_approved`/`review_rejected` webhooks) on a hand-rolled 60s ticker in its own goroutine. This moves the schedule to River:
- New `internal/hitlworker/maintenance.go` — `HITLMaintenanceArgs` (`hitl_ttl_sweep`), a `Sweeper` interface, a `MaintenanceWorker` whose `Work` runs the sweep and returns nil (errors logged+swallowed so a DB blip doesn't spin River's retry), and a `jobs.Registrar` scheduling one periodic on `QueueMaintenance` (`RunOnStart:false`, no `UniqueOpts` — same as the sibling periodics).
- `worker.go` — deleted `Run` (the ticker loop), the `interval` field, and `DefaultInterval`; `RunOnce` (the existing single-pass sweep body, already what tests drive) now returns `error`. `New`, `SetPublisher`, both sweep functions, and all auto-send/reject/release + event logic are **byte-for-byte unchanged**.
- `cmd/e2a/main.go` — removed the hitl ticker goroutine + its `hitlCtx`/`hitlCancel`/`workerWG`; the periodic now drains on shutdown via the shared `jobsClient.Stop`. Registrar appended before `jobs.New`.

Cross-replica safe with no `UniqueOpts` (the store transitions are status-guarded compare-and-set / `FOR UPDATE SKIP LOCKED` + a `send_attempts` exactly-once gate), exactly as when the ticker could already run on every replica.

## Scope
**Out of scope (deferred):** routing the TTL auto-approve send onto `QueueOutbound`. The auto-approve still uses the in-process `Sender.Send` — unchanged here.

## Behavior note
`RunOnStart:false` (consistent with every sibling `QueueMaintenance` periodic) means the first sweep is ≤60s after boot instead of immediate — negligible against hour-scale TTLs.

## Verification
- `go build ./...`, `gofmt`, `go vet ./internal/hitlworker/ ./cmd/e2a/` clean.
- `go test ./internal/hitlworker/ ./internal/jobs/` green. New `maintenance_test.go`: error-swallow, happy-path, single-periodic, and QueueMaintenance/args-kind routing. Deleted the obsolete `TestWorkerRunStopsOnContextCancel` (ticker-only).

## Review
Two adversarial review passes (wiring/shutdown + body-preservation/tests) — **both: faithful, safe, ship-able.** Confirmed the sweep body is unchanged, no dangling refs, drain flows correctly through `jobsClient.Stop`, and `outboxPublisher` is wired before the first tick. A stale wiring comment they flagged is fixed in this commit.

**Tracked follow-up (not a blocker):** the old ticker *coalesced* ticks; a River periodic enqueues every 60s regardless, and this sweep is the only `QueueMaintenance` job doing synchronous SES sends (≤100 × up to ~21s under a backlog), so sweeps could briefly pile up on the 2-worker maintenance pool. Bounded and self-healing (each sweep resolves holds, so later ones get cheaper). The clean fix is the **deferred `QueueOutbound` send-routing** — folding it there removes the synchronous send from the sweep entirely.

### DIFF
--- a/cmd/e2a/main.go
+++ b/cmd/e2a/main.go
@@ -306,6 +306,21 @@ func main() {
 	webhookMaint := webhookdelivery.NewMaintenanceJobs(webhook.NewAutoDisableWorker(store))
 	registrars = append(registrars, webhookMaint)
 
+	// HITL TTL expiration sweep as a River periodic on QueueMaintenance — replaces
+	// the hand-rolled 60s ticker. Transitions pending_review messages past their TTL
+	// into review_expired_approved (auto-send/release) or review_expired_rejected per
+	// the owning agent's hitl_expiration_action. Reuses RunOnce as the sweep body.
+	// SetPublisher is called immediately below (before jobs.New, and the registrar
+	// holds this same *Worker pointer), so the publisher is wired well before the
+	// first tick (RunOnStart:false ⇒ first sweep at +60s).
+	hitlWorker := hitlworker.New(store, sender, usageTracker, cfg.OutboundSMTP.FromDomain)
+	// Fire review_approved/review_rejected on TTL auto-resolution, so a hold resolved
+	// by timeout notifies subscribers exactly like a human-resolved one (same legacy
+	// publisher as the agent API). Load-bearing for inbound approve: a TTL-released
+	// inbound message has no other push signal.
+	hitlWorker.SetPublisher(outboxPublisher)
+	registrars = append(registrars, hitlworker.NewMaintenanceJobs(hitlWorker))
+
 	if len(registrars) > 0 {
 		jc, jerr := jobs.New(pool, jobs.Config{}, registrars...)
 		if jerr != nil {
@@ -607,10 +622,10 @@ func main() {
 	// workerWG tracks every background goroutine that needs to drain
 	// before the process exits on SIGTERM. Without it the main
 	// goroutine would return as soon as httpServer.Shutdown returns,
-	// dropping in-flight webhook deliveries and HITL TTL transitions
-	// mid-iteration. retryCancel/hitlCancel signal the workers to
-	// stop; wg.Wait() at the end of shutdown blocks for the current
-	// iteration to settle.
+	// dropping in-flight webhook deliveries mid-iteration. bgCancel/
+	// cleanupCancel signal the remaining workers to stop; wg.Wait() at the
+	// end of shutdown blocks for the current iteration to settle. (The HITL
+	// TTL sweep is now a River periodic drained by jobsClient.Stop.)
 	//
 	// (The HITL approval-notification email is no longer a detached
 	// goroutine — it's a durable River job on QueueNotify, drained by the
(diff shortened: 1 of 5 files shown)


jiashuoz:
## Post-review update

Ran a deeper verification pass: **2 independent adversarial reviews + a live e2e** against a running service.

**e2e (real service, branch binary):** ✅ approve-on-expiry auto-sent to Mailpit + `review_expired_approved`; reject-on-expiry → `review_expired_rejected` (no mail); `hitl_ttl_sweep` river_job `completed` on queue `maintenance`; first sweep at **+64s** (confirms `RunOnStart:false`).

**Correctness:** ✅ both reviewers traced the `send_attempts` exactly-once gate under mid-sweep ctx-cancellation — **no double-send, no double-meter**, even when the job ctx is killed mid-drain.

**One real regression found + fixed (commit `39f8245`):** River applies a default **60s `JobTimeout`**; the old hand-rolled ticker had none. Because the sweep does up to 100 synchronous, retrying, non-ctx-aware SMTP sends, a backlog was being cancelled mid-drain → the store's ctx-derived transitions fail → false **`[hitl-stuck] needs_manual_intervention`** alarms every 60s on the auto-send path, precisely during SMTP/SES degradation. Fixed by overriding `Timeout()` to disable the per-job cap (restores the ticker's unbounded drain) + a test.

**Tracked follow-up (unchanged):** routing each auto-approve send onto its own `QueueOutbound` job makes the sweep DB-only again and would let the timeout return to a bounded value — the structural fix, deferred to its own PR.

Remaining review notes were low/nit (observability asymmetry vs other maintenance workers; `RunOnStart:false` skips a sweep in a sub-60s crash-loop) — neither blocking.

jiashuoz:
## Extensive e2e verification (real running service) — done

Booted the branch binary as a live service against Postgres + Mailpit and drove the sweep end-to-end. Confirmed independently twice (direct DB/log probe + harness report — every number agreed).

**Phase 1 (functional):** all four hold types resolve correctly (outbound approve→Mailpit+`review_expired_approved`; outbound reject→no mail; inbound approve/reject→released/dropped); first sweep at **+63s** (confirms `RunOnStart:false`); `hitl_ttl_sweep` river_jobs `completed` on queue `maintenance` at ~60s spacing; idempotent across 3+ cycles; no alarms.

**Phase 2 — the JobTimeout fix, A/B with an 8s/msg SMTP tarpit (12 approve holds ⇒ ~80–96s sweep):**

| | resolved in one sweep | job wall-time | `[hitl-stuck] needs_manual_intervention` | ctx-deadline errors |
|---|---|---|---|---|
| **Fixed** (`39f8245`) | **12 / 12** uninterrupted | **80s, completed** | **0** | **0** |
| **Pre-fix** (`2ab5a12`) | **7 / 12** (cut at 60s) | ~64s (internally cut) | **5** | **11** |

Pre-fix log (the regression, reproduced):
```
[hitl-worker] auto-approve msg_p2_08: send failed: timeout: context already done: context deadline exceeded
[hitl-stuck] message=msg_p2_08 ... ACTION=needs_manual_intervention   (×5)
```
The fixed binary shows none of this — a >60s sweep runs to completion. Also observed live: a 2nd periodic tick fired while the 80s sweep was still running, and **both completed safely** (SKIP-LOCKED + status guards), matching the concurrency-safety analysis. Correctness (no double-send/double-meter) held throughout via the `send_attempts` gate.

PR #394, opened 2026-07-09
feat(janitor): hourly cleanup on River (QueueMaintenance periodic)

jiashuoz:
Migrates the hand-rolled **hourly janitor** from a `time.Ticker(1h)` goroutine in `cmd/e2a/main.go` onto a River **periodic** on `QueueMaintenance` — retiring the last hand-rolled background loop, mirroring the webhook auto-disable / inbound-retention periodics. Behavior-preserving consolidation.

## What & why
The janitor prunes expired rows across seven targets (messages, user sessions, webhook deliveries, subscriber deliveries, webhook_events, OAuth rows, idempotency keys). It ran on a hand-rolled 1h ticker in its own goroutine — the last such loop outside River.

- New `internal/janitor` package: narrow prune interfaces (one per store method), an injectable `Metrics` interface, a `Janitor.Sweep(ctx)` that runs all prunes **sequentially** (deliberately not parallel — a janitor must be gentle on Postgres; there's no latency benefit to parallelism) and **continues past any individual prune's error** (`errors.Join`), preserving the exact `metrics.JanitorRowsDeleted` labels and the `oauthStorage != nil` guard. Plus the River layer: `JanitorArgs` (`janitor_sweep`), `MaintenanceWorker` (Work swallows+logs, returns nil), `MaintenanceJobs` registrar → one periodic on `QueueMaintenance`, `RunOnStart:false`, no `UniqueOpts`.
- `cmd/e2a/main.go`: removed the ticker goroutine + `cleanupCtx`/`cleanupCancel`; the periodic drains on shutdown via the shared `jobsClient.Stop`.

## Notable structural change (reviewed, verified safe)
On `main`, `oauthStorage` and `idempotencyStore` were constructed *after* `jobs.New`. Since the registrar must be appended before `jobs.New`, this **hoists those two constructions earlier**. Verified safe: both are pure pointer-wraps needing only `pool`; the OAuth nil-gate (`cfg.HTTP.PublicURL != ""`) is preserved exactly; downstream `oauth.NewProvider` / `api.SetOAuthProvider` / `api.SetIdempotencyStore` still reference the same vars. The `OAuthPruner` is injected via `var oauthPruner janitor.OAuthPruner` assigned only when non-nil, avoiding the typed-nil-interface trap (a nil-deref when OAuth is disabled) — covered by `TestSweep_NilOAuthSkipped`.

## Scope / non-goals
Behavior-preserving only: same prunes, same order, same continue-on-error, same metrics. **No batching/chunking** of the DELETEs (a separate future concern for prod-sized tables). No parallelism.

## Verification
- `go build ./...`, `gofmt`, `go vet ./internal/janitor/ ./cmd/e2a/` clean.
- `go test ./internal/janitor/ ./internal/jobs/` green. Tests: every prune called once; continue-past-error (injects failures mid-sequence, asserts later prunes still run + joined error); nil-OAuth skipped without panic; `Work` swallows Sweep errors; periodic routes to `QueueMaintenance` with the right args/kind.
- Adversarial review: **clean, no correctness bugs**; the main.go reorder verified safe in detail.

## ⚠️ Merge-order note (vs #393)
#393 (HITL TTL sweep) edits the same `main.go` region. A trial 3-way merge shows two conflicts: (A) the registrar-append anchor — **keep both** blocks (purely additive); (B) the shutdown-cancel block — #393 removes `hitlCancel()`, this removes `cleanupCancel()`, so **whoever merges second must drop BOTH**, leaving only `bgCancel()` (a naive resolution leaves a dangling call → compile error). Recommend merging #393 first, then rebasing this onto the new `main`.

### DIFF
--- a/cmd/e2a/main.go
+++ b/cmd/e2a/main.go
@@ -25,6 +25,7 @@ import (
 	"github.com/Mnexa-AI/e2a/internal/idempotency"
 	"github.com/Mnexa-AI/e2a/internal/identity"
 	"github.com/Mnexa-AI/e2a/internal/inboundprocess"
+	"github.com/Mnexa-AI/e2a/internal/janitor"
 	"github.com/Mnexa-AI/e2a/internal/jobs"
 	"github.com/Mnexa-AI/e2a/internal/limits"
 	"github.com/Mnexa-AI/e2a/internal/oauth"
@@ -321,6 +322,29 @@ func main() {
 	hitlWorker.SetPublisher(outboxPublisher)
 	registrars = append(registrars, hitlworker.NewMaintenanceJobs(hitlWorker))
 
+	// Hourly cleanup janitor (expired messages/sessions/webhook delivery
+	// records/webhook events/OAuth rows/idempotency keys) as a River periodic on
+	// QueueMaintenance — replaces the hand-rolled time.Ticker(1h) goroutine that
+	// used to live in the shutdown block. Two of its dependencies (the OAuth
+	// storage and the idempotency store) were previously constructed further
+	// down; they only need `pool`, so they're built here so every janitor dep is
+	// in scope before jobs.New consumes the registrars. Their downstream wiring
+	// (api.SetOAuthProvider / api.SetIdempotencyStore, which need `api`) stays
+	// below. oauthStorage is nil when the OAuth provider is disabled (no
+	// public_url); the janitor skips that pass, preserving the old
+	// `if oauthStorage != nil` guard.
+	var oauthStorage *oauth.Storage
+	if cfg.HTTP.PublicURL != "" {
+		oauthStorage = oauth.NewStorage(pool)
+	}
+	idempotencyStore := idempotency.NewStore(pool)
+	var oauthPruner janitor.OAuthPruner
+	if oauthStorage != nil {
+		oauthPruner = oauthStorage
+	}
+	cleanupJanitor := janitor.New(store, deliveryStore, subscriberStore, webhookOutbox, oauthPruner, idempotencyStore, metrics)
+	registrars = append(registrars, janitor.NewMaintenanceJobs(cleanupJanitor))
+
 	if len(registrars) > 0 {
 		jc, jerr := jobs.New(pool, jobs.Config{}, registrars...)
 		if jerr != nil {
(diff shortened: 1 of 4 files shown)


jiashuoz:
## Review + e2e verification — done

**Two adversarial reviews** (janitor package + the rebase resolution):
- **Rebase resolution: CORRECT AND SAFE** — both registrars register, the oauthStorage/idempotencyStore hoist holds, shutdown reduced to only `bgCancel()`, #393 content intact; build/vet/test green.
- **Janitor package: clean, faithful migration** — all 7 prune signatures match, behavior/metrics/logs preserved, continue-on-error safe. Two worth-fixing items, both addressed in commit `ca24703`:
  - Metric emission was untested (dead fake sink) → added `TestSweep_EmitsMetricsForCorrectTables` asserting exactly `{messages,3},{webhook_subscriber_deliveries,5},{webhook_events,7}` fire in order and the other 4 prunes emit nothing.
  - Documented the two inherited/deferred tradeoffs (unbounded single-statement DELETEs; possible >1h-sweep overlap with no UniqueOpts — safe, idempotent DELETEs on distinct tables).

**Live e2e** (real running service, on-demand `janitor_sweep` river_job trigger):

| Target | Before | After | Result |
|---|---|---|---|
| messages | 5 (3 expired) | 2 | deleted 3, kept 2 fresh |
| user_sessions | 3 (2 expired) | 1 | deleted 2, kept 1 |
| idempotency | 4 (2 completed-old) | 2 | deleted 2, kept fresh-completed + old-`in_progress` (selectivity) |

Log confirmed the metric path fires with the right label/count:
```
Cleaned up 3 expired message(s)
[metrics] event=janitor.delete table=messages count=3
Cleaned up 2 expired user session(s)      (no metric — correct)
Swept 2 idempotency key(s) past TTL        (no metric — correct)
```
Job `completed` on queue `maintenance`; a 2nd sweep was a clean idempotent no-op (counts stable, 0 errors/panics).

jiashuoz:
## Complete e2e (all 7 prunes) — passed

Followed up the initial e2e with full coverage against a running service: seeded expired + fresh rows across **all 7 prune targets** and triggered the sweep.

| Prune | Result |
|---|---|
| `messages` | deleted 3 · metric `messages`=3 |
| `user_sessions` | deleted 2, kept 1 |
| `webhook_deliveries` | deleted 2, kept 1 |
| `webhook_subscriber_deliveries` | deleted 2 · metric `webhook_subscriber_deliveries`=2 |
| `webhook_events` | deleted expired non-pending · metric `webhook_events`=3 |
| OAuth (auth_codes + pkce) | `Cleaned up OAuth rows: codes=1 pkce=1` |
| `idempotency_keys` | swept 2, kept fresh + `in_progress` (selectivity) |

All three metric-emitting prunes fired live; expired rows remaining across all targets = 0 after the sweep; 2nd sweep idempotent; 0 errors/panics. Periodic machinery confirmed live via the sibling `QueueMaintenance` periodics firing during the run (`hitl_ttl_sweep`, `webhook_reconcile`, `webhook_maintenance`); the `janitor_sweep` worker was driven directly (registration covered by unit test + wiring review).

PR #408, opened 2026-07-10
feat(ratelimit): exempt system/internal account classes from request limiters

jiashuoz:
## Problem

The in-memory request limiters (per-IP agent registration = 200/hr, per-user poll = 60/min) are **not account-class-aware**, so trusted first-party traffic — the **system**-class prober and the **internal**-class conformance suite — competes for the same budget as anonymous/real-user traffic.

Worse, `regLimit` keys on the connection IP (`CF-Connecting-IP`, else `RemoteAddr` — never a forgeable XFF). Behind a **non-Cloudflare** proxy (staging is CF **DNS-only**), that's the *proxy's* IP, so **all internal traffic shares one bucket** and the conformance gate exhausts the 200/hr registration budget in a single run.

## Change

- **`usage.RateLimited(class)`** — `system` + `internal` bypass the limiters, mirroring (but deliberately **narrower** than) the metering policy: `demo` stays limited (user-facing), and unknown/empty **fail-closes to limited** (a misclassified account is never silently exempted).
- The account class is **loaded onto the principal at auth** — one extra column on the existing `GetPrincipalByAPIKey` join, hidden from API JSON — so the exemption **follows the account** regardless of source IP/proxy, at zero extra query cost.
- Gated in the httpapi rate-limit middleware for the **poll + reg** limiters (principal reused downstream). The **anonymous per-IP limiters** (OAuth DCR, attachment-download) are **unchanged** — they run pre-auth with no account and are the real abuse surface.

## Why it's safe

- Standard accounts are still limited (existing `TestPollRateLimited`/`TestRegRateLimited` pass unchanged).
- The XFF anti-spoof contract (`TestClientIPIgnoresXForwardedFor`) is untouched.
- Server-internal; **no `/v1` or SDK surface change**.

## Tests

- `usage.RateLimited` table (system/internal exempt; standard/demo/unknown limited).
- httpapi middleware: `TestPollRateLimitExemptInternalClass`, `TestRegRateLimitExemptSystemClass`.
- Full `internal/usage`, `internal/httpapi`, `internal/identity` (auth-path) suites green.

Enables the staging conformance gate + hardens the prober against its own rate limits.

### DIFF
--- a/internal/httpapi/ratelimit.go
+++ b/internal/httpapi/ratelimit.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/Mnexa-AI/e2a/internal/identity"
+	"github.com/Mnexa-AI/e2a/internal/usage"
 	"github.com/danielgtaylor/huma/v2"
 )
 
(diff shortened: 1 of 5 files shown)


jiashuoz:
**Review complete (constructive + adversarial).** Core security property **holds**: no API path can self-elevate `account_class` (only the offline seed CLI writes it), fail-closes in every direction, auth-query correct. Fixes applied (4d74640):
- Dropped the `s.deps.Authenticator != nil` pre-check on the reg path — branch on the `resolvePrincipal` error instead (it prefers `PrincipalAuthenticator`).
- Corrected the `RateLimited` godoc: governs only the reg + poll limiters; the per-agent **send** limiter is intentionally not class-exempt.
- Documented the accepted trade-off the adversarial pass flagged: resolving the principal on the reg path costs a keyed `api_keys` touch even on an over-cap attempt (origin/main rejected those IP-only). Accepted — the exemption must be bucket-independent behind a shared-IP proxy, the caller already holds a valid credential, and the write lands only on the caller's own key row.

PR #409, opened 2026-07-10
test(conformance): staging seed-conformance + de-drift e2e-prod to current API

jiashuoz:
Makes the black-box `tests/e2e-prod` suite gate-able against staging. Two pieces:

## 1. `e2a-prober seed-conformance`

A new prober subcommand that provisions the **persistent conformance account** on a target DB (reuses the prober image + identity store):
- **internal**-class user (unmetered → the suite's message churn never trips the monthly quota),
- a **high `account_limits` row** (count-based caps — agents/domains — get headroom for the suite's create/delete churn; staging defaults to 3 agents / 1 domain),
- the shared domain (find-or-create), a **primary agent**, and an **API key**.

Prints `E2A_AGENT_EMAIL` / `E2A_SHARED_DOMAIN` / `E2A_API_KEY` for the pipeline to capture. (Pairs with #408, which exempts the internal class from the request rate limiters so the suite can run without exhausting the per-IP registration budget.)

## 2. De-drift `tests/e2e-prod` to the current API

Because the suite isn't in CI, it had **silently drifted** — **23/135** tests failed against a live server. **All were stale-suite** (server matches `api/openapi.yaml`), **zero regressions**. Fixes:

- **List envelope** `{items, next_cursor}` (was `{agents}` / `{domains}` / `{conversations}` / `{events}`).
- **DELETE requires `?confirm=DELETE`** — including the **cleanup harness** (the cause of leaked test agents).
- Wrong-method is **404 `text/plain`**, not 405.
- `events` **`limit>200` → 422**; `forward` requires **`[to, body]`**.
- Message labels moved **`{labels}` → `{add_labels, remove_labels}`**; removed the retired **50-label cap** and the **`PATCH /v1/domains`** tests (no `updateDomain` op exists).
- The **401-body anti-oracle** check now compares structural error (`code`+`message`) since the envelope carries a per-request `request_id`.

Typechecks (`tsc`); each suite verified green against live staging.

## Flagged as server findings (not suite drift — for follow-up)

- **`429` inconsistency:** send-rate-limit `429` omits `Retry-After`, while agent-create `429` includes it.
- **WS:** the 3 positive WebSocket tests need the **real HTTPS endpoint** (undici can't hold the post-upgrade socket over an SSH port-forward); they'll validate in the pipeline against `api-staging.e2a.dev`.

Depends on #408 for a clean full-suite run under the rate limiter.

### DIFF
--- a/cmd/e2a-prober/main.go
+++ b/cmd/e2a-prober/main.go
@@ -35,6 +35,8 @@ func main() {
 	switch os.Args[1] {
 	case "seed":
 		err = cmdSeed(ctx, configFromEnv())
+	case "seed-conformance":
+		err = cmdSeedConformance(ctx, configFromEnv())
 	case "validate":
 		err = cmdValidate(ctx, configFromEnv())
 	case "run-once":
@@ -58,10 +60,11 @@ func main() {
 func usage() {
 	fmt.Fprint(os.Stderr, `e2a-prober — e2a critical-path self-test runner
 
-usage: e2a-prober <seed|validate|run-once|serve>
+usage: e2a-prober <seed|seed-conformance|validate|run-once|serve>
 
 env:
-  E2A_DATABASE_URL          Postgres URL (seed, validate)
+  E2A_DATABASE_URL              Postgres URL (seed, seed-conformance, validate)
+  E2A_CONFORMANCE_AGENT_EMAIL   primary agent for the conformance account (seed-conformance)
   E2A_PROBE_BASE_URL        e2a HTTP base, e.g. http://e2a:8080 (run-once, serve)
   E2A_PROBE_SMTP_ADDR       e2a SMTP listener host:port (run-once, serve)
   E2A_PROBE_AGENT_EMAIL     synthetic probe agent address
--- a/cmd/e2a-prober/seed_conformance.go
+++ b/cmd/e2a-prober/seed_conformance.go
@@ -1,0 +1,106 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/Mnexa-AI/e2a/internal/identity"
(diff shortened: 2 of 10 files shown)


jiashuoz:
**Review complete (constructive + adversarial).** Verdict: solid; the adversarial pass confirmed **zero regressions / all stale-drift** across the de-drift, and verified every changed assertion against the OpenAPI SSOT. One real fix + documented non-goals:

**Fixed (f86c412):** the `?labels=` filter test. The de-drift mis-diagnosed a **live** 50-value filter cap as "retired" — the server *does* cap it (`maxLabelsPerOp`, `messages.go` → 400 `invalid_filter`). The old rewrite only saw 200 because it sent 51 **repeated** `labels=` params, which collapse to one value under the `explode:false` param. Restored real coverage: 51 values comma-joined → asserts **400** (verified against staging). The PR-body phrase "removed the retired 50-label cap" was therefore incorrect — the cap is live and the test is restored.

**Flagged server/spec findings (not suite bugs):**
1. **OpenAPI gap:** the `?labels=` filter cap (50) is enforced by the server but **not documented** in `api/openapi.yaml` (the `listMessages` `labels` param has no `maxItems`). Add `maxItems: 50`.
2. **Swallowed 405:** wrong-method on a `/v1` path returns 404 `text/plain` (chi detects `MethodNotAllowed`, then hands to the legacy mux which erases the 405+`Allow`). Spec is silent, so 404 doesn't violate the contract — but it's a REST-quality gap, not "correct." The suite pins 404 with a comment saying so.

**Documented non-goal (both reviewers):** the conformance account is `internal`-class + huge caps, so it is **unmetered and rate-limit/quota/count-cap exempt** — the green gate does **not** attest limit/quota enforcement. Intentional (the suite's churn would trip staging's 3-agent default), tracked as a follow-up: a separate `standard`, low-cap metered account to cover enforcement.

PR #413, opened 2026-07-10
test(conformance): P1 suites — webhooks/templates/reviews/account/attachments

jiashuoz:
Five new black-box conformance suites, each covering an API group the existing suite didn't — closing the 56-op matrix. All written + verified **green against live staging**.

| Suite | Ops | Tests | Notes |
|---|---|---|---|
| `16-webhooks` | 8 | 9 | CRUD, deliveries, rotate-secret, test |
| `17-templates` | 8 | 13 | user + starter, `from_starter` copy, validate |
| `18-reviews` | 4 | 6 | account review queue (drives a real HITL hold) |
| `19-account` | 8 | 8 + 1 skip | api-keys CRUD (throwaway only), export, suppressions; **`deleteAccount` skipped — never invoked** |
| `20-attachments` | — | 13 | `getAttachment` + raw capability-token download, **real attachment bytes** |

**50 tests, 49 pass, 0 fail, 1 intentional skip; `tsc` clean.** All self-cleaning (inline — the shared cleanup harness is agent/domain-only). Each expectation verified against `api/openapi.yaml` **and** the live server; no de-drift masking a bug.

## Server/spec findings (flagged, pinned in-suite — for OSS tickets)
- **`deleteWebhook` + `deleteTemplate` take no `?confirm=DELETE`** — inconsistent with agents/domains (which require it).
- **`listReviews`** returns a paginated `{items,next_cursor}` envelope but documents no `limit`/`cursor` params and **ignores `?limit`**.
- **`POST /v1/account/api-keys`** parses the body before the auth gate → an unauth request with a bad body is `400`, not `401` (no data leak).
- `validateTemplate` → `200 {valid:false}` (verdict in body); `testWebhook` async (`200` on unreachable, `409` on disabled); non-HTTPS webhook URL → `400`; the attachment download route returns plaintext errors (off-openapi by design).
- **Gate config:** `approve`/send hits `500` on staging's SES-sandbox sink — the gate should set `E2E_SINK_EMAIL` to the **SES simulator** rather than `blackhole+e2e@e2a.dev`.

### DIFF
--- a/tests/e2e-prod/suites/16-webhooks.test.ts
+++ b/tests/e2e-prod/suites/16-webhooks.test.ts
@@ -1,0 +1,266 @@
+import { test, after } from "node:test";
+import assert from "node:assert/strict";
+import { ApiClient } from "../harness/client.ts";
+import { uniqueSlug } from "../harness/fixtures.ts";
+import { fail, info, writeReport } from "../harness/report.ts";
+
+// Black-box conformance for the webhooks API (8 ops) against LIVE staging.
+// Shapes/status codes verified against api/openapi.yaml (the drift-gated SSOT)
+// AND curl-probed on the live server before these assertions were written.
+//
+// The shared cleanup harness only tracks agents/domains, so every webhook this
+// suite creates is deleted inline in a `finally`. deleteWebhook takes NO
+// `?confirm=DELETE` (unlike agents/domains) — a plain DELETE returns 204
+// (probed 2026-07-10; openapi documents no confirm param for this op).
+const SUITE = "16-webhooks";
+const client = new ApiClient();
+
+const HOOK_URL = "https://example.com/e2e-webhook";
+
+interface WebhookView {
+  id: string;
+  url: string;
+  description: string;
+  events: string[];
+  filters: Record<string, unknown>;
+  enabled: boolean;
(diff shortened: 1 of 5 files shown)


jiashuoz:
**Review complete (constructive + adversarial).** Adversarial verdict: **SOUND — no assertion blesses a server bug**; every pinned behavior verified against OpenAPI + the live server, and all three safety claims hold (`deleteAccount` never invoked, no self-key deletion, self-clean-on-failure). Coverage: **no gaps** (all 41 ops tested; `deleteAccount` soundly skipped).

Applied the one actionable refinement both reviewers raised: `20-attachments` now registers its 6 real-retrieval/capability-token tests with `{ skip }` so a missing SES fixture surfaces as **skipped**, never a no-op **pass** (the one place a regression could've hidden). Plus dropped a dead import. 22/22 verified.

Flagged (not fixed — API/product findings for OSS tickets, pinned in-suite so a regression is caught): `deleteWebhook`/`deleteTemplate` skip `?confirm` (agents/domains require it); `listReviews` ignores `?limit` yet returns `next_cursor`; `POST /account/api-keys` validates body before auth (400 vs 401); inconsistent create-validation error (webhook 422 vs template 400).

PR #416, opened 2026-07-11
feat(identity): add per-agent HITL notification suppression

Amerxu1002:
## Client surface checklist

- [ ] Go handler + integration tests
- [ ] Migration written + idempotent + safe on prod-sized tables (see [CLAUDE.md](../CLAUDE.md))
- [ ] OpenAPI spec + generated types refreshed (`make generate` is clean)
- [ ] TypeScript SDK — generated base regenerated (`make generate-sdk`) **and** resource added to the hand-written `E2AClient` (`sdks/typescript/src/v1/client.ts`)
- [ ] Python SDK — generated base regenerated (`make generate-sdk`) **and** resource added to the hand-written async `E2AClient` (`sdks/python/src/e2a/v1/client.py`)
- [ ] CLI command or flag in `cli/src/commands/` + wired into `cli/src/bin/e2a.ts`
- [ ] MCP tool in `mcp/src/tools/` + registry assertion in `mcp/tests/{tools,http}.test.ts`
- [ ] Tests at each surface above (positive + at least one negative-path / regression case)

**If a row is intentionally skipped**, state why here so the reviewer doesn't have to guess:

> _e.g. "MCP tool deferred — endpoint is admin-only and not useful to an LLM"_

## Test plan

- [ ]
- [ ]


jiashuoz (approved):
Approved with the remaining client-surface, export, and regression-test gaps tracked for an immediate follow-up PR.

PR #421, opened 2026-07-11
feat(prober): mcp_http_round_trip scenario — gate the deployed HTTP MCP surface

jiashuoz:
## Why

The prober battery (`internal/selftest.All`) covers liveness, auth, inbound round-trip, outbound send, loopback, and agent CRUD — but **nothing exercises the deployed streamable-HTTP MCP server** (`e2a-mcp-http`, co-versioned with the server, behind Caddy's `/mcp` route). Today it ships to prod validated only by "the container started." The e2e MCP suites (`tests/e2e-prod/suites/08-mcp`, `12-mcp-extended`) spawn a **local stdio** MCP server — they test the tool logic, not the deployed HTTP transport/auth.

## What

A SmokeSafe `mcp_http_round_trip` scenario making two **read-only** JSON-RPC calls against the real MCP endpoint:

1. **`tools/list`** — proves the `/mcp` route, SSE transport, Bearer auth, and tool registry are live on the candidate image.
2. **`tools/call whoami`** — a genuine end-to-end MCP dispatch (transport → tool handler → e2a backend → back). No mutation, nothing to clean up.

The server is **stateless** (`sessionIdGenerator=undefined`), so each POST stands alone — no `initialize` handshake, no `Mcp-Session-Id`. It authenticates with the probe agent's own Bearer key (the MCP server forwards it to the backend). The response parser accepts either SSE (SDK default) or plain JSON.

## Config

`E2A_PROBE_MCP_URL` (e.g. `http://mcp-server:3000/mcp`). **Unset ⇒ the scenario skips-as-pass**, so a prober on a stack without an MCP endpoint stays green rather than warning (`StatusWarn` would break the deploy bake-gate's `consecutive_green`). The staging deploy sets it; prod need not until we wire it.

## Tests

Skip path, a happy round-trip over a real SSE `httptest` stub, and three failure modes (registry missing `whoami`, 5xx, unreachable). `gofmt`/`vet` clean; `internal/selftest` fully green.

> Note: `cmd/e2a-prober` `TestCmdValidate_Failures` fails on my **local** dirty test DB ("1600 columns" from repeated migration runs) — pre-existing, references none of these changes; CI's fresh DB is the real signal.

## Follow-on (e2a-ops)

Set `E2A_PROBE_MCP_URL=http://mcp-server:3000/mcp` on the staging `prober` service so the release pipeline's prober gate covers real MCP once a release carrying this lands.

### DIFF
--- a/cmd/e2a-prober/config.go
+++ b/cmd/e2a-prober/config.go
@@ -14,6 +14,7 @@ type config struct {
 	AgentEmail    string
 	APIKey        string
 	WebhookSecret string
+	MCPBaseURL    string        // deployed streamable-HTTP MCP endpoint; empty ⇒ mcp scenario skips
 	SinkURL       string        // what the probe webhook targets (== Listen's public addr + /sink)
 	Listen        string        // serve/run-once bind addr for the sink + status server
 	Interval      time.Duration // serve loop period
@@ -28,6 +29,7 @@ func configFromEnv() config {
 		AgentEmail:    os.Getenv("E2A_PROBE_AGENT_EMAIL"),
 		APIKey:        os.Getenv("E2A_PROBE_API_KEY"),
 		WebhookSecret: os.Getenv("E2A_PROBE_WEBHOOK_SECRET"),
+		MCPBaseURL:    os.Getenv("E2A_PROBE_MCP_URL"),
 		SinkURL:       os.Getenv("E2A_PROBE_SINK_URL"),
 		Listen:        envOr("E2A_PROBE_LISTEN", ":8090"),
 		Interval:      envDuration("E2A_PROBE_INTERVAL", 30*time.Second),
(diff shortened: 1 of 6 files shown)


jiashuoz:
Applied review + adversarial findings (both converged on the same two blockers):

- **Host-allowlist 421 (both reviewers, HIGH)** — resolved as a config/coordination fix, not a code change: the probe uses the **in-cluster URL** (`http://mcp-server:3000/mcp`, consistent with every other probe var) and the **staging compose adds `mcp-server` to `MCP_ALLOWED_HOSTS`** (landing in the e2a-ops pipeline stage-2 PR, in lockstep). Chose this over the `https://api-staging.e2a.dev/mcp` Caddy hairpin, which both reviewers noted is fragile (VM self-hairpin + CF/TLS) and inconsistent with the prober's in-container design. Comment updated to state the allowlist requirement.
- **Inert/masked gate (both reviewers, MED-HIGH)** — the ops `E2A_PROBE_MCP_URL` wiring lands **in lockstep** (same stage-2 PR), and the release pipeline's prober gate now **asserts `mcp_http_round_trip` actually ran (status pass, detail not 'skipped')** on staging, so an unset/typo'd var can't leave the gate silently green.
- **SSE `\n` join, stronger whoami assertion, +branch tests** — applied in the latest commit.

Confirmed by both: logic is sound, auth (Bearer agent key) correct, SmokeSafe (read-only), stateless bare tools/call dispatches, no leaks.

PR #435, opened 2026-07-11
BREAKING: standardize every resource's primary key on `<noun>_id`

jiashuoz:
## BREAKING: standardize every resource's own primary key on `<noun>_id`

Converges the inconsistent primary-key naming across the entire `/v1` surface. Previously `MessageView`/`ConversationView` already used `message_id`/`conversation_id`, but `AgentView`, `WebhookView`, `TemplateView`, `EventJSON`, `ReviewView`, `APIKeyView`, `WebhookDeliveryView`, and `AccountUserView` exposed a bare `id`. This renames every resource's **own** primary key to the explicit `<noun>_id` form, coordinated in one batch across server, spec, both SDKs, CLI, MCP, dashboard, docs, and every test.

This is a **breaking wire-contract change**. A partial rename would be worse than none, so it lands together.

### Renamed response/field JSON tags (old → new)

| Resource (view) | old | new |
| --- | --- | --- |
| `AgentView` | `id` | `agent_id` |
| `WebhookView` | `id` | `webhook_id` |
| `WebhookDeliveryView` | `id` | `delivery_id` |
| `TemplateView` / `TemplateSummaryView` | `id` | `template_id` |
| `ReviewView` | `id` | `review_id` |
| `APIKeyView` (+ `CreateAPIKeyResponse`) | `id` | `api_key_id` |
| `AccountUserView` (`GET /v1/account` → `user`) | `id` | `user_id` |
| `EventJSON` (`GET /v1/events`) | `id` | `event_id` |
| Webhook delivery **envelope** (Stripe-style body / `construct_event`) | `id` | `event_id` |

The webhook delivery envelope moved in lockstep with `EventJSON`: the envelope is documented to be the *same shape* as a `GET /v1/events/{event_id}` object so consumers handle both identically — renaming one without the other would break that invariant.

### Renamed URL path params (cosmetic — the URL is unchanged, positional)

`/{id}` → `/{<noun>_id}` for webhooks, templates, events, reviews, api-keys, messages, conversations, and attachments. The matched URL is byte-identical; only the OpenAPI/Huma parameter name changes for consistency.

### MCP tool params

`get/update/delete/rotate/test/deliveries` webhook tools: `id` → `webhook_id`. `get/update/delete` template tools: `id` → `template_id`. (Event/message/review tools already used `event_id`/`message_id`.)

### Deliberately NOT renamed (cross-references & out-of-scope)

- Genuine cross-references already in the target form: a message's `conversation_id`/`agent_id`, `redeliver_event`'s `event_id` + `webhook_id`, event filters, the WS notification's `message_id`, etc.
- The nested `data.agent.id` sub-object inside webhook **event data** payloads (a compact `{id,email,domain}` agent reference, not the `AgentView` resource).
- The GDPR **export blob** (`GET /v1/account/export` → `UserExport` and its `Message` / `AgentIdentity` / `*ExportEntry` sub-shapes) — a bulk data dump, not a REST resource view. Left as its own future batch.
- The **dashboard session user** (`UserInfo` in `web/`, `useAuth().user`). It is populated from `GET /api/auth/me`, which serializes the shared `identity.User` struct (`internal/identity/store.go`, `ID string json:"id"`) — an `internal/auth` / `internal/identity` path deliberately outside the `/v1` `<noun>_id` scope and untouched by this PR. So `UserInfo` stays on `id`. An earlier revision of this branch renamed it to `user_id` to match `AccountUserView`; that broke the Settings → Profile account-ID field at runtime (`/api/auth/me` emits `{"id":…}` → `user.user_id` was `undefined`) while compiling green and passing tests that mocked the fabricated shape. Reverted in a follow-up commit; the frontend field now matches the endpoint it actually calls. (Possible separate follow-up: give `/api/auth/me` a dedicated DTO that also emits `user_id` for cross-endpoint consistency — intentionally NOT done here to avoid touching the multi-path `identity.User` serialization, e.g. GDPR export.)
- DB column names (this is a wire/contract rename only — no schema migration).

### Surfaces updated

- **Server**: view JSON tags + Huma path params → `api/openapi.yaml` regenerated from the handlers (spec-drift gate green).
- **SDKs**: TS + Python generated bases regenerated from the spec via OpenAPI Generator; hand-written ergonomic layers (`WebhookEvent.event_id` in `construct_event`) and all fixtures updated.
- **CLI**: `whoami` / `keys` / `login` field reads.
- **MCP**: webhook + template tool schemas and result handling.
- **Dashboard** (`web/`): types, reads, and test mocks for the `/v1` resource views. The session-user `UserInfo` type is explicitly excluded — see the `/api/auth/me` carve-out above.
- **Docs**: `api.md`, `events.md`, `templates.md`, `data-handling.md`, root `README.md` path placeholders + event code examples.
- **Tests**: Go handler tests, SDK unit tests (TS + Python), MCP tests, CLI tests, the shared contract `scenarios.yaml`, and the e2e-prod suites.

### Verification

- Go: `go build ./...` clean; `go test ./internal/httpapi/...` green; spec golden/no-drift green; envelope round-trip test green. (Pre-existing DB-integration tests in `internal/webhookpub` fail on a dirty local test DB, unrelated — they use the Go `.ID` field / DB `id` column, untouched.)
- TS SDK: build + 118 unit tests green.
- Python SDK: 170 passed / 18 skipped.
- MCP: 155 tests green.
- CLI: 108 tests green.
- Codegen is idempotent (re-running `make spec && make generate-sdk` produces no further diff).


jiashuoz:
Closing in favor of the opposite convergence. After checking how email/developer APIs name a resource's own identifier — Resend, Mailgun, and Stripe all use bare **\`id\`** for an object's own id, reserving \`<noun>_id\` for foreign-key references, and general REST guidance agrees — we're standardizing e2a on **bare \`id\` for own ids** rather than \`<noun>_id\`. That's the smaller, less-surprising change (the codebase is already mostly bare \`id\`; only \`MessageView\`/\`ConversationView\` deviate). A new PR will do the outlier→\`id\` convergence. Thanks to this PR's inventory + the review/adversarial pass (which caught the \`/api/auth/me\` break) for de-risking the decision either way.

PR #437, opened 2026-07-12
ci: auto-trigger e2a-ops staging pipeline on release publish

jiashuoz:
## What

New workflow `notify-staging-pipeline.yml`: on `release: published`, it deploys that exact candidate to the e2a-ops **staging** environment and runs the prober + conformance gates — automatically.

## Why

Release *publishing* stays a deliberate human act (cutting a `v*` tag / GitHub Release). Everything downstream — staging deploy + gating — should be hands-off. This closes the "auto-on-OSS-release" follow-on the e2a-ops pipeline header called out.

## How

- Fires e2a-ops' **existing `workflow_dispatch`** entry point via `gh workflow run` (Actions REST API). So the token needs only **`Actions: write`** on `Mnexa-AI/e2a-ops` — it cannot push code or read secrets — narrower than a `repository_dispatch` (which needs `Contents: write`). Job runs with `permissions: {}` on this repo.
- **Race-safe:** the staging deploy pins `e2a`/`e2a-mcp-http`/`e2a-prober` to `E2A_TAG`, and those images build in parallel with this release event. The workflow **waits for all three manifests** (public GHCR, anonymous `imagetools inspect`) before dispatching, so the deploy never races a not-yet-pushed image.
- **Scope guard:** `if: startsWith(tag_name, 'v')` — product releases only; SDK releases (`ts-sdk-v*`/`python-v*`) start with `ts`/`python` and are skipped.

## ⚠️ Required before this does anything — add the token

This needs a repo secret **`STAGING_DISPATCH_TOKEN`** on `Mnexa-AI/e2a`:
- Fine-grained PAT (or GitHub App token), resource owner `Mnexa-AI`, **only** repo `Mnexa-AI/e2a-ops`, permission **Actions: Read and write** (Metadata: Read is auto-included).
- Until the secret exists, a published release just fails this one step (no release is imminent). Safe to merge ahead of the token.

## Test

`actionlint` clean (incl. shellcheck on the `run` block); YAML validates.

https://claude.ai/code/session_016Btj6LswdbWqSgMavGJzBp

### DIFF
--- a/.github/workflows/notify-staging-pipeline.yml
+++ b/.github/workflows/notify-staging-pipeline.yml
@@ -1,0 +1,71 @@
+# Auto-trigger the e2a-ops STAGING release pipeline when a product release is
+# published here. Keeps release *publishing* a deliberate act (cutting a v* tag /
+# GitHub Release); everything downstream — deploy the candidate to staging and run
+# the prober + conformance gates — happens automatically from this event.
+#
+# Mechanism: this fires e2a-ops' existing `workflow_dispatch` entry point via the
+# Actions REST API (`gh workflow run`), so the token needs only `Actions: write`
+# on Mnexa-AI/e2a-ops (it cannot push code or read secrets) — narrower than a
+# repository_dispatch, which would need `Contents: write`.
+#
+# Token: STAGING_DISPATCH_TOKEN — a fine-grained PAT (or GitHub App token) scoped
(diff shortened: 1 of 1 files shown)


jiashuoz:
Superseded by the hourly self-poll in e2a-ops #182 — that approach needs no cross-repo token (reads public OSS releases with the built-in token), so this event-driven + PAT design is no longer needed.

PR #468, opened 2026-07-13
fix: low-risk hardening across API, SDKs, and MCP (tests + fixes)

XueyanZhang:
## Summary

The low-risk follow-up to #467. A batch of verified, low-risk **fixes + test hardening** across the API backend, both SDKs, and the MCP server — every change written test-first (or, for the two defense-in-depth items, locked with a regression test that was confirmed to fail without the change). **No breaking changes.** Independent of #467 (branched off `main`; the only file both touch is `sdks/typescript/src/v1/ws.ts`, in non-overlapping regions).

Four self-contained commits, one per area.

## 1. Backend hardening — `internal/agentauth`, `internal/webhook`

- **Pin JWT `alg` to RS256** (`agentauth.VerifyToken`). go-jose v3 doesn't allow-list algorithms at parse time — it rejected `alg:none` and the classic *HS256-keyed-on-the-RSA-public-key* confusion only as a side effect of key-type binding. Enforcing the alg explicitly keeps that closed even if the verifying key ever changes. `TestVerify_RejectsAlgConfusion` crafts both attacks (including a *validly-MAC'd* HS256 token keyed on the PKIX public key) and asserts both are rejected; valid RS256 tokens still verify.
- **Lock the v4-mapped-IPv6 SSRF bypass class.** `ssrf_test` adds `::ffff:127.0.0.1`, `::ffff:10.0.0.1`, `::ffff:169.254.169.254`, unspecified `::`, an `fd00` ULA, and bracketed-IPv6 dial-control cases. `IsDisallowedWebhookIP` already blocks these (via `To4()` unwrapping) — this is a pure regression lock against a future refactor reintroducing the bypass.

## 2. Python SDK — retry honors an HTTP-date `Retry-After`

The retry path parsed `Retry-After` only as integer seconds and silently fell back to exponential backoff on the HTTP-date form (RFC 9110 §10.2.3, common behind CDNs). The error layer already parses both via `_parse_retry_after`; `_retry_after_ms` now delegates to it — numeric and dated values honored identically, and consistently with the TS SDK. Regression test asserts a far-future date drives the (clamped) backoff instead of the ~0.1 s exp-backoff.

## 3. TypeScript SDK — WebSocket lifecycle

- **Zombie connection on `close()`.** A `close()` during the reconnect-backoff window left the scheduled `setTimeout` armed, and `dial()` had no closed-guard, so a fresh socket opened after the caller thought the stream was closed. `close()` now clears the tracked timer and `dial()` bails when closed.
- **Unbounded event buffer.** `WSStream` buffered notifications without limit whenever no iterator was awaiting (despite a comment claiming a bound) — a stalled consumer could exhaust memory. Now hard-capped at `WS_MAX_BUFFERED_EVENTS` (1000): oldest un-yielded event dropped (WS frames are advisory — the message stays fetchable via REST) with a one-time loud warning.

Both regression tests were **confirmed red** before the fix (reverted `ws.ts`, watched them fail, restored).

## 4. MCP server — HTTP transport

- **Body limit vs. attachment contract.** `express.json` capped bodies at 1 MB, but the attachment tools advertise up to 10 MB/attachment and 25 MB combined — so any real attachment `413`'d before reaching the tool. Raised to 40 MB (clears the ~34 MB base64-encoded combined cap + envelope); the fronting proxy remains the outer body guard.
- **Stack-trace leak.** No terminal Express error middleware existed, so a post-route throw fell through to Express's default finalhandler, which dumps the error stack — *including absolute internal file paths* — into the response body when `NODE_ENV != "production"` (bin/http never sets it), and never emits a JSON-RPC error. Added a 4-arg handler that logs detail server-side only and returns a generic JSON-RPC internal error (request id preserved).
- **Type drift.** Corrected the webhook-filter type in `client.ts` from the stale `agentIds` to the real wire field `agentEmails` (`mapFilters` already emits `agentEmails`).

## Testing

- **Go:** `go build ./...` + `go vet` clean; `internal/agentauth` and `internal/webhook` unit tests pass (new alg-confusion + IPv6 cases green).
- **Python:** `pytest` — 269 unit tests pass (+1 new).
- **TS SDK:** `vitest` — 152 pass (+2 new), `tsc` clean.
- **MCP:** `vitest` — 175 pass (+2 new), `tsc` clean.

Note: the Go `/v1` handler tests that need Postgres weren't run locally (no DB in this environment); the changed Go code is unit-covered and CI runs the DB tiers.

## Deliberately deferred

The one remaining audit item held for its own PR: **custom-domain agent-create doesn't validate the email local-part** (CR/LF/spaces reach the address; contained downstream by `sanitizeHeaderValue`). It's a behavior change on a live endpoint that warrants DB-backed handler tests, so it belongs in a dedicated change rather than this low-risk batch.

### DIFF
--- a/internal/agentauth/tokens.go
+++ b/internal/agentauth/tokens.go
@@ -113,6 +113,15 @@ func (s *Signer) VerifyToken(token, expectedType, issuer string) (*VerifiedClaim
 	if err != nil {
 		return nil, fmt.Errorf("%w: parse: %v", ErrTokenInvalid, err)
 	}
+	// Pin the signature algorithm to RS256. go-jose does not allow-list algs at
+	// parse time — it relies on key-type binding to reject alg-confusion (an RSA
+	// public key cannot drive an HS256/`none` verifier, so those fail below).
+	// Enforce the alg explicitly so `alg:none` and the HS256-keyed-on-the-public-
+	// key confusion attack are rejected by OUR check, not merely as a side effect
+	// of the key type — closing the gap if the verifying key is ever changed.
+	if len(parsed.Headers) != 1 || parsed.Headers[0].Algorithm != string(SigningAlg) {
+		return nil, fmt.Errorf("%w: unexpected JWS alg", ErrTokenInvalid)
+	}
 	var std jwt.Claims
 	var priv tokenPrivateClaims
 	if err := parsed.Claims(s.priv.Public(), &std, &priv); err != nil {
--- a/internal/agentauth/tokens_test.go
+++ b/internal/agentauth/tokens_test.go
@@ -1,6 +1,12 @@
 package agentauth
 
 import (
+	"crypto/hmac"
+	"crypto/sha256"
+	"crypto/x509"
+	"encoding/base64"
+	"encoding/json"
+	"encoding/pem"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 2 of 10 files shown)


jiashuoz (dismissed):
Approving — **pass with risks**. Reviewed each area against full file context (not just the diff hunks) and verified the load-bearing claims empirically.

## Verified

- **JWT alg-pin (RS256):** `SigningAlg = jose.RS256` confirmed in `signing.go`; pin check is correct. Ran `TestVerify_RejectsAlgConfusion` + `TestSigningAlgIsRS256` → green.
- **SSRF IPv6 lock:** `IsDisallowedWebhookIP`/`guardedDialControl` already block v4-mapped + bracketed v6 via `To4()` unwrapping. Tests green; correctly characterized as a regression lock.
- **Python retry:** `_retry_after_ms` delegates to `_parse_retry_after` (handles HTTP-date via `parsedate_to_datetime`); far-future-date test is deterministic via the `min(retry_after, max_retry_after_ms)` clamp. TS `parseRetryAfter` confirmed to use `Date.parse` — parity claim is real.
- **TS WS:** `close()` clears `reconnectTimer`, `dial()` guards on `closed`, `deliver()` drops oldest at `WS_MAX_BUFFERED_EVENTS=1000` with a one-time warn. Both are genuine bug fixes.
- **MCP body limit + stack-leak fix:** 40 MB clears the 25 MB→~34 MB base64 attachment contract. The stack-leak fix is the one that needed a version check — confirmed **Express 5** (`^5.0.0`), so async rejections from `handleClientRequest` *do* reach the 4-arg error middleware (in Express 4 the fix would silently no-op). `agentIds→agentEmails` rename is clean (matches generated `WebhookFiltersRequest`/`View`, no stale refs).

## Two low-severity notes (non-blocking, can land as-is)

1. **PR body overstates "confirmed red" for the two defense-in-depth items.** I reverted `tokens.go` to main and re-ran `TestVerify_RejectsAlgConfusion` — it *passes* unpatched; go-jose v3.0.5 already rejects `alg:none` + HS256-confusion via key-type binding (exactly as the code comment says). The alg-pin is valuable hardening against a *future* key change, not a live hole. The SSRF tests are pure additions (no `ssrf.go` change), so "fails without the change" doesn't apply there either. Suggest rewording so a future reviewer doesn't think the pin is load-bearing. (The TS WS tests *are* legitimately red-before-green — not disputing those.)

2. **MCP terminal error handler forces 500 on malformed-JSON bodies.** `express.json()` surfaces parse errors with `err.status = 400`; the handler ignores it and returns 500/`-32603`. Low impact for a JSON-RPC server, but honoring `err.status` (or special-casing body-parser's `SyntaxError`) → 400/`-32700` would be more correct.

## Follow-ups

- The deferred **custom-domain agent-create local-part validation** is correctly held for its own PR (behavior change on a live endpoint → needs DB-backed handler tests).
- Release-note-worthy: WS now *drops* oldest buffered events under sustained backpressure (was unbounded memory). Frames are advisory / REST-fetchable, but consumers should reconcile via `list` if they stall.

jiashuoz (approved):
Re-approving after the rebase onto current main. The range-diff preserves the four reviewed commits; conflicts retained main's idempotency retry path and expanded WebSocket iterator tests alongside this PR's changes. All required CI checks pass.

Issue #493, opened 2026-07-16
GA launch: comprehensive API, SDK, CLI, and MCP audit and blocker remediation


jiashuoz:
## Objective

Perform a comprehensive GA-readiness audit of e2a's public developer interfaces and prepare focused draft PRs for every confirmed critical launch blocker.

The goal is to verify that the REST API, TypeScript SDK, Python SDK, CLI, and MCP interfaces are correct, internally consistent, documented, and safe to freeze for GA. This is an end-to-end quality assignment: audit the interfaces, reproduce and classify defects, file findings, and prepare fixes for critical blockers.

## Start here

Read these before testing:

- [GA API/SDK/MCP readiness decisions log](../blob/main/docs/design/ga-api-review.md)
- [API reference and GA contract](../blob/main/docs/api.md)
- [Repository README](../blob/main/README.md)
- [Cross-interface PR checklist](../blob/main/.github/pull_request_template.md)
- TypeScript and Python SDK READMEs
- CLI and MCP READMEs

Treat the generated OpenAPI 3.1 document and the accepted decisions in the GA readiness log as the source of truth. If they disagree, flag the discrepancy rather than guessing.

## Scope

Audit all user-facing behavior in these surfaces:

- **REST API:** routes, methods, request/response schemas, required and nullable fields, status codes, error codes, authentication and scopes, pagination, filtering, idempotency, rate/quota behavior, WebSocket behavior, and OpenAPI accuracy.
- **TypeScript SDK:** generated and hand-written layers, method coverage, parameter and return types, error mapping, pagination helpers, webhook verification, WebSocket/listen behavior, and runtime behavior.
- **Python SDK:** the same coverage as TypeScript, including async behavior and parity with the TypeScript SDK where intended.
- **CLI:** supported GA commands (`login`, `listen`, and `config`), configuration precedence, hosted/self-hosted behavior, forwarding, error output, exit codes, and graceful handling of invalid credentials or connectivity failures.
- **MCP:** tool inventory, names/descriptions/input schemas, account- versus agent-scoped auth, tool-to-API mapping, pagination, returned content, validation, error translation, and hosted Streamable HTTP behavior.
- **Cross-interface consistency:** naming, required fields, enum values, defaults, IDs, pagination, auth semantics, errors, and documented examples must agree across every surface that exposes the capability.

[...]

PR #495, opened 2026-07-16
fix(api): make createApiKey idempotent via Idempotency-Key

cyj-git-0825:
## Summary

`createApiKey` was **not idempotent**: a network-retried `POST /v1/account/api-keys` with the same `Idempotency-Key` silently minted a **second live API key**. Unlike agent/domain/template creates (which 409-dedup on a natural identifier), API keys have no natural uniqueness — `name` is optional — so a retry can't be caught, doubling the caller's live credentials.

This wires the mint through the existing `runIdempotent` guard (the same machinery used by `send`/`reply`/`forward`/`rotate-secret`): a keyed retry with a byte-identical body **replays the first response — the same key** — instead of re-minting. Without a key the mint runs unguarded (idempotency stays opt-in, matching the rest of the surface).

Confirmed the bug live before the fix (two distinct `apk_…` ids from a same-key retry) and gone after.

## Changes
- `internal/httpapi/apikeys.go` — add `RawBody` + `Idempotency-Key` header to `createAPIKeyInput`; wrap `CreateScopedAPIKey` in `runIdempotent` keyed on `(user, "/v1/account/api-keys", body-hash)`.
- `internal/httpapi/apikeys_idem_test.go` — **regression test (fails before, passes after)**: keyed replay mints exactly once and returns the same key; same-key/different-body → `422 idempotency_key_reuse`; distinct keys still mint.
- `api/openapi.yaml` + `sdks/typescript/src/v1/generated/` + `sdks/python/src/e2a/v1/generated/` — regenerated for the new `Idempotency-Key` param (`make spec` + `make generate-sdk`).

## Test plan
- [x] `go test ./internal/httpapi/ -run TestCreateAPIKey` — fails before the fix (`mints=2`), passes after (`mints=1`, same key; different-body → 422).
- [x] `go test -short ./internal/httpapi/` — green (incl. `TestSpecGoldenNoDrift` after regenerating the spec).
- [x] `go build ./...` — clean.
- [x] `generate-sdk-check` — regenerated TS + Python bases; diff is scoped to the new `Idempotency-Key` param on `createApiKey` only.

## Scope / non-goals
Focused on the single blocker per #493. **Not** included (deliberately, to avoid opportunistic scope creep):
- Extending `Idempotency-Key` to the other non-idempotent creates (`createAgent`/`registerDomain`/`createTemplate`/`createWebhook`) — they have natural 409 dedup; can follow separately.
- Teaching the hand-written SDK layer to mint+retry a key for `createApiKey` now that the server deduplicates it (SDK-ergonomics follow-up).

Refs #493 (GA blocker: createApiKey idempotency). Additive/non-breaking — safe to land before or after the v1 freeze.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -3653,6 +3653,13 @@ paths:
     post:
       description: Mint a new API key; the plaintext key is returned once. scope=account is workspace admin (agent/domain/key management); scope=agent binds the key to one inbox so it can act only as that agent. Account scope only.
       operationId: createApiKey
+      parameters:
+        - description: "Optional idempotency key for safe retries (unique per logical request). A retry with the same key and byte-identical body replays the first request's response — the SAME key — instead of minting a second live credential. Completed keys are remembered for at least 24 hours (the published minimum dedup window). Within the window: same key + different body → 422 idempotency_key_reuse (do not retry as-is); same key while the first request is still executing → 409 idempotency_in_flight (wait, then retry unchanged). Dedup is best-effort: under idempotency-store degradation or a mid-request crash the guarantee degrades to at-least-once — a keyed retry may mint a new key rather than replay."
+          in: header
+          name: Idempotency-Key
+          schema:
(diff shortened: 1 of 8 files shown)


jiashuoz (approved):
Reviewed server idempotency behavior and focused regression coverage. Approving per maintainer direction; Python SDK and explicit contract/documentation follow-ups will land separately.

PR #498, opened 2026-07-16
fix(outbound): qualify SES message-id with region domain so replies thread

mattfili:
## Summary

Replies to e2a's **own outbound (SES-sent) messages never thread** — Gmail and every RFC 5322 client fork the reply into a separate conversation. Root cause: SES's SMTP `250` response returns the assigned message id **bare** (`010f…-000000`, no domain), but the `Message-ID` SES actually stamps on the wire is `<010f…-000000@<region>.amazonses.com>`. `parseMessageIDFromResponse` (internal/outbound/smtp_relay.go) wrapped the bare id in angle brackets and stored it as `provider_message_id`; every reply then copied that domainless value verbatim into `In-Reply-To`/`References` — no msg-id match, thread forked. This affects **every e2a→SES reply-to-own-outbound**, not one integrator.

The fix qualifies the id **at capture**, inside the relay adapter: after parsing, a bare id gains `@<region>.amazonses.com`, with the region derived from the standard SES host (`email-smtp.<region>.amazonaws.com`). A new optional `outbound_smtp.message_id_domain` config (env: `E2A_OUTBOUND_SMTP_MESSAGE_ID_DOMAIN`) overrides the derivation for non-standard endpoints; when explicitly set it wins. If neither yields a domain, the id is left as-is — never fabricate one. Downstream propagation (`Message.ThreadMessageID` → `handleReply` → `BuildReferencesChain`) was already faithful and is untouched.

### Evidence (raw Gmail headers, live probe)

Original's real on-wire Message-ID (Gmail "Show original"):
```
Message-ID: <010f019f6ccbcce4-65395574-33bc-4074-b5bd-c3d1f2e82bc9-000000@us-east-2.amazonses.com>
```
The `/reply` against that outbound message went out with:
```
In-Reply-To: <010f019f6ccbcce4-65395574-33bc-4074-b5bd-c3d1f2e82bc9-000000>
References:  <010f019f6ccbcce4-65395574-33bc-4074-b5bd-c3d1f2e82bc9-000000>
```
Missing right-hand side → no match → two separate Gmail threads (confirmed visually and by headers).

### Why the test suite stayed green on this bug

Two fixtures encoded the wrong assumption; both are corrected here:

1. `internal/testutil/fakesmtp.go` returned `250 Ok <id@us-east-2.amazonses.com>` — bracketed **and** domain-qualified, so no test ever exercised the real bare-id path. It now returns the id bare, the way production SES SMTP does.
2. `internal/outbound/smtp_relay_test.go` asserted the bare-id → domainless mapping as correct. The capture path (parse → qualify) is now pinned to the on-wire form in `TestCapturedMessageIDMatchesOnWireHeader`, plus unit coverage for the qualifier and region derivation.

New end-to-end regression (`internal/e2e/reply_threading_ses_e2e_test.go`): against a fake SES returning bare `250` ids, a send must surface a domain-qualified `provider_message_id`, and a reply to that outbound must carry it verbatim in `In-Reply-To`/`References` — asserted both on the stored raw message and on the bytes the fake SMTP received. Verified red without the fix (fails exactly on the bare id) and green with it.

### Alternative considered (not implemented)

e2a could stamp its **own** `Message-ID` in `internal/outbound/compose.go` (which currently omits it) and store that instead of the SES id — fully provider-independent, no region derivation, survives leaving SES. It's a bigger change and needs verification that SES preserves a supplied `Message-ID` over SMTP rather than overwriting it. Worth a follow-up; this PR ships the minimal capture-side fix.

## Client surface checklist

N/A — backend-only. No `/v1` handler change (the fix is inside `internal/outbound`), so no OpenAPI/SDK/CLI/MCP regeneration applies; the spec drift gate (`TestSpecGoldenNoDrift` in `make test-unit`) passes unchanged. No DB migration: `provider_message_id` keeps its column and shape, only the captured value gains the domain the wire always had.

## Operational risk

- Only **newly captured** provider ids change shape. Existing stored bare ids keep working: `CorrelateBySESMessageID` already matches bare, `<id>`, and `<id@…>` shapes, and `LookupConversationID` prefix-matches domainless stored ids by design — both workarounds for exactly this discrepancy, both untouched.
- Replies to *pre-fix* outbound rows still reference the stored bare id (unavoidable — the qualified form was never stored for them); replies to *post-fix* sends thread correctly.
- New config is optional; default behavior (derive from standard SES host) requires no deploy-time action. Non-SES relays are unaffected (no domain derived → id passes through unchanged, as before).
- Rollback: revert the commit; no data or schema to unwind.

## Test plan

- [x] `make test-unit` — includes the new `TestQualifyMessageIDDomain`, `TestSESRegionFromHost`, `TestCapturedMessageIDMatchesOnWireHeader`, and the spec drift gate
- [x] `make test-integration` — full tier against Postgres, with the corrected bare-id fake SMTP
- [x] `make test` — all tiers (`-tags integration -p 1`), green
- [x] Red/green verification: with the qualify call removed, `TestReplyThreadingSESMessageIDE2E` fails on the bare `provider_message_id`; with the fix it passes
- [ ] Post-merge: send + `/reply` against a real SES-backed deploy, confirm one Gmail thread and `In-Reply-To` == the original's on-wire `Message-ID`

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -61,6 +61,12 @@ outbound_smtp:
   username: ""
   password: ""
   from_domain: "relay.example.com"
+  # Domain the provider stamps on outgoing Message-ID headers (SES:
+  # "<region>.amazonses.com"). Used to qualify the bare id SES returns in the
+  # SMTP 250 response so replies thread correctly. Leave empty to derive it
+  # from a standard SES host (email-smtp.<region>.amazonaws.com); set it
+  # explicitly only for non-standard endpoints.
+  # message_id_domain: "us-east-2.amazonses.com"
 
 # Outbound delivery is always asynchronous and at-least-once. The send API
 # durably persists the message and enqueues a River job in one transaction,
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -123,6 +123,15 @@ type OutboundSMTPConfig struct {
 	// off for dev relays (e.g. Mailpit on :1025 with no TLS). Regardless
 	// of this flag, PLAIN auth is never sent over a cleartext connection.
 	RequireTLS *bool `yaml:"require_tls"`
+	// MessageIDDomain is the domain the upstream provider stamps on the
+	// Message-ID header of relayed mail (SES: "<region>.amazonses.com").
+	// SES's SMTP 250 response returns the assigned id BARE — without this
+	// domain — so the relay appends it to make the captured id match the
+	// on-wire Message-ID; a mismatch there breaks In-Reply-To/References
+	// threading on replies to the agent's own outbound. Empty (the default)
+	// derives it from a standard SES Host (email-smtp.<region>.amazonaws.com);
+	// set it explicitly for non-standard endpoints (e.g. VPC endpoints).
+	MessageIDDomain string `yaml:"message_id_domain"`
 }
 
 // InboundConfig selects the inbound processing model (inbound-message-pipeline-
@@ -297,6 +306,9 @@ func Load(path string) (*Config, error) {
 	}
 	if v := os.Getenv("E2A_OUTBOUND_SMTP_FROM_DOMAIN"); v != "" {
 		cfg.OutboundSMTP.FromDomain = v
+	}
+	if v := os.Getenv("E2A_OUTBOUND_SMTP_MESSAGE_ID_DOMAIN"); v != "" {
+		cfg.OutboundSMTP.MessageIDDomain = v
 	}
 	if v := os.Getenv("E2A_INBOUND_MODE"); v != "" {
 		cfg.Inbound.Mode = v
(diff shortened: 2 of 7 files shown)


mattfili:
Worth noting: The domain derivation only fires for a standard SES host (email-smtp.<region>.amazonaws.com). If e2a's prod relay dials a non-standard endpoint (VPC endpoint / custom relay host), sesRegionFromHost returns "", the fix no-ops, and ids stay bare — unless E2A_OUTBOUND_SMTP_MESSAGE_ID_DOMAIN=us-east-2.amazonses.com is set.

jiashuoz:
Thank you so much for your contribution!!

jiashuoz (approved):
Approved. The focused implementation and regression coverage look solid. Non-blocking follow-up: consider recognizing the official SES api.aws and FIPS SMTP endpoint forms in sesRegionFromHost so domain derivation also works without an explicit message_id_domain override on those endpoints.

PR #506, opened 2026-07-17
fix(api): open evolving response vocabularies pre-freeze (keep normalized classifications closed)

jiashuoz:
## The rule

**Response-side vocabularies that can evolve are OPEN sets** — plain strings whose known values are documented in the field description — while **normalized exhaustive classifications and binary invariants remain CLOSED enums.** A closed response enum is a promise the vocabulary can never grow; we should make that promise only where the server actively guarantees it. Pre-freeze this is the last cheap moment to widen: after GA, removing an enum from a response schema is a breaking change for spec-generated clients.

The two-sided rule is now recorded in docs/api.md → "Versioning & stability" (extending the existing "Enums in responses are open" bullet), and each kept-closed field carries a code comment citing it.

## Opened (5 response fields)

| Field | Was | Now |
| --- | --- | --- |
| `APIKeyView.scope` | `enum [account, agent]` | open string, known values documented |
| `CreateAPIKeyResponse.scope` | `enum [account, agent]` (via embedded `APIKeyView`) | open string |
| `AccountView.scope` | `enum [account, agent]` | open string |
| `DomainSuppressionAddedData.source` (stable event payload) | `enum [bounce, complaint]` | open string |
| `LimitExceededDetails.resource` (402 details) | `enum [agents, domains, messages_month, storage_bytes]` | open string |

Notes:
- `DomainSuppressionAddedData.source`: the frozen stable event payload was **stricter than its own surroundings** — the REST `Suppression.source` is already an open string, and the DB CHECK in `migrations/031_delivery_feedback.sql` already admits `source='manual'`. This is a sanctioned pre-freeze widening that aligns the event payload with both. The golden event fixtures are wire bytes and are unaffected (enum tags are schema-only).
- `LimitExceededDetails.resource`: adding a new capped resource must not break clients branching on the 402 details.
- Each opened field uses the house-style description already used across the spec: *"Open set: new values may be added over time, so treat these as strings and tolerate unknown values. Known values: …"*

## Kept closed (deliberately, with comments citing the rule)

- **`EmailBouncedData.bounce_type`** (`permanent | transient | undetermined`) — a normalized, exhaustive classification: `normalizeBounceType` in `internal/delivery/ses.go` maps every provider value into exactly these three, with `undetermined` as the guaranteed catch-all. The vocabulary cannot grow without a deliberate contract change.
- **`direction`** (`inbound | outbound`, on `MessageView` / `MessageSummaryView` / `ReviewView`) — a binary invariant of the model, not an evolving vocabulary.
- **Request-side enums** (e.g. `CreateAPIKeyRequest.scope`) stay closed too — validating and rejecting unknown request values is intentional input validation, not a stability concern (same stance as the existing request-side event-enum drift gate in `internal/httpapi/event_enum_drift_test.go`).

## Regeneration + gates

- `make spec` — `api/openapi.yaml` regenerated; **`TestSpecGoldenNoDrift` green**; `make spec-check` passes.
- `make generate-sdk` — TS + Python SDK bases regenerated (openapi-generator v7.16.0 via Docker); the five generated union enum types (`APIKeyViewScopeEnum`, `AccountViewScopeEnum`, `CreateAPIKeyResponseScopeEnum`, `DomainSuppressionAddedDataSourceEnum`, `LimitExceededDetailsResourceEnum`) collapse to `string`; `make generate-sdk-check` passes post-commit.
- `TestEventPayloadSchemasAreOpen` + eventpayload golden fixtures pass **without** regeneration (enum removal doesn't change wire bytes).

## Test evidence

- `make test-unit`: all packages touched by this change green (`internal/httpapi`, `internal/eventpayload`, and everything else). Two `internal/relay` DB-backed tests (`TestInbound_ProcessIntake_RealPath`, `TestE2E_InboundInjectionHeldOverSMTP`) fail identically on a clean `origin/main` checkout — pre-existing local-DB-state flakes, unrelated to this diff.
- TS SDK: build + **179 tests pass** (incl. type tests).
- CLI: build + **117 tests pass**.
- MCP server: build + **189 tests pass** (incl. type tests).
- Python SDK: **314 passed, 4 skipped** (contract tests excluded — need a live server).

## Consumer-surface sweep

Grepped `mcp/src`, `cli/src`, `web/src`, `internal/e2e`, `tests/contract/scenarios.yaml`, and the hand-written SDK layers (`client.ts` / `client.py` / `errors.*`) for the removed enum types and over-asserted values: **zero references**. `web/src` already types `scope` as `string`; the api-keys page's `"account" | "agent"` union is a request-side form picker, which is correct under the rule.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -69,10 +69,7 @@ components:
         name:
           type: string
         scope:
-          description: account = workspace admin; agent = bound to one inbox.
-          enum:
-            - account
-            - agent
+          description: "account = workspace admin; agent = bound to one inbox. Open set: new values may be added over time, so treat these as strings and tolerate unknown values. Known values: account, agent."
           type: string
       required:
         - id
@@ -102,9 +99,7 @@ components:
         plan_code:
           type: string
         scope:
-          enum:
-            - account
-            - agent
+          description: "Credential scope. Open set: new values may be added over time, so treat these as strings and tolerate unknown values. Known values: account, agent."
           type: string
         upgrade_url:
           type: string
@@ -531,10 +526,7 @@ components:
         name:
           type: string
         scope:
-          description: account = workspace admin; agent = bound to one inbox.
-          enum:
-            - account
-            - agent
+          description: "account = workspace admin; agent = bound to one inbox. Open set: new values may be added over time, so treat these as strings and tolerate unknown values. Known values: account, agent."
           type: string
       required:
         - key
@@ -959,9 +951,7 @@ components:
         reason:
           type: string
         source:
-          enum:
-            - bounce
-            - complaint
+          description: "How the suppression was created. Open set: new values may be added over time, so treat these as strings and tolerate unknown values. Known values: bounce, complaint."
           type: string
       required:
         - address
@@ -1702,12 +1692,7 @@ components:
           description: The account's plan label.
           type: string
         resource:
-          description: The AccountView usage/limits field stem the cap applies to. Key it to usage.<resource> and limits.max_<resource>.
-          enum:
-            - agents
-            - domains
-            - messages_month
-            - storage_bytes
+          description: "The AccountView usage/limits field stem the cap applies to. Key it to usage.<resource> and limits.max_<resource>. Open set: new values may be added over time, so treat these as strings and tolerate unknown values. Known values: agents, domains, messages_month, storage_bytes."
           type: string
         upgrade_url:
           description: An upgrade affordance URL, when the operator has configured one.
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -2345,12 +2345,7 @@ components:
           minimum: 1
           type: integer
         scope:
-          description: Which byte budget was exceeded.
-          enum:
-            - composed_message
-            - attachment
-            - attachments_total
-            - request_body
+          description: "Which byte budget was exceeded. Open set: new values may be added over time, so treat these as strings and tolerate unknown values. Known values: composed_message, attachment, attachments_total, request_body."
           type: string
       required:
         - scope
@@ -2480,11 +2475,7 @@ components:
       properties:
         action:
           default: flag
-          description: "What a gate non-match does: flag (deliver + annotate), review (hold), block."
-          enum:
-            - flag
-            - review
-            - block
+          description: "What a gate non-match does: flag (deliver + annotate), review (hold), block. Open set: new values may be added over time, so treat these as strings and tolerate unknown values. Known values: flag, review, block."
           type: string
         allowlist:
           description: "Addresses (allowlist) or domains (domain) the gate trusts; ignored for open. Inbound: matched against the message From AS PRESENTED — a match does not by itself prove the sender is authentic (a forged From that fails SPF/DKIM/DMARC can still match). For spoofing-sensitive trust, also check the message authentication result."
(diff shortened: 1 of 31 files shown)


jiashuoz:
Review response — all four findings addressed in bbeb9081.

**1. `PayloadTooLargeDetails.scope` (major) — fixed: opened.** Verified the semantics: the value identifies *which byte budget* was exceeded, and the set of budgets is an evolving vocabulary (a future batch or per-template cap adds a value), not a normalized classification. Same case as the sibling `LimitExceededDetails.resource`, so it now carries the house open-set description with known values `composed_message, attachment, attachments_total, request_body`, plus a rule-citing comment.

**2. Protection view enums (minor) — fixed: all four opened.** Confirmed in code that `ProtectionGateView` / `ProtectionScanView` / `ProtectionHoldsView` are pure RESPONSE types with dedicated `*Request` mirrors for the PUT body (kept distinct for the strict-request/open-response stance). `policy`, `action`, `sensitivity`, and `on_expiry` are all evolving config vocabularies, so they're opened on the views; the `*Request` mirrors keep their closed enums (request-side validation is where a closed enum belongs), and the block comment on the Request types now spells out that asymmetry. `on_expiry` was the only judgment call — opened rather than treated as binary, because expiry disposition is config vocabulary (an `extend`/`escalate` is plausible), unlike message `direction` which is a structural invariant. Beta status made a wrong call cheap, but the rule was applied, not skipped — and the PR body's "every kept-closed field has a rule-citing comment" claim is now true.

**3. Stance test (test gap) — added: `TestResponseEnumsAreOpenSets`** (`internal/httpapi/response_enum_stance_test.go`), following the `TestEventPayloadSchemasAreOpen` idiom. It renders the live spec, walks every response-stance component schema (`additionalProperties: true` — the marker stability.go stamps on response-reachable schemas), and fails on any closed enum not on an explicit allowlist, with a failure message citing docs/api.md "Versioning & stability". The allowlist is exactly six entries: `EmailBouncedData.bounce_type` (normalized exhaustive), the three `direction` fields (binary invariant), and the single-value `code` discriminator constants on the typed 402/429 error envelopes. It also rejects inline response enums under `paths/**/responses` (so the component walk can't be bypassed) and flags stale allowlist entries. Negative-tested: deleting an allowlist entry fails with the expected message.

**4. TS/Python hand-written webhook layer — fixed.** `webhook-signature.ts` `DomainSuppressionAddedData.source` widened from `"bounce" | "complaint"` to `string` using the file's own house pattern for open sets (doc comment "Open set; tolerate unknown values. Known values: bounce, complaint." — matching `sending_status` two interfaces up; the SDK doesn't use the `(string & {})` idiom anywhere, so I stayed with the established pattern). Audited the rest of the file: the only other closed union is `bounce_type`, which correctly mirrors the deliberately-closed enum and now carries a rule-citing comment. Python's `webhook_signature.py` already typed `source: str`; its comment now marks it as an open set.

**Gates (all rerun after regeneration):** `make spec` + `spec-check` (`TestSpecGoldenNoDrift`) pass; `make generate-sdk` + `generate-sdk-check` pass (regen touched exactly the four opened models per language + ObjectSerializer); full `make test-unit` green (including the two `internal/relay` tests that were flaky in the original run); new stance test passes; TS SDK 179, CLI 117, MCP 189, Python SDK 314 passed / 4 skipped.

PR #511, opened 2026-07-17
feat(api): bound remaining GA request fields (names 200, reason 2000, conversation_id 200, addresses 320)

jiashuoz:
Closes out the remaining unbounded free-form request fields before the GA contract freeze (follow-on to the outbound per-field caps and the composed-message ceiling on the approve path).

## ⚠️ MERGE GATE — production data scan required

**Do not merge until `scripts/ga-bounds-prod-scan.sql` has been run against the production database and every `violations` count is zero (or each offender has been reviewed and knowingly accepted).** The bounds are request-edge validation only, so existing rows keep reading fine — but a stored value over a new cap means a real user exceeded it and a write path (rename replay, reviewer override, held-draft approve) could start rejecting input that used to work.

- [ ] **RUN AGAINST PROD BEFORE MERGE:** `psql "$PROD_DATABASE_URL" -v ON_ERROR_STOP=1 -f scripts/ga-bounds-prod-scan.sql`
- [ ] All `violations` counts are `0`, or each offender row is listed and accepted in a comment on this PR

The scan counts `char_length()` (Unicode code points) to match the enforcement semantics below, reports max lengths, and samples the top-5 offenders per field. It was executed against the local dev DB (`e2a_test` on :5433) to prove the SQL runs end-to-end (exit 0; all counts 0 on dev data).

## Bounds shipped

| Field | Bound | Declarative (spec + SDKs) | Runtime enforcement | Semantics |
|---|---|---|---|---|
| `CreateAgentRequest.name` | 200 | `maxLength` tag → Huma 422 | `identity.ValidateAgentName` in `createAgent` (store) | code points |
| `UpdateAgentRequest.name` | 200 (pre-existing tag) | `maxLength` tag → Huma 422 | `identity.ValidateAgentName` in `UpdateAgentName` (store) — **fixed from byte-count to rune-count** | code points |
| `CreateAPIKeyRequest.name` | 200 | `maxLength` tag → Huma 422 | dashboard POST /api/keys rejects >200 (400); shared `identity.MaxAPIKeyNameLen` | code points |
| `RejectRequest.reason` | 2000 | `maxLength` tag → Huma 422 | magic-link reject form CLAMPS to 2000; shared `identity.MaxRejectReasonLen` | code points |
| `conversation_id` (send/reply/forward) | 200 | `maxLength` tags → Huma 422 | `validateConversationID` (shared backstop, now length + CR/LF) | code points |
| `CreateAgentRequest.email` | 320 | `maxLength` tag → Huma 422 | schema-only (email syntax/domain checks follow in handler) | code points |
| `reply_to` (send/reply/forward) | 320 | `maxLength` tags → Huma 422 | `validateReplyTo` rune count **before** the address parser | code points |
| `to`/`cc`/`bcc` items (send/reply/forward **and** approve overrides) | 320 per item | `maxLength` on array fields → Huma applies it to the **items** schema and pre-validates each element (422 `invalid_request`) | `agent.ValidateRecipients` rune count **before** `mail.ParseAddress` (shared by every send path incl. legacy) | code points |

All bounds are expressed as named constants: `identity.MaxAgentNameLen` (200, one constant shared by create + update), `agent.MaxAddressLen` (320, shared by schema tags and the runtime recipient check), `httpapi.maxConversationIDLen` (200), `httpapi.maxAPIKeyNameLen` (200), `httpapi.maxRejectReasonLen` (2000). `TestGABoundTagsMatchConsts` guards tag/const drift (struct tags are literals), mirroring `TestOutboundFieldLimitTagsMatchConsts`.

The 320-character address cap applies to the **full string item** (display name + `<addr>` combined) — previously a recipient was only `mail.ParseAddress`-validated, so a multi-megabyte display name passed. Oversized items now fail at the schema edge with the standard 422 `invalid_request` envelope.

## Unicode agreement (bytes vs. code points)

OpenAPI/JSON-Schema `maxLength` counts **Unicode code points**, not bytes. Verified in the vendored Huma source (`huma/v2@v2.38.0/validate.go`): both `MinLength` and `MaxLength` validate with `utf8.RuneCountInString` — i.e. **Huma already enforces code-point semantics**, matching the spec.

Every hand-written Go check on these fields was audited to agree:

- `identity.UpdateAgentName` previously used `len(name)` (**bytes**) while claiming "characters" — a 200-code-point CJK name (600 bytes) would pass the schema and then fail in the store. Fixed: `identity.ValidateAgentName` counts runes, and both create and update call it.
- `agent.validateRecipients`, `validateReplyTo`, and `validateConversationID` count `utf8.RuneCountInString`, checked **before** any parsing so oversized strings are rejected on a cheap length count.
- The message-list `conversation_id` **filter** and `GET /conversations/{id}` previously capped at 200 **bytes** — which would have made a schema-legal 200-code-point non-ASCII conversation_id unfilterable. It now uses the shared rune-count check, preserving the invariant that every accepted conversation_id stays filterable. (The webhook `conversation_ids` filter values are charset-restricted to ASCII `[a-zA-Z0-9_-]`, where bytes == runes, so no change was needed there; the field descriptions promise only that an accepted id is *never too long* to filter by.)

Every bound's field description states the cap in "characters (Unicode code points)". Per-field multi-byte boundary tests (CJK `日`, 3 bytes/code point, **exactly at** each limit and at limit+1) prove spec-vs-runtime agreement end-to-end over the wire: `TestCreateAgentNameUnicodeBoundary`, `TestUpdateAgentNameUnicodeBoundary`, `TestCreateAgentEmailUnicodeBoundary`, `TestCreateAPIKeyNameUnicodeBoundary`, `TestRejectReasonUnicodeBoundary`, `TestSendConversationIDUnicodeBoundary`, `TestSendReplyToUnicodeBoundary`, `TestSendRecipientItemUnicodeBoundary`, `TestApproveOverrideRecipientUnicodeBoundary`, plus direct unit checks on the runtime backstops (`TestValidateRecipientsRuneSemantics`, `TestValidateAgentNameRuneSemantics`, `TestReplyToAndConversationIDBackstopRuneSemantics`).

## No DB constraints — stored data unaffected

Request-edge validation only; no migration. `TestOverlongStoredConversationIDStillReadable` (DB-backed) writes a 5000-char conversation_id via direct SQL and confirms both the detail read and the list read return it intact.

## Compat-gate policy note (reviewer attention)

The new oasdiff GA gate (`--fail-on WARN`) flags first-time `maxLength` settings as `request-property-max-length-set` (WARN). This PR downgrades **that one check** to `info` in `api/oasdiff-levels.txt` — setting a first bound on a previously-unbounded request field is exactly this reviewed pre-GA program. `request-property-max-length-decreased` (tightening an already-published bound) keeps its failing default, so post-GA cap reductions still break the gate. **Restore step at the GA freeze** (oasdiff rejects comment lines, so the marker lives in `docs/api-compatibility-gate.md`): delete the `request-property-max-length-set info` line so first-time maxLength additions fail the gate again.

## Test evidence

- `make spec` regenerated `api/openapi.yaml`; `TestSpecGoldenNoDrift` passes; `make openapi-compat-check` + `make openapi-compat-test` pass.
- `make generate-sdk` (Docker, pinned generator v7.16.0) regenerated both SDK bases; the Python models gain pydantic `max_length` constraints (client-side pre-validation), TS models gain the documented bounds.
- Go: `internal/httpapi` unit suite green (incl. all new tests); DB-backed `internal/identity` + `internal/agent` green on an isolated database (`-p 1`, fresh DB — the shared `e2a_test` DB showed unrelated cross-process contention from concurrent local runs).
- TS SDK: `npm test --workspace @e2a/sdk` — 179/179 passed.
- Python SDK: unit tests pass except `test_validation_field_location_is_required`, which fails identically on a pristine checkout of this base (pre-existing, unrelated).
- Pre-existing, unrelated on this base (verified against a stashed pristine tree): `internal/relay` failures in `make test-unit`.
- Consumer surfaces grepped (`mcp/src`, `cli/src`, `web/src`, `tests/contract/scenarios.yaml`, `internal/e2e`): nothing asserts the old unbounded behavior or sends over-cap values (the one large-string e2e fixture is a 50 KB event *payload body*, not a bounded field).

## Out of scope (follow-ups)

- Local-part charset/length validation for custom-domain agent emails (the scan SQL includes an informational `local-part > 64` count to size that follow-up).
- Fields already bounded by the earlier outbound per-field caps / composed-message ceiling work (subject, text/html bodies, attachments, webhook URL/description/filters, template fields).
- `CreateAPIKeyRequest.agent_email` carries no explicit tag — it must resolve to an existing owned agent (itself now ≤ 320) or the request 404s.


jiashuoz:
## Review findings — dispositions (fixed in 6ab0fd4d)

**1. MAJOR — GET /conversations/{id} counted bytes (VALID, fixed).** `handleGetConversation` kept a `len() > maxFilterStr` pre-check, so a schema-legal 200-code-point non-ASCII conversation_id (600 bytes) 400'd on fetch. Replaced with the shared `validateConversationID` rune-count check (same fix as the list filter). New test `TestGetConversationAcceptsMaxLenUnicodeID`: 200×日 id reaches the store lookup (404), 201×日 is a 400. Sweep for a third byte-count check on bounded fields came back clean — the remaining `len()` caps are all on fields this PR does not bound (`from`/`subject_contains` list filters, the feedback-form email at `internal/agent/api.go`, the user-profile display name in `internal/auth`).

**2. MINOR — two non-/v1 entry paths bypassed the caps (VALID, fixed).**
- *Magic-link reject reason*: now **CLAMPED** to 2000 code points (rune-boundary-safe) rather than rejected — rationale: it's a human-facing HTML form reached from an email; failing the reviewer's whole rejection over a long note is worse than losing the note's tail, and the reviewer has no JSON error surface to react to. The `/v1` API path (machine callers) still rejects with 422. Test: `TestMagicRejectPOSTClampsOverlongReason` (stored reason is exactly the first 2000 code points of a CJK string).
- *Dashboard POST /api/keys*: now rejects a >200-code-point name with 400 `name too long — N characters, max 200`. Test: `TestHandleCreateAPIKey_NameBoundRuneSemantics` (200×日 → 201, 201×日 → 400).
- Both bounds moved to one canonical home: `identity.MaxAPIKeyNameLen` / `identity.MaxRejectReasonLen`, with the httpapi consts aliased. This corrects the PR body's "schema-only (single entry path)" claim for these two fields — every entry path is now covered.

**3. MINOR — compat-gate downgrade needs a restore marker (VALID, fixed).** Tested empirically: oasdiff **rejects comment lines** in the severity file (`failed to load custom severity levels … invalid line #24: # comment test`, exit 106), so no in-file TODO is possible. The restore step is documented in `docs/api-compatibility-gate.md` ("Pre-GA exception — restore at the freeze"): **at the GA cut, delete the `request-property-max-length-set info` line** so first-time maxLength additions fail the gate again. `request-property-max-length-decreased` was never downgraded.

**4. Prod scan gap — reply_to (VALID, fixed).** `messages.reply_to` is `text[]` (migration 005); the scan now unnests it and measures `char_length` per element against 320, same counts/max/top-5 shape as the recipient arrays. Whole script re-ran clean against the local dev DB (`ON_ERROR_STOP=1`, exit 0). The prod merge gate now covers it.

Gates after the fixes: `make spec` → **zero spec diff** (none of these touch the /v1 schema); `make spec-check` green; `internal/httpapi`, `internal/agent`, `internal/auth` suites green (`-p 1`, dedicated isolated DB).

PR #513, opened 2026-07-17
fix(delivery): emit email.failed when SES rejects after acceptance

jiashuoz:
## Root cause

`internal/delivery/ses.go` parses SES `Reject` into `RecipientOutcome{Status: StatusFailed}` and the consumer durably persists the `sent → failed` transition — but `pushEventFor` mapped only delivered/bounced/complained. A post-acceptance Reject therefore flipped the message to `failed` **silently**: event-driven consumers held the earlier `email.sent` result forever unless they polled.

## Fix

A correlated SES Reject now also publishes the canonical stable **`email.failed`** event, reusing `eventpayload.EmailFailedData` unchanged (the exact shape the async send worker emits — no second failed-event shape, **no schema change**, so no OpenAPI/SDK regeneration; `TestSpecGoldenNoDrift` passes untouched).

### Cardinality + dedup design
- **One message-level event per rejected message**, never per recipient: a Reject is SES refusing the whole accepted message (content scan), and `EmailFailedData` is message-level — its `to`/`cc`/`bcc` lists (from the correlated row) preserve the recipient information. Per-recipient `failed` outcomes are still recorded in `message_recipients` as before.
- **Deterministic event id shared with the worker path**: the dedup key is `message_id|email.failed`, which hashes (via `webhookpub.DeterministicEventID`) to the *identical* id the send worker uses. Duplicate SNS deliveries AND any cross-path double emission collapse in the durable outbox via `ON CONFLICT (id) DO NOTHING` — subscribers can never receive two conflicting terminal events for one message. Pinned by `TestRejectEmailFailedIDCollapsesAcrossPaths`.
- `ParseSESNotification` now captures `reject.reason` (e.g. `"Bad content"`) as the recipient detail / event `reason`, with a stable fallback so the required field is never empty. `reason_code`/`retryable` stay unset, matching the worker emission and the committed golden fixtures.
- `CorrelateBySESMessageID` widens to a `CorrelatedMessage` struct — same single SELECT/join, now also carrying `conversation_id`, `method`, `message_type`, `from`, `to/cc/bcc` so the canonical payload is fully populated.
- **Reject never suppresses** (no explicit rule requires it — suppression stays bounce/complaint-only). Bounce, complaint, delivery, and delivery-delay behavior untouched. Uncorrelated/unknown SES ids remain acked no-ops.

### Docs
`docs/events.md`: the `email.failed` guarantee row now names both emission paths honestly — **at-least-once** from the send path (same-tx outbox), **best-effort** via delivery feedback (fire-after-the-fact `OutboxPublisher`) — plus a note on the message-level, at-most-once-per-message id.

## Tests (TDD; all on a dedicated `e2a_test_sesreject` DB to avoid shared-DB contention)
- Reject parsing (multi-recipient, reason, no-suppress, missing `reject` block) — `ses_test.go`
- Consumer: exactly one message-level `email.failed` for a 3-recipient Reject; duplicate-notification dedup key stability; no suppression; fallback reason; nil→`[]` for `to`; uncorrelated no-op — `consumer_test.go`
- Golden fixtures: the Reject-path payload byte-matches the committed `email.failed.json` / `email.failed.min.json` (same files the SDKs assert against)
- Cross-path deterministic-id equality with the worker path — `consumer_crosspath_test.go`
- DB integration: `sent→failed` rollup, both recipient rows failed, exactly one event with correlated payload, no suppression rows, idempotent duplicate processing; correlation field widening under the bracketed-SES-id shapes — `integration_test.go`

`go test -tags integration -p 1 ./internal/delivery/` ✅ · `./internal/identity/ ./internal/webhookpub/ ./internal/eventpayload/...` ✅ · `TestSpecGolden*` ✅ · `go vet`, `gofmt` clean.

### DIFF
--- a/cmd/e2a/main.go
+++ b/cmd/e2a/main.go
@@ -56,14 +56,16 @@ import (
 // derived deterministically from dedupKey so duplicate SNS notifications
 // produce the same id — receivers dedup on it (at-least-once delivery).
 func deliveryEventFirer(pub webhookpub.Publisher) delivery.Firer {
-	return func(ctx context.Context, userID, agentID, eventType string, data any, dedupKey string) {
+	return func(ctx context.Context, e delivery.FiredEvent) {
 		pub.Publish(ctx, webhookpub.Event{
-			ID:        webhookpub.DeterministicEventID(dedupKey),
-			Type:      eventType,
-			CreatedAt: time.Now().UTC(),
-			UserID:    userID,
-			AgentID:   agentID,
-			Data:      data,
+			ID:             webhookpub.DeterministicEventID(e.DedupKey),
+			Type:           e.Type,
+			CreatedAt:      time.Now().UTC(),
+			UserID:         e.UserID,
+			AgentID:        e.AgentID,
+			ConversationID: e.ConversationID,
+			MessageID:      e.MessageID,
+			Data:           e.Data,
 		})
 	}
 }
--- a/docs/events.md
+++ b/docs/events.md
@@ -58,7 +58,7 @@ for e in client.events.list(type="email.received", limit=20):
 | `email.received` | Inbound SMTP message accepted | **At-least-once** end-to-end |
 | `email.flagged` | Inbound message accepted but did not match the agent's `inbound_policy` (delivered + flagged, never dropped) | **At-least-once** end-to-end |
 | `email.sent` | Outbound `/send` accepted by SES | Best-effort |
-| `email.failed` | Outbound send terminally failed (retries exhausted / permanent reject) — carries `reason` | **At-least-once** |
+| `email.failed` | Outbound send terminally failed — retries exhausted / permanent reject at send time, or the provider rejected the already-accepted message (SES `Reject` delivery feedback, e.g. content scan) — carries `reason` | **At-least-once** from the send path; best-effort when emitted via delivery feedback |
 | `email.review_requested` | Message held for human review (outbound HITL or inbound screening) — carries `direction` | **At-least-once** |
 | `email.review_approved` | Review approved (outbound: sent; inbound: released to the inbox) | Best-effort |
 | `email.review_rejected` | Review rejected (outbound: discarded; inbound: dropped) | **At-least-once** |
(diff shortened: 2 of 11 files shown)


jiashuoz:
## Review finding addressed — commit 0887e78e

**Finding (VALID, verified):** `deliveryEventFirer` (cmd/e2a/main.go) dropped the `ConversationID`/`MessageID` envelope routing keys, because `delivery.Firer`'s positional signature had no slots for them. Consequence confirmed by tracing: `webhookpub` persists `message_id`/`conversation_id` from `e.MessageID`/`e.ConversationID` (NULL when empty), and `GET /v1/events` filters `WHERE message_id = $1` / `conversation_id = $1` — so a Reject-path `email.failed` (and every existing `delivered`/`bounced`/`complained`) was stored with `message_id=NULL` and **invisible to the reconciliation query** an integrator uses to learn a message's fate. This undermined the PR's goal and split routing between the two `email.failed` emitters.

### Per-item disposition
1. **Extend `Firer`** — done. Used a **`FiredEvent` struct** (`UserID/AgentID/ConversationID/MessageID/Type/Data/DedupKey`) rather than 8 positional args — the two adjacent same-typed `conversationID`/`messageID` strings were swap-prone; the struct reads cleaner and matches the growing-arg-list guidance.
2. **`deliveryEventFirer`** now sets `ConversationID`/`MessageID` on the `webhookpub.Event`. Purely additive (NULL→populated), no migration (columns exist).
3. **Call sites** — the three message-backed fires (`delivered`/`bounced`/`complained` + `Reject`/`failed`) pass `m.MessageID`/`m.ConversationID` (both already on `CorrelatedMessage` — no store/query change). **`suppression_added` deliberately leaves all routing keys empty** (verified correct): it is account-scoped, not tied to one message thread; the triggering message already rides in the payload's `source_message_id`, and leaving the envelope keys empty keeps it from being filtered out of a message/thread-scoped events query it was never about. One-line comment added.
4. **Golden fixtures / spec — no change.** The wire `Envelope` (`AsEnvelope`) does not carry these fields (they're DB filter columns only), so `internal/eventpayload` goldens, `internal/delivery` goldens, `TestSpecGoldenNoDrift`, and `api/openapi.yaml` are all unchanged (`git diff` clean on the spec). No `make generate-sdk` needed. `email.failed` envelope now matches the worker-path envelope (full payload+envelope parity). Dedup id = `DeterministicEventID(dedupKey)` — independent of routing keys, so cross-path id equality is preserved (asserted, still green).
5. **DB regression test** — `TestDeliveryFeedback_FindableByMessageID` (internal/agent, `//go:build integration`): wires the consumer to the production `OutboxPublisher`, consumes a Reject and a Delivery, then asserts `ListEventsForUser` (the exact `GET /v1/events` query func) returns the `email.failed` by `message_id` **and** by `conversation_id`, and the `email.delivered` by `message_id`, plus a cross-message negative check proving the filter is applied.

**Behavior note:** delivery-feedback events (`delivered`/`bounced`/`complained`/`failed`) become findable via `?message_id=`/`?conversation_id=` where they weren't — desired GA semantics, purely additive.

### Gates (all on a dedicated `e2a_test_sesreject2` DB, never shared `e2a_test`)
- `go test -tags integration -p 1 -count=1 ./internal/delivery ./internal/webhookpub ./internal/agent` → **ok / ok / ok**
- `TestSpecGolden*` (= spec-check gate) → **ok**; `internal/eventpayload/...` goldens → **ok**
- `git diff api/openapi.yaml` → empty (no /v1 schema change); `go build ./...`, `gofmt`, `go vet ./internal/delivery` clean

**New test:** `TestDeliveryFeedback_FindableByMessageID`. Also added envelope-routing assertions to the existing delivery consumer + integration tests (delivered/failed set the key; suppression_added leaves it empty).

### Coordination
Expected conflict with sibling **#519** (`fix/ga-failed-correction-guard`): both touch `internal/delivery/consumer.go`'s `fire()` call sites. My change converts those call sites to the `FiredEvent` struct form — the conflict is mechanical (re-wrap #519's added/edited fire call in the struct literal). Firer-signature change kept scoped to that.

PR #522, opened 2026-07-17
refactor(env): rename E2A_BASE_URL to E2A_API_URL; split it from the CLI's E2A_URL

jiashuoz:
## Why

The client-side base-URL env vars named one concept badly and two concepts identically, so **following our own docs silently misrouted traffic**.

`E2A_URL` and the SDKs' `E2A_BASE_URL` are not the same thing:

- The CLI's `api_url` is the **deployment root**. `login` opens a browser against it (`cli/src/commands/login.ts:310`) and points users at `/get-started` — a Next.js route (`web/src/app/(app)/get-started`) the Go backend does not serve. `https://e2a.dev` works because the web front also proxies `/v1` (`web/Caddyfile:35`).
- The SDKs and the MCP server only ever call `/v1`, and want the **API host** alone.

Two names for "the base URL", with two different defaults, hid that distinction. The concrete failures:

| | |
|---|---|
| **Docs taught a var the SDKs don't read** | `README.md` and `docs/deployment.md` teach `E2A_URL`; the SDKs read only `E2A_BASE_URL`. A self-hoster who followed the docs got their SDK silently pointed at production `api.e2a.dev`. |
| **The CLI alias broke `e2a login`** | The CLI accepted `E2A_BASE_URL` as an alias for its deployment root, so exporting `E2A_BASE_URL=https://api.e2a.dev` to configure an SDK silently repointed the CLI at the API host. |
| **tether forced the CLI onto the API host** | `tether.env.example` taught `E2A_BASE_URL` ("self-host: your API base") and `lib.sh:51` defaulted it to `api.e2a.dev` — overriding the CLI's own `e2a.dev` default through that alias. |

## What changed

- **SDKs (TS + Python)** read `E2A_API_URL`, matching the name the server already uses for the same concept (`internal/config/config.go:344` — its externally visible API base). `E2A_BASE_URL` still works, with a deprecation warning (`console.warn` / `DeprecationWarning`).
- **MCP** moves to `E2A_API_URL` (`E2A_URL` + `E2A_BASE_URL` deprecated) and now defaults to the API host — it is a pure API client.
- **CLI** keeps `E2A_URL` alone and no longer reads `E2A_BASE_URL`. Dropping the alias *silently* would strand tether self-hosters on production, so it warns — gated to the actually-dangerous case (legacy var set, `E2A_URL` unset, root still at default), so a legitimate SDK user who also configured their CLI gets no noise.
- **tether** migrates to `E2A_URL`, carries a legacy `~/.e2a-tether.env` over with a warning (rather than reverting to prod), unsets the stale name so the CLI doesn't double-warn, and tracks the CLI's default instead of forcing `api.e2a.dev`.
- **Docs**: dropped the `E2A_AGENT_EMAIL` fallback claim from both SDK READMEs and the root README — it was never implemented in either SDK (they raise `missing_email`), though it is real in the CLI and tether. Added the CLI's env-var table (it documented none despite reading five) and split `E2A_URL` vs `E2A_API_URL` in `docs/deployment.md`.

Also removed the MCP one-shot deprecation guard: `main()` calls `loadConfig` exactly once, so it deduped nothing and only leaked state across tests.

## Compatibility

Additive. Every previously-working configuration still works; the old names warn rather than break. The one behavior change is the CLI ignoring `E2A_BASE_URL` — loud, not silent, and tether (its main consumer) migrates itself.

## Verification

Suites: TS SDK typecheck+unit+types, MCP vitest+types, Python 323 passed / 18 skipped, CLI 120 passed, tether `bash -n` clean.

Driven for real, not just unit-tested:
- Both SDKs against actual dialed origins across all four resolution cases (default → `E2A_API_URL` → deprecated `E2A_BASE_URL` + warning → both set, canonical wins).
- The built CLI binary: warning fires for the self-hoster case, silent when `E2A_URL` is set.
- `t_load_config` through the legacy-migration path: carries over, warns, unsets, and defaults correctly.

> ⚠️ `pytest` inside a git worktree silently tests the **main checkout** via the editable install — run it with `PYTHONPATH=$(pwd)/src` or you are testing the wrong tree.

## Follow-ups (not in this PR)

- `mcp/src/config.ts` is dead code (nothing in `src/` imports it but a type; no `bin` field since npm stdio publishing was retired in #247) and now also contradicts the new canonical name. Deleting it is a separate call.
- The split-host topology `internal/config/config.go:377` describes ("web on one host, API/MCP on another") still can't be expressed by the CLI, which holds only one URL. That's the deeper fix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,7 +287,7 @@ WebSocket (no public URL needed):
 from e2a.v1 import AsyncE2AClient
 
 async with AsyncE2AClient(api_key="e2a_…") as client:
-    async for notif in client.listen("bot@your-domain.com"):  # falls back to E2A_AGENT_EMAIL
+    async for notif in client.listen("bot@your-domain.com"):
         # notif is lightweight metadata — fetch the body when you want it
         email = await client.messages.get(notif.recipient, notif.message_id)
         await client.messages.reply(notif.recipient, notif.message_id, {"body": "Got it!"})
--- a/README.md
+++ b/README.md
@@ -310,7 +310,8 @@ Three audiences each configure a different surface:
 | Audience | What they configure | Where |
 |---|---|---|
 | **Server operator** — runs the Go backend | DB, signing key, SMTP, OAuth, optional shared domain | `config.yaml` + `E2A_*` env |
-| **CLI / SDK user** — calls the API from their machine | Just the deployment URL (and login) | `E2A_URL` + `e2a login` |
+| **CLI user** — drives an inbox from a terminal | Deployment URL + login | `E2A_URL` + `e2a login` |
+| **SDK / MCP user** — calls `/v1` from code | API host + key | `E2A_API_URL` + `E2A_API_KEY` |
 | **Web dashboard deployer** — hosts the Next.js dashboard | Public site URL + branding | `NEXT_PUBLIC_*` build-time env |
 
 The Go binary runs on any container host; storage is plain Postgres 14+; outbound mail goes through standard SMTP. Most workers coordinate via `SELECT … FOR UPDATE SKIP LOCKED`, so multi-replica is safe — the two real horizontal-scaling caveats are in-memory WebSocket fan-out and per-process rate limits.
--- a/cli/README.md
+++ b/cli/README.md
@@ -87,6 +87,28 @@ e2a config get agent_email
 e2a config set agent_email bot@acme.com
 ```
 
+## Environment variables
+
+| Variable | Default | Description |
+|---|---|---|
+| `E2A_API_KEY` | — | API key. Skips `e2a login` — useful in CI and scripts |
+| `E2A_URL` | `https://e2a.dev` | The e2a deployment root. Set for self-host |
+| `E2A_AGENT_EMAIL` | — | Default sending/listening inbox (what `--agent` overrides) |
+| `E2A_SHARED_DOMAIN` | auto-discovered | Force the shared domain instead of discovering it via `GET /v1/info` |
+
+**Precedence:** command-line flags beat environment variables, which beat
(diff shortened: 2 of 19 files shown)


jiashuoz:
Both addressed in 6029c087.

**[P1] Docs no longer misroute SDK users.** The `CLI / SDK user` row and the single `export E2A_URL` walkthrough were the misdirection — the `E2A_API_URL` row I'd added lower down didn't undo the headline instruction. Split into two audiences with two setup examples, in both the `docs/deployment.md` table and the README table:

- **CLI user** → `E2A_URL` + `e2a login`
- **SDK / MCP user** → `export E2A_API_URL=…` + `E2A_API_KEY`

Each section now names the var the *other* audience's tooling ignores, e.g. "The CLI does **not** read `E2A_API_URL` … so a self-hoster who only exports `E2A_API_URL` leaves the CLI pointed at production."

**[P2] Legacy override is no longer silently ignored.** You're right that gating on `resolved === DEFAULT_URL` missed the stored-host case: `E2A_BASE_URL` used to override `~/.e2a/config.json`, so stored host A + legacy env host B silently kept A. The warning now fires whenever `E2A_BASE_URL` is set and `E2A_URL` is absent, and names the host actually in use so the mismatch is visible. Driven against the built binary with a seeded non-default `api_url`:

```
e2a: E2A_BASE_URL is set but the CLI does not read it — that name configures the SDKs.
     The CLI uses E2A_URL for the deployment root and is talking to https://stored.selfhost.dev.
     Set E2A_URL to point it at a self-hosted deployment.
```

Regression test added with a stored non-default `api_url` asserting both the warning and that it names the stored host. CLI suite 121 passing.

PR #593, opened 2026-07-20
feat(sdk): add inbound email facade

jiashuoz:
## Summary

- add `client.inbound.fromEvent(event)` / `client.inbound.from_event(event)` with TypeScript, Python async, and Python sync `InboundEmail` facades
- hydrate the full canonical `MessageView` from lightweight `email.received` webhook or WebSocket envelopes
- expose parsed RFC 5322 From, nullable SMTP envelope identity, full SPF/DKIM/DMARC authentication evidence, aligned-DMARC `verified`, reply-target preview, parsed-body truncation, policy flags, and safe metadata projections
- bind reply, forward, and lazy attachment `get()` operations to the hydrated inbox/message identifiers
- add shared full, minimal, adversarial, and invalid conformance vectors consumed by both SDK suites
- document webhook verification, authenticated WebSocket use, untrusted content, HITL outcomes, and attachment behavior across the root, SDK, plugin, and web documentation surfaces

## Trust and safety boundary

- raw webhook bodies must pass `constructEvent` / `construct_event`, which verifies the HMAC signature and replay timestamp before returning a typed event
- authenticated `client.listen(...)` WebSocket events can be passed directly to the same facade
- `fromEvent` / `from_event` validates the schema-v1 `email.received` envelope and non-empty `message_id` / `delivered_to` fetch keys, but does not establish trust for hand-built events
- `verified` is true only when the hydrated authentication evidence reports `dmarc.status == "pass"`; this authenticates domain-authorized use of the RFC 5322 From domain, not the mailbox local part, a person, or message content
- From, Reply-To, bodies, filenames, and content types remain sender-controlled; `replyTargets` is an inspection preview, not a trust signal
- `flagged` exposes the current policy-gate signal and is explicitly not a comprehensive prompt-injection verdict
- reply and forward preserve the existing `SendResultView`, including `pending_review`, rather than implying that an accepted HTTP response was delivered
- `toJSON()` / `to_dict()` omit raw MIME, attachment download data, transport objects, and client internals

## Compatibility and scope

- additive handwritten SDK surface only; no server, database, OpenAPI, or generated-model changes
- existing generated models and lower-level `webhooks.fetchMessage` / `fetch_message` remain supported
- no package version bump and no committed TypeScript `dist/`
- webhook and WebSocket envelope formats are unchanged

## Validation

- TypeScript SDK: typecheck, 197 unit tests, type tests, and build pass
- Python SDK: 348 passed, 19 skipped; mypy passes
- TypeScript and Python contract suites pass
- repository text-integrity and SDK example-contract guards pass
- generated-code freshness, OpenAPI contract gates, and design-system freshness pass
- `git diff --check origin/main...HEAD` passes


jiashuoz:
## Review: 4-dimension agent review (contract fidelity, SDK design, security, doc rigor)

Docs-only PR, so nothing ships broken. But the plan is what an implementer executes from, and it contains steps that cannot run and guarantees that are false. Three of the four reviewers independently found items 1–4 below with the same file:line evidence.

**Verdict: needs revision before implementation.**

---

### Confirmed by 3+ independent reviewers

**1. `sdks/typescript/dist/` is gitignored and untracked — five plan steps depend on it being committed.**
`.gitignore:27` lists `sdks/typescript/dist/`; `git ls-files sdks/typescript/dist` returns zero files. Affected: `design.md:214`, `plan.md:21`, `:442`, `:477-483`, `:504` (`git add sdks/typescript/dist` is a no-op on an ignored path), and acceptance criterion `plan.md:525` ("committed TypeScript build output"), which is unsatisfiable. An implementer will believe a verification gate ran that did not.
*Decision needed:* drop the dist workflow, or change `.gitignore` (separate, contentious).

**2. The "rejected locally before an HTTP request" guarantee is false.**
`design.md:227-229` and `plan.md:144` promise an event missing either fetch key is rejected locally. But `isEmailReceived` / `is_email_received` check only two fields — `webhook-signature.ts:278-280`, `webhook_signature.py:374-376`:
```ts
return e.schema_version === "1" && e.type === "email.received";
```
`data` is `unknown`/`Any` and is merely *cast*, so the guard's narrowing is an assertion, not a check. The reference implementation (`plan.md:99-106`, `:258-269`) relies on the guard alone, then dereferences `event.data.delivered_to` — yielding `GET /v1/agents/undefined/messages/undefined` instead of a local `E2AValidationError`.
Sharpest framing: **this makes the new facade strictly weaker than the helper it supersedes** — `fetchMessage` / `fetch_message` (`client.ts:544-551`, `client.py:779-786`) *do* validate both keys.

**3. The Python sync-bridge failure mode is misdiagnosed, so Task 4's red step won't fail as described.**
`design.md:178-180` and `plan.md:349` predict the async facade is "returned unchanged." It won't be: `_wrap_value` → `_is_resource` (`sync_client.py:250-271`) already proxies any non-callable `e2a.*` object lacking `model_fields`, and `_wrap_attr` already makes `reply`/`forward` blocking.
The real gaps are narrower: `attachments` is a plain `list` that `_wrap_value` passes through untouched, so `attachments[0].fetch()` *does* leak a coroutine; and `isinstance(email, InboundEmail)` (`plan.md:337`) fails against a `_SyncResourceProxy`.
Compounding: `plan.md:254` says "frozen/value-oriented facade classes" without choosing dataclass vs. Pydantic — and that unmade choice silently decides everything, since a Pydantic facade has `model_fields`, fails `_is_resource`, and leaks coroutines from *every* method.

**4. `parsed.truncated` is required by the contract and silently dropped.**
`MessageParsedView` requires `truncated` (`api/openapi.yaml:2058-2062`), and `internal/httpapi/messages.go:141-154` documents `text` as quote-stripped and length-capped. So `email.text` — the headline field — can present a server-truncated body as complete, with no signal on the facade. Also: `design.md:150`'s "absent text normalizes to the empty string" only fires when `parsed` itself is absent, since `parsed.text` is required. And the design never says why `MessageView.body` is ignored in favor of `parsed`.

---

### Most serious finding: the design reasons about envelope authenticity, not message authenticity

The webhook-signature boundary is handled well (`design.md:48-49`, `229-235`). The per-message authenticity/screening verdict is never engaged, and the field selection inverts it.

**5. BLOCKER — `authenticatedFrom` is not a verification result, but the facade makes it the identity field.**
`design.md:130`, `:171` treat it as the safe counterpart to display `from`. It isn't: `internal/relay/server.go:330-332` parses the From header, and `:419-421` passes it as `authenticatedFrom` **unconditionally, with no branch on the verdict**. The actual pass/fail is `Verified: domainAuth.DomainAuthenticated()` (`server.go:369`), persisted as structured `MessageView.auth` (`internal/httpapi/messages.go:106`).
The facade exposes **neither** `auth` nor `authHeaders` (`design.md:116-140`, `165-174` — both field lists omit them); they survive only two levels down `.message.auth`. The server deliberately delivers auth failures rather than dropping them (`server.go:389-393`: "A non-match is FLAGGED — still delivered (never dropped)"), so a spoofed `spf=fail; dkim=fail` message renders byte-identically to a DMARC-aligned one — and the prescribed README example (`plan.md:459-463`) auto-replies to both.
*Fix:* promote `auth` (or a derived `verified: boolean`) to a first-class property; rename or document `authenticatedFrom` as "the identity the verdict pertains to, not a verdict."

**6. BLOCKER — no screening / prompt-injection signal anywhere, and `.message` is not a fallback.**
The strings "prompt injection", "screening", "flagged", "SPF", "DKIM" appear nowhere in 836 lines. This is worse than a normal omission because the content-scan verdict is genuinely unreachable from the facade:
- `MessageView.protection[]` / `hold_reason` are deliberately nil on agent-facing APIs (`internal/httpapi/messages.go:78-86`); only `handleGetReview` populates them (`internal/httpapi/reviews.go:221-222`).
- `scan_score` / `scan_action` are written (`migrations/040_screening.sql:21-23`) but never selected by `GetMessageWithContent` (`internal/identity/store.go:3687-3706`).

So for a piguard `action=flag` message the entire reachable signal is `.message.flagged` — the one field the design omits. Its docstring states the reason it exists: *"Polling agents need this signal because no review item is created for the flag action."*
Mitigating: `action=review`/`block` emit no `email.received` and 404 on fetch (`internal/identity/review.go:131`). The gap is precisely the "delivered but untrusted" tier — the tier this ergonomic path is built for.

**7. HIGH — `.reply()` sends to attacker-controlled Reply-To, and no facade property says so.**
Two attacker-steered hops the design never connects: `.from` is the Reply-To when present (`server.go:384-386`), and `.reply()`'s destination is resolved server-side the same way (`internal/outbound/reply.go:18`, `40-47`: Reply-To if present, else From). So `await email.reply(...)` — a bound method whose whole pitch is "reply to this email" — sends to whatever the attacker put in Reply-To, which is *not* `.authenticatedFrom`. `From: ceo@victim.com` + `Reply-To: attacker@evil.com` routes the agent's reply to the attacker. `.replyTo` exists on the facade but is never linked to `reply()`.

**8. MEDIUM — the canonical fixture hardcodes an empty attestation.**
`plan.md:61` sets `auth_headers: {}` in the single shared cross-SDK fixture (`design.md:290-294`) driving every test in Tasks 1/3/4. Empty `auth_headers` is the WebSocket drain-path *degradation* case (`webhook-signature.ts:134-136`), normalized here as the happy path. No test exercises `spf=fail`, `flagged=true`, or a divergent Reply-To. Add a second fixture covering all three.

---

### Also blocking or misdirecting

**9. `pending_review` is absent from the entire design.** `design.md:138` types `reply`/`forward` as `Promise<SendResultView>` and the edge-case list (`:226-247`) covers auth/404/rate-limit/malformed/closure — never the hold. Both endpoints register as **202** (`internal/httpapi/outbound.go:322-338`), with `SendResultView.status` in `{accepted, sent, pending_review, review_approved, failed}` and the contract note *"Always branch on this field, not the HTTP status code"* (`outbound.go:126`). Both also return **409 `message_not_yet_delivered`**. Both README examples (`plan.md:449-463`) discard the result, teaching callers to treat a held, unsent draft as delivered — the CLI reserves exit code 3 for exactly this.

**10. Python `from_event` neither typechecks nor runs as specified.** `plan.md:258` types the param `EventLike`, which declares only `type` and `data` (`client.py:102-115`); `is_email_received` is annotated `(e: WebhookEvent)` and reads `e.schema_version`. mypy (`plan.md:423`) rejects this. Underneath is a real asymmetry: TS has `export type WSEvent = WebhookEvent` (`ws.ts:101`), but Python's `WSEvent` (`websocket.py:187`) is a *separate* dataclass — so `design.md:31-32`'s "accept either envelope" goal needs a union the plan never specifies.

**11. `fromEvent` would reject every persisted event in TypeScript but accept it in Python.** `design.md:60` claims webhooks, WS, and persisted events share one envelope. True on the wire, false at the SDK layer: `client.events.get()` returns generated `EventView`, which is **camelCase** in TS (`EventView.ts:19,30,34` — `schemaVersion`), so `isEmailReceived` reads `schema_version` → `undefined` → `false`. Python's `EventView` happens to be snake_case and duck-types through. Undocumented cross-SDK behavioral split; neither doc scopes `fromEvent` to exclude `EventView`.

**12. `attachment.fetch()` never returns bytes.** `design.md:37`/`:186` promise "lazy content retrieval," but `getAttachment` returns `AttachmentView` = metadata + `downloadUrl` + `expiresAt`; `data` appears only when `inline: true` **and** under the 256 KB cap (`client.py:484-491`). So the default `fetch()` returns metadata plus a signed URL — never bytes — and `design.md:208` hand-waves the >256 KB case, which is the common one. Rename (`link()`/`describe()`/`download()`) or make it return bytes. Compounding: `:243`'s no-caching rule means an N-attachment loop is N round trips and a second `fetch()` mints a second short-lived URL.

**13. Attachment source contradicts between spec and plan.** `design.md:146-147` derives metadata from the event payload; `plan.md:148` says use the hydrated message. These are different types **exporting the same name** — the event's `AttachmentMetaView` is hand-written snake_case (`size_bytes`), the message's is generated camelCase (`sizeBytes`); `index.ts:48-54` documents the resulting star-export collision. They also differ in practice: `EmailReceivedData.attachments` is optional and omitted when empty, `MessageView.attachments` is required. Pick the hydrated message (authoritative for `index`) and write down `InboundAttachment`'s field list — `design.md:214` exports it as a stable v1 type without ever enumerating it.

**14. Surface duplication frozen at GA.** After this ships, `messages.get(delivered_to, message_id)`, `webhooks.fetchMessage(event)`, and `inbound.fromEvent(event)` all do the same thing and are all permanent — `design.md:64-66` concedes `fetchMessage`'s namespace is wrong, then `:108-110` keeps it unchanged forever. Also, every existing `E2AClient` resource maps 1:1 to an API path; `inbound` would be the only namespace with no endpoint behind it. `client.messages.fromEvent(event)` fits the house idiom — `messages` already owns `get`/`reply`/`forward`/`getAttachment`, exactly the dependency set `plan.md:88-93` re-declares as `InboundMessageOperations`. If `inbound` stays, soft-deprecate `fetchMessage` in the same release.

**15. Rollout is asserted but never executed.** `design.md:296-301` says it ships in a minor SDK release after version sync, but no step bumps `sdks/typescript/package.json` or `sdks/python/pyproject.toml` (both `5.2.0`). `plan.md:495` runs `check-sdk-version-sync.mjs` expecting PASS — that script only verifies `mcp`/`cli` declare a range satisfying the current SDK major; it is not a bump gate and passes unchanged. No target version or owner named. `plan.md:469` also assumes an "unreleased section" exists in both changelogs without verifying.

**16. Repo gates on the exact edited files are never run.** The plan edits `README.md` and both SDK READMEs (`plan.md:30`, `:437-440`) but never runs `scripts/check-sdk-example-contracts.mjs`, which asserts forbidden stale-example regexes in precisely those three files plus `web/public/sdk.md` and `docs/events.md`. `check-repository-text-integrity.sh` is also unrun. And `web/public/sdk.md` — the agent-facing doc that already teaches the `n.delivered_to` / `n.message_id` dance this facade replaces — is absent from the file list, leaving the primary consumer doc contradicting the new path.

**17. Python 3.9 floor vs. the plan's own fixtures.** `pyproject.toml` pins `python_version = "3.9"`; `plan.md:255` says only "parse `received_at` with the standard library." Fixtures at `plan.md:62`, `:242` use `"2026-07-20T08:00:00Z"`, and `datetime.fromisoformat` rejects `Z` before 3.11 — the obvious reading fails the plan's own test on the supported floor.

**18. Contract-test tier omitted with no decision recorded.** The repo has a real harness (`test:contract`, `tests/test_contract.py`, `cmd/e2a-contract-server`). Every verification step here is unit-only — vitest with stubbed message ops, pytest with `httpx_mock`, mypy. Since `from_event` composes a real `GET /v1/agents/{email}/messages/{id}`, stubs cannot catch a wrong path or parameter. `design.md:289-294`'s "cross-SDK contract checks" is a shared fixture, not the contract tier.

---

### Smaller, but will fail as written

- `plan.md:287`, `:340` assert `result.id == "msg_reply"` — `SendResultView` has `messageId`/`message_id`, no `id`.
- `plan.md:44-65` defines the canonical fixture with `attachments: []`, then `:135` calls `email.attachments[0].fetch(...)`.
- `design.md:138` makes `options` optional but `plan.md:127-133` expects the delegate called with a literal `{}` as the 4th arg — `toHaveBeenCalledWith` fails unless the facade default-fills.
- `design.md:240` aborts `fromEvent` on an unparseable `received_at`, so a server date glitch disables `reply`, `forward`, and attachments too. Also miscategorizes: `E2AValidationError`/`status: 0` is documented as the *caller-input* convention (`errors.py:99-105`). Prefer a nullable `receivedAt`.
- `design.md:131`/`:239` normalize absent `cc`/`reply_to` to `[]`, erasing absent-vs-empty — load-bearing for `reply_to`, since "the relay never silently falls back to `from`" (`webhook-signature.ts:130-133`).
- `design.md:57-58` overstates the Python guard: only TS narrows the data payload; Python's `TypeGuard[WebhookEvent]` is a no-op and `data` stays `Any`.
- `design.md:152-154` keeps `event`/`message` public but never states `message` is a point-in-time snapshot — read-state, labels, and trash drift immediately. One sentence, or a `refresh()`.
- `plan.md:415-417` punts an unresolved decision to the implementer ("according to the Task 4 implementation boundary"), which Task 4 never states. Relatedly `sync_client.py:58`'s `__all__` is never extended, but Task 5 needs `from e2a.v1 import InboundEmail`.
- `design.md:186-208` presents attachment `filename`/`contentType` as trusted metadata — both are attacker-authored MIME values, inviting `fs.writeFile(att.filename, ...)` path traversal and content-type dispatch. One doc line needed.
- `design.md:118-119` retains full `MessageView` (carrying `rawMessage` base64 MIME) and `downloadUrl` (a credential-bearing presigned URL) on an object designed to feel like a friendly value type — it will get `console.log`'d and pasted into LLM context. Consider `toJSON`/`__repr__` eliding `rawMessage`.
- `plan.md:15` re-creates `src/v1/inbound-email.ts` with `export class InboundEmail` — the exact path and identifier retired in #227, whose old getter-gate semantics (`UnverifiedEmailError`) `design.md:48-49` explicitly disowns. Fine to reuse as the fast-follow, but `index.ts:6-7`'s retirement comment becomes untrue and isn't in the file list.
- Naming: async client resources are unprefixed (`MessagesResource` on `AsyncE2AClient`), but the plan introduces `AsyncInboundResource`. The two existing bridge wrappers are `SyncAutoPager`/`SyncStream` — the opposite prefix convention from the one chosen. State the rule.
- `plan.md:13` heads "File structure" where every sibling plan uses "File map".

---

### Cleared

Worth recording so it isn't relitigated:

- **Codegen survival is fine.** `generate-oag.sh` only writes `sdks/typescript/src/v1/generated`, `.openapi-generator-ignore` is empty, and the new modules sit outside that tree — the layer survives regeneration and `make generate-sdk-check` cleanly.
- **No SSRF, no unbounded download.** Server caps inline at 256 KB (`internal/httpapi/attachments.go:28,176,214`) and `fetch()` returns `downloadUrl` without following it, so the SDK makes no attacker-influenced request. `design.md:243`'s no-caching-of-short-lived-URLs is correct.
- Envelope shape, event type names, `messages.get/reply/forward/getAttachment` signatures, `EmailReceivedData` field names/nullability, WS↔webhook envelope identity, error classes, and idempotency mint-once all check out.
- The additive, no-server-change scoping is sound; the GA risk is surface duplication (#14), not schema drift.
- Structurally both docs conform to siblings, and the 6-task slicing is independently verifiable.

---

### Two judgment calls for the author, not defects

1. **`inbound` namespace vs. `client.messages.fromEvent`** (#14) — a real fork, worth deciding deliberately before GA freeze.
2. Whether bound `reply`/`forward` are even valid on a message held in `pending_review`, or one soft-deleted / trash-expired before hydration. Given this repo's HITL and 30-day trash semantics that's a product decision, not an error-propagation detail, and neither document addresses it.

<sub>Reviewed by 4 parallel agents across contract fidelity, SDK design/parity, security, and doc rigor. Items 1–4 were found independently by 3+ reviewers.</sub>

jiashuoz:
## Re-review after fixes — 4 agents (regression, TS impl, Python impl, security + spec)

Scope changed since the last review: this went from docs-only to a full implementation (37 files, +1821/−87), and it now touches `api/openapi.yaml`, `internal/eventpayload/payloads.go`, and both generated SDK trees. **The PR description is stale** — it still says "no runtime, API, schema, database, or generated-code changes," which is now false on four counts.

**Most of the previous review is genuinely fixed in behavior, not reworded.** Verified by re-running every gate on a clean worktree: TS 196 pass + `tsc` clean; Python 347 pass / 19 skipped; mypy clean; `check-sdk-example-contracts.mjs`, `check-repository-text-integrity.sh`, and `TestSpecGoldenNoDrift` all pass; `git ls-files sdks/typescript/dist` = 0.

Of the 18 prior findings: **12 fixed**, **4 resolved by explicit scoping decisions** (#6, #11, #14, #18), **2 partial**. Particularly well done — #2 is now real runtime validation that is *stricter* than the `fetchMessage` helper it supersedes, with 14 invalid vectors and both suites asserting zero transport calls; #17 was fixed at the root by sourcing `received_at` from the already-parsed `MessageView.created_at` rather than parsing the wire string.

Three things block. The first is new, introduced by the fix itself.

---

### B1 — BLOCKER, new. `authenticated_from` is the **From header**, not the SMTP envelope. The wrong name is now a frozen public field.

The previous review said `authenticatedFrom` was presented as a verdict when it is only an identity. That was right, and the verdict half is now correctly fixed. But the remedy renamed the field to **`envelopeFrom` / `envelope_from`**, and that name is wrong.

Verified directly in the relay:

```go
// internal/relay/server.go:329-332
senderEmail := extractEmail(in.EnvelopeFrom)
if threadInfo.From != "" {
    senderEmail = threadInfo.From      // ← From header overrides the envelope
}
// "SPF/DKIM/DMARC against the TRUE envelope MAIL FROM (RFC 7208), not the From
//  header — else SPF-alignment is a tautology"
```

`senderEmail` becomes `authenticated_from` (`server.go:419-421`). The envelope is only a fallback for mail with no parseable `From:`. An existing e2e test already pins this: `internal/e2e/inbound_policy_e2e_test.go:221` asserts `authenticated_from == "friend@trusted.com"` with the comment *"want the gated From"*.

The Go doc string and `api/openapi.yaml` are **correct** ("The From-header identity…"). Nine downstream places now contradict them, including `webhook-signature.ts:122-124`, `webhook_signature.py:179`, `docs/events.md:113`, the `inbound.ts:92` / `inbound.py:95` error strings, and — the part that is expensive to undo — the public field `inbound.ts:146`, the `toJSON()` key `inbound.ts:54`, and `inbound.py:165,204`.

Two compounding harms:

1. **Name collision.** `messages.envelope_from` already exists and means the *outbound* SMTP envelope sender (`internal/identity/delivery_store.go:414,510`). The SDK now publishes that exact name meaning something different.
2. **It inverts the threat model.** The correct model is "attacker-*authored* (it's the From header) but attacker-*attested* (SPF/DKIM/DMARC pertain to it)." The new wording describes a server-observed transport fact, inviting the inference that `envelopeFrom` reveals the true sender behind a spoofed From. The old wording overclaimed the verdict but was right about the identity; the new one fixes the verdict and breaks the identity.

*Fix:* keep `authenticatedFrom` / `authenticated_from` (matches the wire, the Go source, and the spec; no collision), and replace "SMTP envelope identity" with "From-header identity that the SPF/DKIM/DMARC verdict pertains to."

*Note on reviewer disagreement:* one agent recorded this rename as "FIXED, correctly." I checked the relay source directly and it is not — the three citations above are dispositive. Flagging the dissent so it is on the record rather than silently dropped.

---

### B2 — BLOCKER. The shared fixtures encode wire states the server cannot emit, and both SDKs deep-equality-assert them.

Two independent defects, both confirmed against `server.go`:

**(a) `message.from` is wrong in every fixture.** `server.go:384-387` sets `displaySender = threadInfo.ReplyTo[0]` when Reply-To is present, and *that* is what is persisted as the message sender (`server.go:546`) and surfaced as `MessageView.From` (`internal/httpapi/messages.go:167`). So `full.json` — which has `reply_to: ["reply@customer.example.com"]` and correctly sets the **event's** `from` to `reply@customer.example.com` (line 12) — sets the **message's** `from` to `"Alice <alice@customer.example.com>"` (line 51). The server cannot produce that pairing; it would be `reply@customer.example.com` in both. Same defect in `adversarial.json`.

This also means `message.from` carries a display name (`"Alice <…>"`), which relay never emits — `displaySender` is always a bare address. That matters because `replyTargets` falls back to `[message.from_]` (`inbound.ts:167`, `inbound.py:170`), so the facade would emit `"Alice <alice@…>"` as a *reply target*, which is not an address. `minimal.json` hides this by using a bare address.

**(b) The adversarial fixture teaches the inverse of the real attack.** `adversarial.json:10-11` pairs `from: "CEO <ceo@trusted.example>"` with `authenticated_from: "attacker@evil.example"`. Per B1, a message with `From: CEO <ceo@trusted.example>` yields `authenticated_from = ceo@trusted.example` — **the same spoofed identity**. The fixture depicts the SDK handing you the attacker's true address; it never does. The real delivered-but-untrusted case is `from == authenticated_from == ceo@trusted.example` with `auth` failing, where the *only* discriminator is the verdict — which is exactly the lesson the fixture should be teaching.

Because both suites do full deep-equality against these vectors (`inbound.test.ts:58`, `test_v1_inbound.py:75`), the incorrect model is now ratified in both languages. A cross-SDK conformance gate locked to an impossible server response proves less than it appears to.

---

### B3 — BLOCKER (quality). Mutation testing shows the TS suite does not hold the implementation to its own contracts.

Four mutations applied to `inbound.ts`; **all 6 new tests passed against each**:

| Mutation | Result |
|---|---|
| attachment index from array position, scaled ×100 | **passes** |
| `this.auth = undefined` — drop the auth verdict entirely | **passes** |
| `verified` drops the `\|\| dkim.status === "pass"` clause | **passes** |
| `flag_reason`/`html` emitted as explicit `undefined` instead of omitted | **passes** |

Causes:
- Every fixture has exactly one attachment at `index: 0`, so server `index` and array position are indistinguishable — the design's central attachment claim (stable server index is the fetch key) is untested. Needs ≥2 attachments at non-contiguous indices, e.g. `[0, 3]`.
- **`auth` is on the instance (`inbound.ts:147`) but absent from `toJSON()`/`InboundProjection`**, and no test reads `email.auth`. The projection is what an agent serializes into a prompt, and it carries only the collapsed `verified` boolean. (`flagged`, `flag_reason`, `text_truncated` *are* in the projection — correct.)
- No fixture has `dkim=pass` with `spf!=pass`.
- **`toEqual` ignores `undefined`-valued properties**, so the omitted-vs-`undefined` contract is unenforced in TS while Python's dict `==` against the *same* fixtures would catch it. The TS gate is strictly weaker than its Python twin, which undercuts the point of sharing fixtures.

The suite does catch real things (dropping `inline`, misreading `sizeBytes`, validate-after-fetch), and `expect(ops.get).not.toHaveBeenCalled()` at `inbound.test.ts:93` is a genuinely good assertion. The gap is specific, not general.

---

### High

**H1 — `verified` ignores DMARC.** `inbound.ts:167` / `inbound.py:167` compute `spf=="pass" || dkim=="pass"`. This faithfully mirrors `emailauth.AuthVerdict.DomainAuthenticated()` (`internal/emailauth/checker.go:44-46`), so it is inherited server semantics, not a new computation. But SPF is evaluated against the **envelope** MAIL FROM while `from`/`authenticated_from` are the **From header**, so an attacker with valid SPF on their own domain sending `From: ceo@victim.com` gets `dmarc=fail` and **`verified: true`**. Two agents reproduced this independently; a third classified it as inherited-and-documented. Both readings are defensible — but shipping it under the name `verified` on a *new* public field is a new harm even if the computation is old. Either fold in `dmarc`, or rename to `domainAuthenticated` / `spf_or_dkim_passed` and document that alignment lives in `.auth.dmarc`. Note `adversarial.json` sets `spf: fail`, so the fixture built to probe spoofing sidesteps the one case that breaks.

**H2 — No example branches on the trust signals before replying.** Credit: the field omission from the last review is properly fixed — `auth`, `verified`, `flagged`, `flagReason`, `textTruncated` are all first-class in both SDKs. But every prescribed example prints `verified` and then replies unconditionally: `README.md:161-165,180-183,289-292`, both SDK READMEs, `web/public/sdk.md:79-82,97-101`. `flagged` appears in no code example at all, only prose. The last review's concrete complaint — "the prescribed README example auto-replies to both" — stands verbatim. One `if (!email.verified || email.flagged) return;` per example closes it.

**H3 — Missing optional arrays crash with a raw `TypeError`.** `inbound.ts:174-176,185` spread `message.cc` / `message.replyTo` and call `message.attachments.map(...)`. A `MessageView` without `cc` throws `TypeError: message.cc is not iterable` out of `fromEvent()` as an untyped runtime error rather than an `E2AError`. The generated types mark these required, but this is the SDK's defensive boundary and the path it replaces never dereferenced them — a new crash surface. All fixtures supply `[]`, so none catch it. Fix: `[...(message.cc ?? [])]`.

---

### Medium

- **M1 — `console.log(email)` still leaks base64 MIME in TS only.** Python overrides `__repr__` on all four classes and asserts it (`test_v1_inbound.py:75-76`). TS overrides only `toJSON`, so `JSON.stringify` is safe (asserted) but Node's inspect does not call `toJSON` — `console.log(email)` dumps the public `message` field including `rawMessage`. Exactly the "fixed in one SDK, not the other" case.
- **M2 — `InboundResource` (sync) is not exported** while `AsyncInboundResource` is (`v1/__init__.py:24-30`); `inbound.__all__` lists it. `test_exports.py:60-66` codifies the asymmetry as expected rather than catching it.
- **M3 — Latent sync-bridge regression.** The attachment coroutine leak is genuinely fixed (verified live: `attachments[0].get()` returns `AttachmentView`, not a coroutine) because `from_event` rebuilds the list explicitly. But `_is_resource(AsyncInboundEmail)` still returns `True`, and routing one through `_wrap_value` *does* leak coroutines. Nothing takes that path today; any future async helper returning an `AsyncInboundEmail` silently regresses. No test guards it — add `AsyncInboundEmail`/`AsyncInboundAttachment` to the `_wrap_value` isinstance chain or a negative `_is_resource` assertion.
- **M4 — `data` is cast, not verified.** `inbound.ts:99` `return event as EmailReceivedEvent`. The three fetch/identity keys *are* checked (good, and proven by the no-transport assertion), but the remaining ~9 fields are promised by the type and unverified. An event with `data.from = 12345` and no `data.to` is accepted, after which `email.event.data.to` is typed `string[]` and is `undefined` — a lying type on a public `readonly` field.
- **M5 — `replyTargets` is a preview with no way to pin the destination.** The docs are honest that the server re-resolves, which is the problem: a caller who inspects `replyTargets`, judges it safe, and calls `.reply()` has no `to` override on `ReplyInput`. Two real divergences from `outbound/reply.go:40-47`: the server parses `From:` as an address *list* (multi-address From → server replies to all, facade previews one), and when `raw_message` is empty it falls through to the stored `Sender` (`reply.go:36-38`), a branch the facade cannot model.
- **M6 — `authenticated_from` is now a hard local reject but is not guaranteed non-empty on the wire.** No `omitempty`, and with `MAIL FROM:<>` plus an unparseable From header both sources are `""`. `fromEvent` then throws where `webhooks.fetchMessage` succeeds — a strictness the wire contract does not back.
- **M7 — `toJSON()` throws `RangeError` on a bad timestamp.** `inbound.ts:182` + `:214`: an unparseable `created_at` yields `Invalid Date` at construction and detonates later inside serialization, far from the cause.
- **M8 — Python parity: no immutability.** TS hardens the trust object (`readonly` + `Object.freeze` on the arrays, `inbound.ts:141-186`); Python ships plain mutable attributes, so `email.verified = True` and `email.attachments.append(...)` both work on a security-facing projection.
- **M9 — Python sync `InboundEmail` has two sources of truth.** `__init__` copies every field onto `self` (`inbound.py:262-284`) but `to_dict()` delegates to `self._target.to_dict()` (`:303`); attribute reads and serialization can diverge. Same in `InboundAttachment.to_dict` (`:241`). The async class has no such split.

---

### Low

- Design/impl drift: design `:79` types `conversationId?: string`, implementation and `MessageView.ts:27` make it required; design `:165` calls the attachment field `size`, actual is `sizeBytes`/`size_bytes`.
- `_validated_data` runs twice per email (`inbound.py:161`) — harmless, but the constructor can now raise a validation error *after* the HTTP fetch is billed.
- `AsyncInboundAttachment.to_dict()` passes `self` to `_attachment_dict(meta: AttachmentMetaView)` under `# type: ignore[arg-type]` (`inbound.py:149`) — duck-typing suppressed rather than modeled.
- `inbound.test.ts:61-62` asserts `JSON.stringify(email)` lacks the strings `"rawMessage"`/`"downloadUrl"`. `toJSON()` enumerates snake_case keys, so those camelCase names can never appear — near-tautological, and it never checks the actual base64 payload from `full.json:53` is absent.
- `WSEvent` is never tested in either Python suite; `typecheck_inbound.py` uses only `WebhookEvent`. Since Python (unlike TS) has two structurally-distinct envelope types, this is the parity gap most worth closing. Verified working today — just unguarded.
- No sync-path test that `from_event` raises `E2AValidationError`; `invalid.json` has no case for `data` as a list or a missing `schema_version`.
- `test_v1_inbound.py:52` constructs an `AsyncE2AClient` and never closes it.
- 409 `message_not_yet_delivered` is still unmentioned anywhere, though `pending_review` is now handled well.
- The `.message` point-in-time-snapshot caveat exists only at design `:58-59`; it reached no shipped surface (grep finds nothing in either SDK README, root README, `web/public/sdk.md`, `docs/events.md`, or the sources).
- `adversarial.json` diverges `content_type` between event (`text/plain`) and message (`application/x-executable`) — a genuinely interesting attack — but nothing asserts on it; it reads as fixture noise.
- Naming rule still unstated: prefixed (`AsyncInboundResource`) and unprefixed conventions now coexist in Python.

---

### Cleared / confirmed good

- **Spec + codegen are clean.** The `api/openapi.yaml` diff is exactly two `description:` lines — no shape, enum, `required`, or `type` change, so `openapi-compat-check` passes. Descriptions byte-match the new Go `doc:` tags, so `spec-check`/`TestSpecGoldenNoDrift` pass. Generated models changed only the corresponding docstrings, with no hand-edits a generator would not emit.
- **The Python 3.9 concern is dead** — no `fromisoformat` in the new code; timestamps come from pydantic. Verified by compiling every new file at `_feature_version=9`: no 3.10+ syntax.
- **Sync parity is genuinely tested**, not assumed — `test_v1_sync_client.py:148-171` drives the sync path through real `httpx_mock` responses against the same shared vector as the async test.
- `invalid.json`'s 14 vectors are really exercised in both suites with zero-transport assertions.
- Attachment semantics (`get()` returns a URL not bytes; uncached; lazy) match the design explicitly and are correctly documented. The `fetch` → `get` rename addresses the prior naming complaint.
- Public TS export surface is clean; `client.inbound` correctly shares `this.messages`.
- `toJSON`/`__repr__` eliding `downloadUrl` is right.

---

### Judgment calls, not defects

1. **#14 surface duplication** — `webhooks.fetchMessage` is retained with no deprecation marker, defended at design `:65-68`. Your call, but it is now three permanent ways to hydrate an event, frozen at GA.
2. **`test_exports.py:107`** drops `InboundEmail` from the retired-surface assertion list, so code importing the legacy flat-SDK `InboundEmail` now silently receives a structurally different object instead of `ImportError` — filed under "Added" on a minor bump.
3. **#6 screening ceiling** — `scan_action`/`scan_score` remain server-unreachable; recorded as a non-goal and documented in all three shipped docs. That is the honest ceiling without a server change, but it means the piguard `flag` tier is still only observable via `flagged`.

<sub>Re-reviewed by 4 parallel agents: regression verification against all 18 prior findings, TS implementation (with mutation testing), Python implementation (tests + mypy executed), and security + spec. B1 and B2(a) were verified by hand against `internal/relay/server.go` after two agents disagreed.</sub>

PR #594, opened 2026-07-20
Token Canopy branding + agent-native marketing page, plus a review-page crash fix

jiashuoz:
## Summary

Makes e2a a Token Canopy product and, along the way, fixes a user-facing crash and two docs/snippet inaccuracies found while verifying the work.

Six commits, each independently reviewable:

1. **`feat(design)`** — the Token Canopy palette + typography. Canopy green leads, warm gold is the accent, Inter + Fraunces replace Geist + Instrument Serif. Ember is gone repo-wide including the Go-served HITL page, the notification email, the Redoc theme and the brand SVGs. Foundation neutrals, semantics, radii and shadows are unchanged — a re-accenting, not a new palette.
2. **`fix(web)`** — the review page crashed with `Cannot read properties of undefined (reading 'conversation_id')`. Two surfaces cached different shapes under one SWR key; the cache now always holds the raw wire and each surface projects at point of use.
3. **`feat(web)`** — in-app attachment viewer (images + PDFs), and outbound attachments now surface at all.
4. **`refactor(web)`** — deletes the Topbar, whose only content app-wide was a breadcrumb repeating the nav tab and the page H1. The sidebar becomes the single identity anchor: Token Canopy over e2a.
5. **`feat(web)`** — repositions the marketing page around agent-native onboarding (plugin / MCP first, SDKs second).
6. **`docs`** — corrects the Cursor install instructions, which were wrong in both clauses.

## Client surface checklist

No API, schema, or client-surface change — this is web, design tokens, two Go HTML/email templates (string constants only), and docs. The rows below are deliberately skipped:

- Go handler / migration / OpenAPI / SDK regen / CLI / MCP — **not applicable.** The only Go edits are colour and font-family constants inside `hitl_magic_api.go`'s inlined CSS and `notifier.go`'s email template. No handler, no route, no wire shape.
- [x] Tests at each surface touched, including regression cases for the crash and the attachment viewer.

## Operational risk

Low, and no data or auth paths are touched.

- **User-visible chrome changes.** The Topbar is gone on every page and breadcrumbs with it. Deep pages (`/templates/edit`, `/reviews/review`) lose their only crumb trail; they keep in-page back affordances and the browser's. Worth a look if any screen feels stranded.
- **The crash fix changes a cache key** (`pending-message` → `message-detail`, now keyed by id alone). No persistence involved — worst case is one extra fetch after deploy.
- **Attachment previews fetch full bytes** into a blob URL on open, revoked on unmount. Previously the bytes only moved on an explicit download.
- **The HITL approval page and notification email are customer-facing** and change colour/typeface. Both were verified rendering locally against Mailpit.
- No feature flags; rollback is a revert.

## Test plan

- [x] `npm test` in `web/` — 416 passing (51 suites), including new regression coverage
- [x] `npx tsc --noEmit` clean; `npm run lint` unchanged at one pre-existing warning; `npm run build` (static export) clean
- [x] `go build ./...`; `internal/agent` + `internal/hitlnotify` pass on a dedicated test DB
- [x] `@e2a/ui` rebuilt so `design-system/dist` matches `src`
- [x] Crash reproduced in the browser before the fix and confirmed gone after, in both cache orderings
- [x] Attachment viewer exercised live with seeded mail: inline `cid:` image, PDF, and a non-previewable CSV falling back correctly
- [x] Marketing snippets verified symbol-by-symbol against the SDK sources in this tree (5.2.0)
- [x] Cursor MCP instructions checked against Cursor's docs; the server's `registration_endpoint` confirmed live
- [ ] Dark mode spot-check across the app after merge (checked on the sidebar and inbox only)
- [ ] Confirm the HITL email renders acceptably in a real client, not just Mailpit

## Known gaps

- **The marketing snippets target the SDKs in this tree (5.2.0), but npm/PyPI currently publish 4.0.1**, where `listen()` yields flat notifications rather than `WSEvent` envelopes and `isEmailReceived` doesn't exist. Intentional per discussion — 4.0.1 is being deprecated — but the samples won't run until 5.2.0 ships.
- `web/src/app/components/loft/PageShell.tsx` has its crumb removal in commit 1 rather than commit 4, because both edits landed in that file before the first commit. Only matters if these get split into separate PRs.
- Branched fresh off `main` (`1e4df382`), so it already includes #592.
- The two manifests disagree on MCP tool count (marketplace says 37, plugin says 50). Untouched here.


web/src/app/(app)/reviews/_components/PendingRow.tsx
```diff
@@ -89,16 +89,19 @@ export function PendingRow({
...
+  // same entry, so both must agree on the shape (see lib/swrKeys.ts) —
+  // and projects below. Uses the review endpoint, the only one that
+  // carries `hold_reason`/`protection`.
+  const { data: wire, error: fetchError, isLoading } = useSWR(
+    expanded ? messageDetailKey(id) : null,
+    () => getReviewDetailWire(id),
     {
       keepPreviousData: false,
-      onSuccess: (data) => {
+      onSuccess: (w) => {
```
jiashuoz: When another surface recently populated `messageDetailKey(id)`, SWR returns cached data without invoking `onSuccess`, so these form fields remain empty. `handleApprove` nevertheless diffs those empty fields against `msg` even when Edit was never opened, producing overrides that clear the subject, body, and recipients. Initialize from `wire` in an effect—as the focus page now does—or only compute overrides while `editing` is true.

web/src/app/page.tsx
```diff
@@ -880,6 +940,74 @@ export default function Home() {
...
 // ─────────────────────────────────────────────────────────────────────────
 
+// A hover-opened nav dropdown. Each menu owns its open state so adding one
+// doesn't mean threading another boolean through the page component.
+function NavMenu({ label, items }: { label: string; items: NavItem[] }) {
+  const [open, setOpen] = useState(false);
+  return (
+    <div
+      className="relative"
+      onMouseEnter={() => setOpen(true)}
```
jiashuoz, line 956: The new Product and Resources menus open only from mouse enter. Focusing or activating the button with a keyboard does nothing, leaving every dropdown link unreachable; touch devices with desktop-width layouts are also unreliable. Support click/focus and keyboard dismissal, ideally with standard menu-button focus management.

jiashuoz:
Both review comments confirmed and fixed in c04407d9 — thanks, the first one was a genuine data-loss path.

**1. Approving from a warm cache blanked the message** — valid, and worse than "fields remain empty". Reproduced it before fixing: the approve POST body came back with `subject: ""`, `text: ""`, `to: []`. Since `edits.ts` documents that the server treats a present field as "use this value, including empty string → clear the field", approving a cached row sent a message with no subject, no body and no recipients.

Worth noting this was a second-order effect of the cache fix in this same PR: before it, a cross-surface cache hit *crashed* the focus page, which masked the empty-form path. Making the shared key safe turned a loud crash into a silent destructive one.

Fixed as suggested — seeding moved to an effect on the projected draft so it runs on cache hits, guarded by a ref so background revalidation can't stomp an in-progress edit. Took the second suggestion too: an editor that was never opened now produces no overrides at all. Regression test seeds `messageDetailKey` against the module-level SWR cache (so `onSuccess` genuinely doesn't fire), approves, and asserts no destructive overrides; it fails on the pre-fix code.

**2. Hover-only nav menus** — valid, and a reachability *regression* rather than a pre-existing gap: Human-in-the-loop, Use cases and Blog were top-level keyboard-reachable links before this PR regrouped them behind these menus. They now open on hover, click and focus, close when focus leaves the menu subtree (checked via `relatedTarget` so focus moving button→link doesn't close it), and dismiss on Escape. `aria-haspopup` added alongside the existing `aria-expanded`. Two tests cover keyboard open/Escape and click-toggle.

419 tests passing, typecheck/build clean, lint unchanged at the one pre-existing warning.

PR #636, opened 2026-07-22
fix(mcp): advertise https for public hosts in discovery metadata

### LINKED ISSUE #635: Hosted MCP OAuth broken: oauth-protected-resource returns http:// for the https endpoint
## Summary

The hosted MCP server's RFC 9728 protected-resource metadata advertises the
`resource` identifier over **`http://`** even though the endpoint is served over
`https://`. OAuth-capable MCP clients (Claude Code, Cursor, MCP Inspector)
correctly reject the scheme mismatch, so **nobody can authenticate to the hosted
MCP server** (`https://api.e2a.dev/mcp`).

Discovered while testing the new `send_at` (scheduled send) feature end-to-end
through the MCP tools — the flow never got past the OAuth handshake.

## Repro

Install the `e2a` plugin in Claude Code and try to authenticate the MCP server.
The auth probe fails with:

```
SDK auth failed: Protected resource http://api.e2a.dev does not match
expected https://api.e2a.dev/mcp (or origin)
```

## Live evidence

```
$ curl -s https://api.e2a.dev/.well-known/oauth-protected-resource
{"resource":"http://api.e2a.dev",
 "authorization_servers":["https://api.e2a.dev"],
 "scopes_supported":["agent","account"],
 "bearer_methods_supported":["header"]}
```

Note that `authorization_servers` is correctly `https://` (it is built from the
static `baseUrl` config), while `resource` is `http://` (it is synthesized from
the inbound request scheme). Same host, two different schemes — the client
requires `resource` to match the connection origin (`https://api.e2a.dev`), so
it aborts.

## Root cause

In `mcp/src/http-server.ts`, `resolveResourceUrl()` synthesizes the resource URL
as `${externalScheme(req)}://${host}` when `MCP_PUBLIC_URL` is unset, and
`externalScheme()` returns `https` only when `req.protocol === "https"`.
`req.protocol` honors `X-Forwarded-Proto` **only for proxy hops trusted by the
Express `trust proxy` setting** (`E2A_TRUST_PROXY`, default `"loopback"`).

The production deployment behind its TLS-terminating proxy is landing on the
`http` fallback — either the fronting proxy hop is not loopback-adjacent (so
`X-Forwarded-Proto: https` is ignored) or `MCP_PUBLIC_URL` is simply unset.

## This is NOT deploy lag

Both candidate images already contain the relevant scheme fix (#92, landed
2026-06-08):

| Tag | Built | Revision | Contains #92? |
|-----|-------|----------|---------------|
| `:latest` (`v1.1.0`) | 2026-07-21 | `c6537982` | yes |
| `:main` | 2026-07-22 | `8955b889` | yes |

So the running code can already emit `https://`; it just isn't being given the
information to do so at runtime.

## Impact

All OAuth-based MCP clients fail to authenticate against the hosted server.
Bearer-token flows that skip discovery are unaffected, but the plugin / IDE
onboarding path is broken.

## Proposed fix

1. **Immediate (ops):** set `MCP_PUBLIC_URL=https://api.e2a.dev` on the hosted
   MCP deployment. `publicUrl` is used verbatim and bypasses all proxy-trust
   guesswork.
2. **Code hardening (PR to follow):** when `MCP_PUBLIC_URL` is unset, the
   synthesized discovery URL should default to `https://` for a non-loopback
   public host — a publicly reachable OAuth-protected resource is always
   TLS-fronted, and the current `http` fallback is a silent footgun that just
   broke prod. This complements #92 (loopback dev over plain `http` still works
   because local dev sets `publicUrl` explicitly).

cyj-git-0825:
## What

When `MCP_PUBLIC_URL` is unset, the synthesized RFC 9728 discovery URL now
forces `https` for any non-loopback host, instead of deriving the scheme from
`req.protocol`.

Fixes #635.

## Why

The hosted server serves `https://api.e2a.dev/mcp` but was advertising:

```
$ curl -s https://api.e2a.dev/.well-known/oauth-protected-resource
{"resource":"http://api.e2a.dev",            ← http
 "authorization_servers":["https://api.e2a.dev"], ← https
 ...}
```

`resource` is synthesized from the request scheme, which is only `https` when
the fronting proxy's `X-Forwarded-Proto` is trusted (`E2A_TRUST_PROXY`, default
`loopback`). In prod that isn't the case, so it fell back to `http`, and every
OAuth-capable MCP client (Claude Code, Cursor, MCP Inspector) rejects the
`http` vs `https` scheme mismatch — nobody can authenticate to the hosted MCP
server. (`authorization_servers` was already correct because it comes from the
static `baseUrl` config, not the request.)

Discovered while testing the new `send_at` (scheduled send) feature end-to-end
through the MCP tools — the flow never got past the OAuth handshake.

## Approach

A publicly reachable MCP endpoint is always TLS-fronted, so a non-loopback host
should never be advertised over `http`. This forces `https` for public hosts
while leaving loopback dev untouched:

- Loopback (`localhost` / `127.0.0.1`) still derives the scheme from the trusted
  request protocol, so `http://localhost` dev keeps working. Local dev also sets
  `publicUrl` explicitly, which bypasses synthesis entirely.
- This **complements #92** (trusted `X-Forwarded-Proto` derivation) rather than
  reverting it — it just removes the silent `http` footgun for public hosts when
  the proxy-trust config isn't perfectly aligned.

Setting `MCP_PUBLIC_URL=https://api.e2a.dev` on the deployment is still the
recommended belt-and-suspenders remedy; this change makes the server correct by
default even without it.

## Tests

- New: `synthesizes https for a public host without X-Forwarded-Proto (#635)`.
- All existing discovery-scheme tests (loopback fallback, trusted/untrusted
  X-Forwarded-Proto, `publicUrl` override) unchanged and still pass.
- Full MCP suite: 212 passing; `tsc` build + type-tests clean.

### DIFF
--- a/mcp/src/http-server.ts
+++ b/mcp/src/http-server.ts
@@ -16,10 +16,11 @@ export interface HttpServerOptions {
    * Externally reachable URL of this MCP server. When set, used verbatim
    * for the RFC 9728 protected-resource metadata `resource` field and
    * the `WWW-Authenticate: resource_metadata=...` value on 401s. When
-   * unset (production default behind Caddy), we synthesize
-   * `https://{Host}` from the inbound request. The local-dev runbook
-   * sets this to `http://localhost:8765` so Claude Code's OAuth probe
-   * can resolve the metadata over loopback http.
+   * unset (production default behind Caddy), we synthesize the URL from
+   * the inbound request Host, forcing `https` for any non-loopback host
+   * (a public MCP endpoint is always TLS-fronted — see #635). The
+   * local-dev runbook sets this to `http://localhost:8765` so Claude
+   * Code's OAuth probe can resolve the metadata over loopback http.
    */
   publicUrl?: string;
   /**
@@ -141,7 +142,7 @@ export function buildApp(opts: HttpServerOptions): BuiltApp {
     if (!host) return null;
     const bare = host.split(":")[0]!.toLowerCase();
     if (!allowedHosts.has(bare)) return null;
-    return `${externalScheme(req)}://${host}`;
+    return `${discoveryScheme(req, host)}://${host}`;
   };
 
   app.get("/.well-known/oauth-protected-resource", (req, res) => {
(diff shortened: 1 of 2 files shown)


jiashuoz (approved):
Verified the public-host HTTPS synthesis and added regression coverage for both protected-resource discovery and the WWW-Authenticate resource_metadata challenge. MCP workspace: 212 tests passing; build and type checks clean.

PR #664, opened 2026-07-23
feat: scheduled send via send_at

cyj-git-0825:
## Scheduled send (`send_at`)

This adds scheduled sending to e2a. `send`, `reply`, and `forward` now accept an optional RFC 3339 `send_at` timestamp. When `send_at` is in the future, the message is accepted immediately and returned with `status=scheduled` (plus the resolved `scheduled_at`), then submitted to the provider at approximately that instant — "not before", accurate to the scheduler's poll interval rather than exact to the millisecond. A `send_at` at or before now behaves exactly like an ordinary immediate send, and a `send_at` more than 90 days out is rejected with `400 invalid_request`.

### What it builds on

The work is built on the latest `main` and integrates cleanly, with no conflicts.

### How it works, and what did (and didn't) change

Outbound send is already queue-first on River, so scheduling reduces to *when the existing send job runs*. The accept transaction enqueues the same `outbound_send` job using River's `InsertOpts.ScheduledAt` and records the requested instant on the message; a future-scheduled job simply waits in River's `scheduled` state until it fires.

Deliberately, most of the pipeline did **not** need to change. The message stays `delivery_status='accepted'` (no new delivery status is introduced), the terminal reconciler already ignores jobs sitting in `scheduled` state, and claim/rollup, usage metering (still counted at fire time), and the approve/reconcile callback signatures are all untouched. The only new logic is small edge validation: cap `send_at` at 90 days and treat a past instant as immediate (clock-skew tolerant).

Two product decisions for this first cut: a scheduled send is cancelled by trashing the message (the existing claim-time trash-cancel already stops it from going out), and scheduling does not survive a review hold (if held, the message sends when it is approved).

### How it fits with the rest of the surface

The feature is wired through every client, not just the HTTP API:

- **TS & Python SDKs** — the generated bases carry the new request field (`sendAt` / `send_at`) and the `scheduled` result, so scheduling works with no hand-written SDK glue.
- **CLI** — `send` and `reply` gain a `--send-at` flag, and a `status=scheduled` result is treated as a successful (exit 0) outcome with a note.
- **MCP** — `send_message`, `reply_to_message`, and `forward_message` accept a validated `send_at` and pass it through to the SDK, so an agent operating over MCP can schedule mail.

### Testing

- **Go** — the full build is clean, and the handler/edge unit tests pass: future → `scheduled` (and `wait=sent` correctly skips its poll loop), past → immediate, beyond-horizon → `400`, plus the `scheduledInstant` normalization unit. The DB-backed accept test (persists the instant, enqueues a scheduled job, stays `accepted`) compiles under the integration build tag.
- **SDKs / MCP** — `@e2a/sdk` and the MCP server type-check and build cleanly. The OpenAPI spec regenerates byte-identical to what the live handlers emit, and both SDK bases regenerate cleanly from that spec.
- **End-to-end** — running this branch in Docker against a Mailpit catch-all: a message scheduled two minutes out returned `202 status=scheduled` with `scheduled_at` echoed, was absent from the mailbox before its time, and was delivered just after it (never early); a past `send_at` was delivered immediately; a `send_at` beyond 90 days was rejected with `400`.

The branch is based on the latest `main`, integrates with no conflicts, and everything above is green.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -1886,6 +1886,10 @@ components:
           description: Sets the Reply-To header — where replies to this message are directed. A single RFC 5322 address, optionally with a display name. At most 320 characters (display name + address combined). Defaults to the sending agent's own address.
           maxLength: 320
           type: string
+        send_at:
+          description: Optional scheduled-send time (RFC 3339 with a UTC offset). When set to a future instant the forward is accepted immediately and returns status=scheduled; it is submitted at approximately this time ("not before", accurate to the scheduler poll interval). A value at or before now sends immediately. Must be no more than 90 days ahead (over → 400 invalid_request). Scheduling does not survive a review hold. Cancel by moving the message to trash.
+          format: date-time
+          type: string
         text:
           maxLength: 1048576
           type: string
@@ -2144,6 +2148,9 @@ components:
         scan_score:
           format: double
           type: number
+        scheduled_at:
+          format: date-time
+          type: string
         sent_as:
           type: string
         size_bytes:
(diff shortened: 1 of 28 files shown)


cyj-git-0825:
Superseded by a fresh PR from the same branch (clean history; all CI green and Copilot review items addressed). Closing this one.

PR #665, opened 2026-07-23
feat: scheduled send via send_at

cyj-git-0825:
## Scheduled send (`send_at`)

This adds scheduled sending to e2a. `send`, `reply`, and `forward` now accept an optional RFC 3339 `send_at` timestamp. When `send_at` is in the future, the message is accepted immediately and returned with `status=scheduled` (plus the resolved `scheduled_at`), then submitted to the provider at approximately that instant — "not before", accurate to the scheduler's poll interval rather than exact to the millisecond. A `send_at` at or before now behaves exactly like an ordinary immediate send, and a `send_at` more than 90 days out is rejected with `400 invalid_request`.

### What it builds on

The work is built on the latest `main` and integrates cleanly, with no conflicts.

### How it works, and what did (and didn't) change

Outbound send is already queue-first on River, so scheduling reduces to *when the existing send job runs*. The accept transaction enqueues the same `outbound_send` job using River's `InsertOpts.ScheduledAt` and records the requested instant on the message; a future-scheduled job simply waits in River's `scheduled` state until it fires.

Deliberately, most of the pipeline did **not** need to change. The message stays `delivery_status='accepted'` (no new delivery status is introduced), the terminal reconciler already ignores jobs sitting in `scheduled` state, and claim/rollup, usage metering (still counted at fire time), and the approve/reconcile callback signatures are all untouched. The only new logic is small edge validation: cap `send_at` at 90 days and treat a past instant as immediate (clock-skew tolerant).

Two product decisions for this first cut: a scheduled send is cancelled by trashing the message (the existing claim-time trash-cancel already stops it from going out), and scheduling does not survive a review hold (if held, the message sends when it is approved).

### How it fits with the rest of the surface

The feature is wired through every client, not just the HTTP API:

- **TS & Python SDKs** — the generated bases carry the new request field (`sendAt` / `send_at`) and the `scheduled` result, so scheduling works with no hand-written SDK glue.
- **CLI** — `send` and `reply` gain a `--send-at` flag, and a `status=scheduled` result is treated as a successful (exit 0) outcome with a note.
- **MCP** — `send_message`, `reply_to_message`, and `forward_message` accept a validated `send_at` and pass it through to the SDK, so an agent operating over MCP can schedule mail.

### Testing

- **Go** — the full build is clean, and the handler/edge unit tests pass: future → `scheduled` (and `wait=sent` correctly skips its poll loop), past → immediate, beyond-horizon → `400`, plus the `scheduledInstant` normalization unit. The DB-backed accept test (persists the instant, enqueues a scheduled job, stays `accepted`) compiles under the integration build tag.
- **SDKs / MCP** — `@e2a/sdk` and the MCP server type-check and build cleanly. The OpenAPI spec regenerates byte-identical to what the live handlers emit, and both SDK bases regenerate cleanly from that spec.
- **End-to-end** — running this branch in Docker against a Mailpit catch-all: a message scheduled two minutes out returned `202 status=scheduled` with `scheduled_at` echoed, was absent from the mailbox before its time, and was delivered just after it (never early); a past `send_at` was delivered immediately; a `send_at` beyond 90 days was rejected with `400`.

The branch is based on the latest `main`, integrates with no conflicts, and everything above is green.


jiashuoz (commented):
Really nice PR — the design leans on the existing queue-first/River machinery cleanly, and the tricky interactions (terminal reconciler leaving `scheduled` jobs alone, the startup reconciler not double-enqueuing since `send_job_id` is stamped in the accept-tx, `wait=sent` skipping the poll for a scheduled send, and the HITL hold returning before the scheduled accept-tx) are all handled and mostly test-covered. No correctness or security concerns from me.

One product gap I'd like to see addressed before/with this feature: **a user can't see *if* or *when* a message is scheduled from the web dashboard.**

Because the change deliberately introduces no new `delivery_status` (a scheduled message stays `delivery_status='accepted'`), the dashboard's `MessageStatusChip` (`web/src/app/components/messages/MessageStatusChip.tsx`) hits its `case "accepted"` branch and renders **"Queued"** — visually identical to an ordinary immediate send, with no indication it's deferred and no send time shown. The string `scheduled` appears nowhere in the web message components.

There's also a backend gap underneath it: `scheduled_at` is only exposed on the detail `GET` (`GetMessageWithContent` / `MessageView`), **not** on the list endpoint — `MessageSummaryView` has no `scheduled_at` field and the list query doesn't select it. So even a dashboard that wanted to show "Scheduled" in the inbox couldn't, since the list rows carry neither a distinguishing status nor the timestamp.

To let users see if/when a message will send, I think this needs:

1. **Expose `scheduled_at` on the list contract** — add it to `MessageSummaryView` + the list query, then `make generate` to refresh `api/openapi.yaml` and both SDK bases. (Contract change, so the drift gates will require the regen.)
2. **Web** — add `scheduled_at` to the message type in `web/src/app/components/types.ts`, and have `MessageStatusChip` render a distinct **"Scheduled"** chip when `scheduled_at` is in the future (rather than "Queued"), plus surface the time in the thread row/bubble (e.g. "Sends Aug 1, 9:00 AM").
3. Tests for the chip branch and the list-view serialization.

Happy to pair on the web side if useful. Everything else looks great.

---
_Generated by [Claude Code](https://claude.ai/code)_

cyj-git-0825:
**Status update — scheduled-visibility gap (PR #665)**

Split this into a backend half (done here) and a frontend half I'd like to pair with you on 👀.

**Done in this PR (backend / contract):**
- `scheduled_at` is now exposed on the list endpoint too (`MessageSummaryView` + the list query), not just the detail view — so the inbox list carries both the timestamp and enough to tell a scheduled send apart. Regenerated the spec + both SDK bases; added tests (list serialization + summary/detail consistency). CI's green.

**Left for us to pair on (frontend):**
- The "Scheduled" chip + surfacing the send time. It's your design system, so I'd rather settle the semantics with you than guess.

**Open questions (to make pairing quick):**
1. **Chip tone** — `info` (same blue family as "Queued", distinguished by the label + a clock icon) or `accent` (your signal color, distinct from "Queued" at a glance)? Your call.
2. **Chip priority** — I'm assuming: `review status > sent/sending/failed > (accepted && scheduled_at in the future → "Scheduled") > "Queued"`. So terminal states win (`scheduled_at` stays on the row after it fires), a past scheduled time that's still `accepted` falls back to "Queued", and a held send stays "Pending review". Match your expectations?
3. **Time** — absolute "Sends Aug 1, 9:00 AM" in the viewer's local timezone, shown in both the inbox row and the detail bubble. OK?
4. **Scope** — display-only this round; cancel/reschedule from the dashboard deferred to a follow-up (cancelling already works via trash).

jiashuoz (commented):
Review pass on the scheduled-send implementation — the overall design (schedule = `InsertOpts.ScheduledAt` on the existing outbox job) looks right, and the accept transaction, idempotent replay, reconciler safety, and hold behavior all check out as described. A few items to confirm and, if they hold, address before merge:

## 1. Retry horizon is measured from accept time — long-scheduled sends terminally fail on their first fire attempt (blocker)

`internal/outboundsend/worker.go:109` (`pastRetryHorizon`) compares `time.Since(j.AcceptedAt)` against the 72h `sendRetryHorizon`, and `AcceptedAt` is `messages.created_at` (`internal/agent/outbound_async.go:141`) — the *accept* instant. With `send_at` up to 90 days out, a send scheduled >72h ahead that hits a sending-ramp deferral (`worker.go:309`) or a provider outage (`worker.go:394`) at fire time takes the past-horizon branch on **attempt 1** → terminal `failed` + `email.failed`, zero snoozes, where an immediate send would get up to 72h of tolerance. Ramp deferral is a routine condition for ramped domains, so this will produce false `email.failed`s at fire time exactly when the pipeline is designed to ride it out.

Suggest: start the horizon clock at `max(created_at, scheduled_at)` — the `scheduled_at` column already exists, so threading it into `SendJob` should be small.

## 2. Monthly quota is bypassable by scheduling into future months (should-fix or explicit product decision)

Quota enforcement (`internal/limits/enforcer.go:154`) counts at **accept** time; metering happens at **fire** time, and `RecordAndCheckTx` never refuses. So a scheduled accept doesn't count against any month's cap — a user can schedule an unbounded number of messages (modulo the accept rate limit and storage cap) into future billing periods, and they all deliver, merely recorded. "Metered at fire time" currently means "uncapped at fire time." Options: enforce/refuse at fire in the MarkSent path, or count scheduled-but-unfired messages at accept. At minimum this wants a conscious documented decision.

## 3. Self-send loopback silently ignores `send_at` (should-fix)

The `isSelfSend` branch (`internal/agent/api.go:1302`) runs before the scheduling logic, and `performSelfSend` never reads `req.ScheduledAt` — a send to the agent's own address with a future `send_at` delivers immediately and returns `sent`, dropping the scheduling intent with no error. Suggest rejecting `send_at` on the self-send path (400) or documenting the exception in the field docs.

## 4. Conversation view drops `scheduled_at` (should-fix)

`GetConversationByID` (`internal/identity/store.go:4355`) wasn't updated to select `m.scheduled_at`, so `GET /v1/conversations/{id}` omits the field while message list/detail return it — a third projection of the same summary contract that the new parity test doesn't cover. Two-line fix (SELECT + Scan).

## Smaller items

- CLI `parseSendAt` uses raw `new Date()`: date-time without offset parses as local time, date-only as UTC midnight, loose formats accepted — while the MCP tool is strict (`datetime({offset: true})`). The CLI should require an explicit offset too.
- Trash-cancel is reversible: restoring a trashed scheduled message before `send_at` re-arms it (the River job is never cancelled, only claim-gated). Consistent with existing trash semantics, but "cancel by trashing" reads as final — document it or actively cancel/reschedule on trash/restore.
- Domain verification is only checked at accept; nothing re-validates at fire up to 90 days later. Probably fine if verification can't lapse independently of agent deletion (claim-time cancel covers that), but worth a documented rationale.
- Replying to/forwarding your own scheduled message 409s for the whole schedule window (`parentNotYetSubmitted`) with a "retry after it is sent" error that doesn't mention the parent is scheduled — could surface `scheduled_at` in the error.
- MCP `get_message`/`get_review` don't surface `scheduled_at` (`messageViewForTool` enumerates fields explicitly) — one-line pass-through so MCP clients can read back when a scheduled send fires.
- No e2e-prod/conformance coverage; a cheap `send_at = now+5s` case would lock the real River scheduled→available promotion path that no unit test here exercises.

Item 1 is the only one I'd treat as merge-blocking. Thanks — the rest of the pipeline reuse is genuinely clean.

cyj-git-0825:
Thanks for the thorough review, Josh — learned a lot here; these were mostly production/operational angles I'd underweighted. All addressed except one that's blocked on timing (below). Per-item:

**Blocker**
- **1. Retry horizon from accept time** — fixed: the horizon is now measured from `max(created_at, scheduled_at)`, so a long-scheduled send gets its full outage tail from fire time instead of failing on attempt 1. Threaded `scheduled_at` through `OutboundSendPayload` → `SendJob`. Unit test added. (`6543b21d`)

**Should-fix**
- **2. Monthly quota bypassable via future scheduling** — enforced at fire: for scheduled sends, `ClaimSend` checks the cap in the fire month and refuses over-cap sends terminally (`MarkFailed`, `email.failed`) rather than delivering uncapped. Wired from the limits enforcer via a setter; transient lookup errors fail open. DB integration test added. (`6543b21d`)
- **3. Self-send ignores `send_at`** — now rejected with `400 invalid_request` (the loopback path delivers immediately and can't schedule). DB test added. (`6543b21d`)
- **4. Conversation view drops `scheduled_at`** — `GetConversationByID` now selects + scans it; DB parity test added. (`6543b21d`)

**Smaller items**
- CLI `parseSendAt` now requires an explicit RFC 3339 offset, matching the MCP tool. (`6543b21d`)
- MCP `get_message`/`get_review` pass `scheduled_at` through (`messageViewForTool`). (`6543b21d`)
- Reply/forward to a still-scheduled parent → the 409 now names the scheduled instant instead of the bare "retry after it is sent". (`31cc8695`)
- Trash-cancel reversibility documented (restoring before `send_at` re-arms; the job is claim-gated, not cancelled) across the API field docs, MCP, CLI, and the claim-time comment. (`31cc8695`)
- Domain-verification-at-fire rationale documented: `DeleteDomain` is blocked while agents exist, and agent deletion trips the claim-time cancel, so a verified domain can't lapse independently before fire. (`31cc8695`)

**Deferred (timing)**
- e2e-prod `send_at = now+5s`: the suite targets the deployed prod server, which won't have scheduled send until this ships — so the case can only pass post-release. I'll add it as a follow-up once deployed. The scheduled→available path is covered by unit + DB tests in the meantime (minus the real River timer, which is exactly what that e2e would lock).

**Also** (the visibility gap from your first pass) — the dashboard now shows a distinct **"Scheduled"** chip (existing `info` tone + a clock glyph) with the send time in the inbox row and thread bubble; `scheduled_at` is on the list contract. (`34656d63`)

jiashuoz (approved):
Reviewed API/SDK/CLI/MCP contract and interface changes; conformance, generated-code, e2e, coverage, and all CI checks pass.

PR #670, opened 2026-07-24
feat(cli): read-only `e2a doctor` diagnostics command

jiashuoz:
## Summary

Adds `e2a doctor` — a strictly read-only diagnostic command for the production email path, per the production-trust strategy (§6.3 / roadmap "Ship `e2a doctor` with human and JSON output"). It diagnoses configuration and the production path end-to-end while guaranteeing it never sends mail and never mutates server state, DNS, or webhooks.

**Checks** (18 stable IDs): CLI config + credential scope · API connectivity/version (`GET /v1/info`) · key validity/scope (`GET /v1/account`) · agent existence/access · per registered custom domain, **live DNS lookups** of the server-prescribed records (ownership TXT, inbound MX, DKIM TXT, MAIL FROM MX, SPF with the server's inclusion semantics) plus an advisory warn-only DMARC check and the SES sending-status rollup · MCP endpoint reachability + advertised OAuth metadata (RFC 9728) · webhook config + recent delivery history · outbound SMTP env visibility for self-hosted operators (credential *presence* only — values never printed).

**Read-only guarantee**: only GETs and client-side DNS. Deliberately avoids `POST /v1/domains/{d}/verify` (mutates verification state + enqueues SES provisioning), `POST /v1/webhooks/{id}/test` (delivers a real event), and `POST /v1/agents/{email}/test` (sends real mail). Webhook reachability is an explicit `skip: no_safe_probe`, with delivery history as the observed signal.

**Output**: human `pass/warn/fail/skip` lines with `fix:` remediations; `--json` emits a versioned `e2a.doctor/v1` report (`id`, `status`, `reason_code`, `target`, `detail`, `evidence`, `remediation`) locked by two golden fixtures (regenerate with `UPDATE_DOCTOR_FIXTURES=1`).

**Exit codes** (additive to the frozen contract — nothing renumbered; values pinned by a new test): `0` healthy · `8` (new `EXIT.WARN`) warnings only · `9` (new `EXIT.CONFIG`) definite configuration failure · `4` auth failure · `1` transient connectivity · `2` usage. Precedence auth > config > transient > warn, each rung individually tested.

**Timeouts**: every network op bounded by 5s (`AbortSignal.timeout` for HTTP; `Resolver` timeout plus a hard per-lookup deadline race for DNS, since the resolver option is per-server/per-try).

The second commit applies fixes from an adversarial review pass: malformed `--mcp-url` crash (was: unhandled `new URL()` throw discarding the whole report), MCP 404 false-positive, SPF strict-equality false-negative, DNS deadline hardening, trailing-slash `E2A_URL` handling, and coverage for previously untested branches.

## Client surface checklist

- [x] CLI command or flag in `cli/src/commands/` + wired into `cli/src/bin/e2a.ts`
- [x] Tests at each surface above (positive + negative-path cases: 52 doctor tests + exit-code pinning)

> All other rows intentionally skipped: **no `/v1` API change** — doctor only consumes existing read-only endpoints, so there is nothing to mirror into the Go handlers, spec, SDKs, MCP, or dashboard. No migrations.

## Operational risk

None to the server — the command is client-side and read-only by construction (a test asserts no probe URL contains `/test` or `/verify`, and the review pass verified every SDK call maps to a GET). New exit codes 8/9 are additive; existing scripts branching on 0–7 are unaffected. Credentials are never echoed (test-asserted).

## Test plan

- [x] `npm test --workspace @e2a/cli` — 226 passed (15 files), including 52 doctor tests covering every check status (pass/warn/fail/skip) and every exit-code path (0/1/2/4/8/9)
- [x] `npm run build --workspace @e2a/cli` — clean
- [x] Golden JSON fixtures lock the `e2a.doctor/v1` schema
- [x] `scripts/check-repository-text-integrity.sh`, `scripts/check-sdk-version-sync.mjs` — pass
- [x] Live smoke vs production: healthy account → exit 8 with only genuine DMARC advisories across 6 custom domains (all live DNS checks pass); `--json` parses; `--mcp-url notaurl` → exit 2; `-h` short-circuits before any network call
- [ ] Post-merge: run `e2a doctor` from a machine with a stale/invalid key → expect exit 4

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,6 +245,7 @@ domains/webhooks in the **web dashboard**.
 |---------|-------------|
 | `e2a login` | Open a browser login and save an account-scoped API key to `~/.e2a/config.json` (does not set a default agent — use `e2a config set agent_email` or `--agent`) |
 | `e2a whoami` | Show the key identity: user, scope, bound agent, plan |
+| `e2a doctor` | Read-only diagnostics of the production email path: config, API, agent access, custom-domain DNS (live), MCP, webhooks, outbound SMTP visibility. Never sends mail or mutates anything; `--json` for a versioned report |
 | `e2a agents list\|create\|get` | Manage inboxes (requires an account-scoped key) |
 | `e2a keys create\|list\|delete` | Mint, list, and revoke API keys (requires an account-scoped key) |
 | `e2a protection get\|set` | Show or update an agent's HITL screening/review config |
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -1,5 +1,25 @@
 # Changelog
 
+## Unreleased
+
+**Added:** `e2a doctor` — read-only diagnostics for the production email path.
+Checks CLI config and credential scope, API connectivity (`GET /v1/info`),
+agent access, custom-domain DNS (live ownership TXT / MX / DKIM / SPF lookups
+against the server-prescribed records, plus an advisory DMARC check and the
+SES sending status), MCP reachability and advertised OAuth metadata, webhook
+configuration and delivery history, and outbound SMTP config visibility for
+self-hosted operators (credential presence only — values are never printed).
+It never sends mail and never mutates DNS, webhooks, or any other resource;
+in particular it does not call `POST /v1/domains/{domain}/verify` (which
+mutates verification state), `POST /v1/webhooks/{id}/test` (which delivers a
+real event), or `POST /v1/agents/{email}/test` (which sends real mail).
+`--json` emits a versioned `e2a.doctor/v1` report. Two exit codes were ADDED
+to the frozen contract (existing codes are unchanged): `8` — diagnostics
+completed with warnings only; `9` — diagnostics found a definite
+configuration failure. `0`/`1`/`4` keep their existing meanings for healthy,
+transient-connectivity, and authentication outcomes. Every network operation
+is bounded by a 5-second timeout.
+
 ## 2.0.0
 
 Current release. A major bump: the published 1.6.0 and this tree had diverged
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -5,10 +5,16 @@
 **Added:** `e2a doctor` — read-only diagnostics for the production email path.
 Checks CLI config and credential scope, API connectivity (`GET /v1/info`),
 agent access, custom-domain DNS (live ownership TXT / MX / DKIM / SPF lookups
-against the server-prescribed records, plus an advisory DMARC check and the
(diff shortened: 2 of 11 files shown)


jiashuoz:
Addressed the consolidated 4-agent review in 8dcb41b3. Every claim was first verified against the server source before changing anything.

**MAJOR — all fixed**
1. **DKIM semantics** — confirmed against `internal/agent/api.go` `classifyDKIM` + `internal/dkim/dkim.go` `ExtractPublicKeyFromTXT`: the server compares only the `p=` payload. Doctor now does the same (`extractDkimKey`, a direct port), so extra tags and reordering pass, and a differing `p=` reports `record_mismatch`. Covered by reorder-pass and mismatch tests.
2. **Trailing-slash E2A_URL** — confirmed the SDK uses `baseUrl` verbatim (`client.ts:210`). `createClient` now accepts the caller's config, and doctor passes its normalized copy through, so probes and SDK calls hit the identical base URL. Side benefit: `loadConfig()` no longer runs twice per invocation (double `E2A_BASE_URL` warning gone). Pinned by a test asserting the normalized URL reaches `createClient`.
3. **Silent webhook truncation** — now a `too_many_webhooks` warn, symmetric with domains.
4. **Crash isolation** — per-domain and per-webhook iterations are wrapped; an unexpected throw records a `doctor.internal` fail (transient) and the run continues. Test feeds a malformed domain object and asserts the full report (including later phases) still lands.

**Test gaps** — all taken: per-domain isolation test (domain A pass / domain B fail in one run), the fake client's resources now sit behind a Proxy that throws on any non-read method access (every test is now a read-only proof, not just the dedicated one), an API-key leak pin on the serialized report, plus the missing fail branches (delivery API error, agent connection error, `sending_not_provisioned`, DKIM fail, 429, MCP 5xx).

**Minors** — took: DMARC warn-only even on DNS errors (matching its documented contract), 408/429 transient, malformed `E2A_URL` → `invalid_deployment_url` (config), MCP 5xx transient, skip remediations now printed, root README doctor row, userinfo stripped from webhook targets, control chars stripped from human output, `dig` names quoted, ownership substring + SPF case-fold (both verified against `checkDomainRecords`).

**Declined (with reasons):** `redirect: "error"` on probes — following redirects is load-bearing for `http→https` deployment roots, and fetch only follows http(s) targets; global run deadline — per-lookup deadlines are now hard (race), and a whole-run cap would truncate large accounts' reports mid-way, which is worse than being slow for a diagnostics tool. Both can be revisited if they bite.

Golden fixtures are byte-identical — the `e2a.doctor/v1` schema is untouched by this round. 244 CLI tests pass (70 doctor), build clean, live smoke against production still exits 8 on genuine DMARC advisories.

PR #671, opened 2026-07-24
feat(telemetry): SLO measurement primitives — SLI metrics, probes, prober coverage

jiashuoz:
## Summary

Open-source measurement primitives for operating e2a against production SLOs (trust-strategy roadmap "Now" item 3, and the long-deferred "SLI exposure on e2a /metrics" fast-follow from `docs/design/prober-selftest.md`). No hosted status-page frontend, no closed-source analytics — this is the instrumentation, probes, prober coverage, and documentation the hosted service and self-hosters both consume.

### 1. Telemetry: SLI instruments + Prometheus backend (`internal/telemetry`)
- `telemetry.Metrics` grows 13 SLI instruments (HTTP, SMTP intake, outbound queue/attempts/terminals, webhook attempts, WebSocket lifecycle, async inbound, River queue gauges). Existing 8 methods and both backends unchanged in behavior.
- New `Prom` backend (first direct `prometheus/client_golang` dep): private registry, `promhttp` exposition, and **label-cardinality enforcement in the backend** — enum allowlists collapse unknown values to `"other"`; hard series caps (256 routes / 64 event types) bound the two code-defined open label sets. Labels can never carry message content, addresses, URLs, or credentials; unit tests pin both the emission and the boundaries.
- `Log` backend: moderate-rate outcome events log as before; per-request instruments are deliberately Prometheus-only (no log flooding).

### 2. Config + probes
- New `metrics:` block (`enabled`, `listen_addr`; env `E2A_METRICS_ENABLED` / `E2A_METRICS_LISTEN_ADDR`). **Off by default — zero behavior change.** When on, `/metrics` binds a separate loopback-default listener, never the public API handler.
- `/readyz` is now **drain-aware**: 503 `"draining"` from the moment SIGTERM lands (ahead of every dependency check) while `/api/health` liveness stays green — a terminating instance leaves LB rotation before in-flight work drains. Liveness stays shallow (unchanged, per the never-deepen rule).

### 3. Instrumentation (all consumers use narrow package-local interfaces, nil-safe, janitor pattern)
- **HTTP**: chi-root middleware — method / chi **route pattern** (never raw paths; legacy fallthrough = `/legacy`) / status class + latency; WS-upgrade-safe writer (Hijacker/Flusher/Unwrap passthrough).
- **SMTP edge** (`internal/relay`): one outcome per transaction — `accepted`, `accepted_dedup`, `tempfail` (451), and the three policy rejects (550/550/552), with DATA duration.
- **Outbound** (`internal/outboundsend`): queue wait (River `attempted_at − created_at`), timed submission attempts, terminal outcomes (`sent` / `failed_provider` / `failed_suppressed` / `failed_local_retries`), exactly once per message — a deferred final attempt is counted by the terminal reconciler when it settles (as `sent` on provider evidence, else a failure).
- **Queues** (`internal/jobs`): 30s maintenance periodic sampling `river_job` → `e2a_queue_depth{queue,state}` (zero-filled) + `e2a_queue_oldest_age_seconds{queue}`.
- **Webhooks** (`internal/webhookdelivery`): per-attempt outcome + endpoint status class (`none` = no HTTP response, incl. SSRF-blocked) + duration.
- **WebSocket** (`internal/ws`): active gauge, connects, disconnects by reason, drain counts, send failures — with a documented hub/handler ownership split so nothing double-counts.
- **Async inbound** (`internal/inboundprocess`): worker outcomes + processing duration.

### 4. Prober (`cmd/e2a-prober` + `internal/selftest`)
- New **`websocket_round_trip`** scenario: Bearer WS handshake → nonce'd loopback self-send → await the `email.received` push frame. SmokeSafe (no egress, no HITL, no metering) — closes the "WS not probed at all" gap.
- `/metrics` now emits **`e2a_selftest_scenario_duration_seconds{scenario}`** — per-critical-path latency, not just pass/fail.
- **`E2A_PROBE_REQUIRE_MCP=true`** turns the silent MCP skip-as-pass into a failure, so a misconfigured prod prober can't stay green while never probing MCP (the 99.9% MCP target is measured by this scenario, independently of the MCP process).
- `/status` + `consecutive_green` bake-gate contract untouched.

### 5. Docs
- New **`docs/observability.md`**: enabling metrics, the cardinality contract, probe depth split, full metric catalog, prober signals, PromQL SLI definitions, and initial 30-day SLO targets. Ends with the explicit **product-SLO vs inbox-placement** boundary: submission, provider feedback, and observed placement are separate signals; **no inbox-placement SLO is stated** (per strategy §9/§10).
- `docs/deployment.md`: operator-facing observability section.

## Migrations / API compatibility

- **No DB migrations** (the queue sampler reads `river_job` only).
- **No `/v1` API changes** — `make spec-check` passes with zero drift; no SDK/MCP/CLI/dashboard parity work needed. `/readyz` and `/metrics` are operational endpoints outside the `/v1` contract.
- Config change is additive and default-off.

## Tests

All on a private database (`-p 1`), per repo policy:

```
go test ./internal/telemetry/                                     ok  (emission + cardinality boundaries)
go test -short -p 1 <make test-unit package set>                  ok  (9/9 packages)
go test -tags integration -p 1 ./internal/{telemetry,apiserver,selftest,relay,outboundsend,jobs,webhookdelivery,inboundprocess,ws}/ ./cmd/e2a/ ./cmd/e2a-prober/   ok  (11/11)
go test -race -p 1 ./internal/sendramp/ ./internal/outboundsend/  ok  (CI race set)
make spec-check                                                   ok  (no OpenAPI drift)
coverage vs floors: httpapi 84.5%≥73, outboundsend 82.1%≥75, inboundprocess 76.1%≥65
```

Plus a live smoke: booted the server with metrics enabled — `/api/health`, `/readyz`, route-pattern + `/legacy` HTTP series, zero-filled queue gauges on `/metrics`, clean SIGTERM drain. The full selftest battery (including the new WS scenario) passes against a real in-process server.

## Notes for review

- `make test-unit` run verbatim can flake on the **shared** `e2a_test` DB (pre-existing truncation contention, documented in AGENTS.md); the identical package set is green serialized on a private DB.
- The Prometheus dep was already in the module graph (indirect); this makes it direct at v1.24.0.
- Ops follow-ups (out of scope here, e2a-ops): scrape `127.0.0.1:9091/metrics`, set `E2A_PROBE_REQUIRE_MCP=true` on prod/staging probers, alert per `docs/observability.md`.


jiashuoz:
Adversarial self-review (multi-agent) ran after the PR opened and found five substantive SLI-semantics issues, all fixed in f67dce1d:

1. **Queue wait** was `attempted_at − created_at` — cumulative message age, so every retry backoff / outage snooze / ramp deferral recorded multi-hour samples against a p95<30s SLO. Now `attempted_at − scheduled_at` (due→pickup, per attempt).
2. **Terminal reconciler mislabeled evidence-settled sends as failures**: `MarkFailed` is a guarded write that settles the row as *sent* when provider-accept evidence exists — exactly the population the reconciler prioritizes — but the metric emit assumed failure. `MarkFailed` now returns the settled outcome (`sent`/`failed`/no-op) and both emit sites report truth.
3. **`deferred_terminal` double-counted** (deferral + reconciler settle = two increments per message). Removed; the reconciler's settle is the single count.
4. **WS connections pinned the HTTP p99**: the WS handler blocks for the connection lifetime, so every disconnect dropped an hours-long observation into the /v1 duration histogram. Hijacked connections are now excluded from the histogram (still counted in `requests_total` as 1xx).
5. **Prober WS scenario grew unbounded unread backlog** (~2,880 messages/day) and every connect re-drained the oldest 100, making `e2a_ws_drained_messages_total` pure prober noise. The scenario now trashes both probe-message copies after each run.

Plus minor hardening: `agent_identities` added to the janitor table allowlist, queue-stats sampling errors log instead of feeding River retries, explicit empty `metrics.listen_addr` defaults to loopback instead of binding `:80`, and docs corrected to match (SMTP per-RCPT vs per-transaction units, queue-wait formula, exactly-once terminal contract).

Full touched-package suite re-verified after the fixes: 13 packages + `internal/agent` green with `-tags integration -p 1` on a private DB; `-race` clean on `outboundsend`.

jiashuoz:
Second review round (kimi) addressed in 036e1ee7. Disposition of every finding:

**Majors — all fixed**
1. *Panics invisible to the availability SLI* — confirmed (no recover middleware exists in the chain). The sample now emits from a defer: panic → recorded as 5xx → re-panicked, so upstream teardown is unchanged. Test pins both the 5xx sample and the re-panic. The availability PromQL note documents that crash loops now count.
2. *Prober WS scenario leaked the sent copy* — confirmed: the send response field is `message_id` (SendResultView has no `id`), so the sent-copy trash silently no-oped. Fixed; the stub now returns the real response shape and the happy-path test asserts BOTH copies (`msg_inbound_copy` from the push frame, `msg_sent_copy` from the send response) get trashed — a field-name regression now fails the test.
3. *Unmeasurable SLO targets* — the three rows (outbound 5-min terminal, webhook event→first-attempt p95, WS handshake success) are now footnoted **"target adopted; instrument pending"** with exactly which instrument each needs, and must not be reported as measured until it lands.

**Minors — fixed**: stale `created_at` wording (Prom help text + outboundsend docstring → `scheduled_at`); webhook `webhook_deleted`/`skipped_disabled` no longer `Observe(0)` (negative = count-only, same contract as the HTTP hijack path; tests updated) and the attempt-health query note explains why `skipped_disabled` is outside both numerator and denominator; relay interface docs now state per-RCPT vs per-DATA units; availability PromQL excludes `route="/legacy"` with rationale; non-loopback `E2A_METRICS_LISTEN_ADDR` binds log an explicit UNAUTHENTICATED warning; metrics listener now shuts down **after** the API server (drain window observable, error logged); the "zero behavior change" claim is reworded in docs (queue sampler + new log lines exist when disabled); "first-attempt health" renamed to per-attempt with the no-attempt-label caveat; **the 20-file whitespace-only gofmt sweep is reverted** (my `gofmt -w internal/` was too broad — apologies; main has pre-existing gofmt drift and no CI gate, worth a separate chore PR).

**Accepted as-is (with rationale)**
- *MarkSent no-op emit*: unreachable double-count in practice — `ClaimSend` refuses terminal rows, so a re-drive after a crash-between-MarkSent-and-completion exits via the `j == nil` path, which does not emit. Making MarkSent status-aware would ripple another signature change for a path with no observed trigger.
- *WS disconnect-reason inference races*: known narrow window (supersede-then-disconnect); counters are not conserved by design and the active gauge — the alerting signal — is lock-consistent.
- *Cancellation rows settled by the reconciler labeling `failed_local_retries`*: only reachable when the inline suppressed-terminal write failed all retries AND the preserved provenance lost its blocked-recipient list; the preserved-provenance path maps to `failed_suppressed` normally.
- *Vocabulary triplet* (`temporary_failure`/`retryable_failure`/`retryable`): each matches its domain's existing state vocabulary (submission attempts / webhook delivery rows / River outcomes) — unifying them would detach metric labels from the code they describe.
- PR body's stale `deferred_terminal` mention is edited out.

Full touched-package suite re-verified green after the fixes (`-tags integration -p 1`, private DB).

jiashuoz:
Correction on CI status: the round-2 run's **Go coverage gate** job was canceled at 12m16s by its own `timeout-minutes: 12` — after every package passed and every coverage floor cleared (sendramp 70.6≥70, webhookpub 69≥60, outboundsend 82.1≥75, httpapi 84.5≥73, inboundprocess 76.1≥65, webhook 83.3≥60), but before the enforcement step could report. Not a test failure or floor breach: this PR's added DB-serialized suites push `make cover` to ~12m, exactly at the old cap. d741004b raises the job timeout to 18m (~50% headroom over observed); no other job is near its cap (e2e: 3m48s of 12m).

jiashuoz:
Landed the one round-3 item worth code (0ac0d98c): **`failed_cancelled` terminal label**. `terminalOutcome` is now reason-aware — a reconciler-settled cancelled job (ReasonSubmissionCancelled, no preserved suppression list) no longer lands in `failed_local_retries`, so cancellation volume can't mask a retries-exhausted regression in the alert signal. Suppression still takes precedence when blocked recipients are preserved; the documented success-SLI denominator excludes `failed_cancelled` alongside `failed_suppressed`. The integration test's cancelled-state candidate now pins the mapping (three `failed_local_retries` + one `failed_cancelled`).

Also added the guard comment for the MarkSent race: the emit's exactly-once property currently depends on the SNS-feedback path (`FinalizeProviderAcceptedTx`) staying uninstrumented — anyone instrumenting it later is pointed at making the emit status-aware first.

Remaining open items (3–6 + nits) stay as documented/accepted semantics per the earlier disposition.

PR #675, opened 2026-07-24
test(harness): per-package test databases; parallelize DB-backed runs

jiashuoz:
## Summary

Follow-up to #671's review discussion: DB-backed test runs no longer serialize.

The `testutil` harness truncates tables between tests, so one shared database forced `-p 1` on every DB-backed run — the documented cross-package flake source (AGENTS.md), and a ~12-minute wall for the coverage job that recently started brushing its CI timeout.

**Mechanism** (`internal/testutil/db.go`):
- `TestDBURL()` derives a **per-package database** inside test binaries: `<base>_pkg_<package>`, from the test binary's name (`relay.test` → `…_pkg_relay`; base names are unique across the repo). Every URL consumer in one test binary — `TestDB`, hand-built pools, the in-process contract server — lands on the same database, so intra-package behavior is unchanged.
- `OpenPreparedTestDB` **self-provisions** a missing database: SQLSTATE 3D000 → `CREATE DATABASE` via the base URL's server (42P04 duplicate-race tolerated) → retry, then the existing migrate+truncate path. First run per package pays one migration pass; after that it's a no-op check.
- Non-test binaries (`cmd/e2a-contract-server`) and `E2A_TEST_DB_SHARED=1` get the base URL verbatim. Concurrent sessions running the *same* package still follow the per-session base-URL guidance (updated in AGENTS.md, including that `*_pkg_*` databases are disposable).

**Parallelism**: Make targets (`test`, `test-integration`, `test-e2e`, `cover`) and the CI coverage/e2e jobs move from `-p 1` to `-p 4` — capped below core count so N × pgxpool connections stay under Postgres's default `max_connections=100`. The coverage job's race step goes to `-p 2`.

## Impact

- Local full cover run: ~12m serialized → **4m14s** at `-p 4` (44 packages).
- Full `-tags integration` suite: 49 packages, 0 failures, 6m33s uncached.
- Coverage floors verified on the parallel profile (`go-test-coverage` PASS, total 77.7%).
- `make test-unit`'s intermittent shared-DB failures (relay/webhook packages truncating each other) are structurally fixed — it runs default-parallel and now each package is isolated.

No API changes, no migrations, no production code paths touched (test harness + Make/CI/docs only).

## Tests

```
go test ./internal/testutil/                                  new: suffix derivation, shared-mode escape, self-provision + idempotent reopen
go test -p 4 -covermode=atomic ./internal/...                 44 ok, floors PASS   (4m14s)
make test-e2e (parallel)                                      5 ok, 0 FAIL         (1m42s)
go test -tags integration -p 4 -count=1 ./...                 49 ok, 0 FAIL        (6m33s)
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -60,10 +60,8 @@ jobs:
   go-coverage:
     name: Go coverage gate
     runs-on: ubuntu-latest
-    # 18, not 12: `make cover` runs every DB-backed package serialized (-p 1)
-    # and sits at ~12m after the SLO-instrumentation suites landed — the old
-    # cap canceled a fully-passing run 16s over budget, after the last
-    # package but before floor enforcement. Keep ~50% headroom over observed.
+    # Headroom over observed: the parallel (-p 4, per-package DBs) cover run
+    # is well under this; the cap exists so a hung suite can't burn an hour.
     timeout-minutes: 18
     services:
       postgres:
@@ -85,15 +83,16 @@ jobs:
         with:
           go-version: "1.26"
           cache: true
-      # Generate the coverage profile with the test DB available (the webhook
-      # delivery/publish + httpapi suites are DB-backed; -p 1 avoids cross-
-      # package test-DB contention). Floors live in .testcoverage.yml.
+      # Generate the coverage profile with the test DB available. Packages
+      # run in parallel (-p 4): the testutil harness gives each package its
+      # own database, self-provisioned from the service's e2a_test base.
+      # Floors live in .testcoverage.yml.
       - name: Generate coverage profile
         run: make cover
       - name: Race-check ramp and outbound worker
         env:
           E2A_TEST_DATABASE_URL: postgres://e2a:e2a@localhost:5433/e2a_test?sslmode=disable
-        run: go test -race -p 1 ./internal/sendramp ./internal/outboundsend
+        run: go test -race -p 2 ./internal/sendramp ./internal/outboundsend
       - name: Enforce per-package coverage floors
         uses: vladopajic/go-test-coverage@v2
         with:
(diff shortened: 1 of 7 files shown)


jiashuoz:
Adversarial review (multi-agent, all findings empirically reproduced against PG16) caught three real issues that green CI could not — all fixed in 56153720:

1. **BLOCKER**: `contract_server_river_test`'s reachability gate raw-pinged the derived per-package URL with no self-provisioning — on every cold database (which is every CI run) it hit 3D000 and **silently skipped the River-state/migration-ledger regression gate forever**. It now goes through `OpenPreparedTestDB` (self-provisions) and Fatals on preparation errors. Proven by dropping the package DB and watching the test run cold: PASS, not SKIP.
2. **Skip-vs-fail classification**: provisioning failures on a *reachable* server (e.g. a role without CREATEDB — the exact pre-created-DB setup AGENTS.md used to prescribe) were classified as "DB unavailable" → the entire DB tier skipped green. Now a preparation error → Fatal; only base-server-unreachable skips.
3. **Wrong race SQLSTATE**: a true concurrent CREATE DATABASE race loses with 23505 (unique violation on pg_database_datname_index — reproduced with 8 parallel creates), not the 42P04 the code tolerated. Both tolerated now.

Minors also fixed: derivation is idempotent (a child `.test` re-exec handed an already-derived URL can't double-suffix and mint junk DBs) and gated to postgres:// URLs (DSN keyword/value form passes through verbatim); `E2A_TEST_DB_SHARED` accepts 1/true/yes; CONTRIBUTING.md's stale `-p 1` line; AGENTS.md notes `_pkg_` siblings survive a base-DB reset.

Review confirmed clean: consumer split-brain (all TestDBURL consumers stay on one DB per test binary; ts/python contract jobs use the non-.test server binary → base URL verbatim), URL derivation with query params/encoded passwords, package name uniqueness across all 27 test dirs, cold `-p 4` self-provisioning, connection budget (observed peak 17 vs max_connections=100), and no other workflow/script assumes the single shared DB.

PR #680, opened 2026-07-24
feat(auth): signup hook for new-user provisioning notifications

jiashuoz:
## What

Adds an optional **signup hook**: when a **brand-new** user account is created via Google OAuth signup, the server fires an HMAC-signed HTTP POST to a configured URL. This lets an external service — e.g. a hosted deployment's billing sidecar — run first-touch provisioning such as sending a welcome email. Self-host default is off (empty URL = disabled, zero behavior change).

## Why

The hosted deployment wants to welcome-email new signups. The server is the only component that knows *at the moment of signup* whether the OAuth upsert created a new user or matched a returning one, so it exposes that moment as an internal webhook, mirroring the existing account-deletion billing hook.

## Design notes

- **Mirrors the deletion billing hook exactly** (`notifyBillingUserDeleted` in `internal/agent/user_data_rights_api.go`): JSON body, `X-E2A-Internal-Signature: hex(HMAC-SHA256(internal_api_secret, body))`, 5s timeout, expects `204 No Content`, logs and continues on any failure.
- **New-user discriminator:** `CreateOrGetUser`'s `INSERT … ON CONFLICT DO UPDATE` upsert gains a `(xmax = 0) AS inserted` RETURNING column — Postgres's system column `xmax` is 0 only on a freshly inserted row. Exposed as `CreateOrGetUserWithCreated` returning `(user, created, err)`; `CreateOrGetUser` stays a thin wrapper so its many existing callers (prober seeds, test harness, dozens of tests) are untouched.
- **Trigger lives in `HandleCallback`, not the store upsert** — deliberately, so synthetic accounts (prober seeds, `-bootstrap-email`, test harnesses) never fire the hook. OIDC login never provisions users, so it's out of scope by construction.
- **Best-effort, never blocks login:** fired from a goroutine with its own timeout-bounded `context.Background()` (the request context is canceled the instant the login response is written). Any failure is logged and dropped.
- **Config:** `limits.signup_hook_url` (yaml) / `E2A_SIGNUP_HOOK_URL` (env), sitting next to `billing_hook_url` and signed with the same `limits.internal_api_secret`. Like `billing_hook_url`, no extra validation — empty disables.

Payload:

```json
{"user_id": "…", "email": "…", "name": "…"}
```

## Test coverage

- `internal/auth/signup_hook_test.go` (drives `HandleCallback` through the existing fake-Google harness, real test DB):
  - new user → hook receives `{user_id, email, name}` with a matching independently-recomputed HMAC; login still 302s to `/dashboard`
  - returning user (same `google_subject`) → hook stays silent
  - empty hook URL (self-host default) → no outbound call
  - hook endpoint down (connection refused) → login still succeeds and the user row exists
- `internal/identity/user_upsert_test.go`: xmax discriminator regression — `created=true` on first insert, `false` on the DO UPDATE path (with profile-field refresh and stable user ID), `true` again for a distinct subject; wrapper contract preserved.
- Full runs of the changed packages (`internal/identity`, `internal/auth`, `internal/config`, `internal/agent`) green against an isolated test DB; `make test-unit` green; `go build ./...` + `go vet` on changed packages clean.

Not a client API change — server-internal config + internal webhook only; SDKs/CLI/MCP untouched.

### DIFF
--- a/cmd/e2a/main.go
+++ b/cmd/e2a/main.go
@@ -483,6 +483,10 @@ func main() {
 
 	// User auth (Google OAuth for agent developers)
 	userAuth := auth.NewUserAuth(&cfg.OAuth, store, cfg.IsProduction())
+	// Optional new-user signup hook (hosted welcome email etc.), signed
+	// with the same internal secret the billing hook uses. Empty URL =
+	// disabled, the self-host default.
+	userAuth.SetSignupHook(cfg.Limits.SignupHookURL, cfg.Limits.InternalAPISecret)
 	// Generic OIDC Authorization Code login. Disabled configurations perform
 	// no discovery and leave both OIDC routes unregistered. Enabled
 	// configurations construct synchronously (no network call) and discover
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -185,6 +185,15 @@ limits:
   # docker compose deployment alongside the hosted billing sidecar:
   # billing_hook_url: "http://billing:9000/api/internal/billing/cancel"
   billing_hook_url: ""
+  # URL the OSS server POSTs (HMAC-signed with the same
+  # internal_api_secret) when a brand-new user account is created via
+  # OAuth signup, so an external service can run first-touch
+  # provisioning such as sending a welcome email. Returning logins
+  # never fire it. Empty disables the call (self-host without a
+  # provisioning service). Override with E2A_SIGNUP_HOOK_URL. Example
+  # for a docker compose deployment alongside a hosted sidecar:
+  # signup_hook_url: "http://billing:9000/api/internal/billing/signup"
+  signup_hook_url: ""
 
 # Per-user budget shared by authenticated message, conversation, and webhook
 # reads across every agent client and dashboard session on the account. The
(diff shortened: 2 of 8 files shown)


jiashuoz:
Closing — superseded by the tokencanopy consolidation. Auth is moving to WorkOS, so the Google-OAuth callback in `HandleCallback` will no longer be the signup trigger; the equivalent signal will come from WorkOS user webhooks (`user.created`, `WorkOS-Signature`, at-least-once with exponential-backoff retries) or the events API.

The design itself is sound and worth keeping on this branch for reference:
- The `xmax = 0` insert-vs-update discriminator was verified empirically on PG 16 (insert → true, DO UPDATE → false).
- The hex-encoded HMAC-SHA256 wire format matches the existing billing hook, and the consumer side (tokencanopy/e2a-ops#299, closing in lockstep) reuses everything except the trigger wiring.
- Bonus when we redo it: WorkOS webhooks are at-least-once, which fixes the accepted "welcome lost on outage" tradeoff of the best-effort goroutine — the existing send-side idempotency key already handles duplicates.

PR #682, opened 2026-07-25
fix(webhookdelivery): retry transient webhook-lookup errors, never mislabel them deleted

jiashuoz:
## Summary

Fixes a pre-existing delivery-loss bug in `webhookdelivery.DeliverWorker.Work` (surfaced during PR #676's review): **any** error from `GetWebhookByIDInternal` — including a transient Postgres blip, pool exhaustion, or network reset — was treated as "webhook deleted", terminally failing the delivery row (`markFailedReliably` + `webhook_deleted` metric + `river.JobCancel`). A one-second DB blip could permanently kill deliverable webhook deliveries, mislabeled "webhook not found" and hidden behind the customer-fault `webhook_deleted` metric (excluded from the attempt-health SLI denominator, so even a batch loss during a failover wouldn't burn any SLO).

The fix branches on the store's sentinel: `GetWebhookByIDInternal` returns `identity.ErrWebhookNotFound` only for a genuine miss (`pgx.ErrNoRows`) and propagates every other error raw.

- **Genuine miss (`errors.Is(err, identity.ErrWebhookNotFound)`)** → terminal path unchanged (terminal write + `webhook_deleted` + cancel).
- **Anything else** → returned as a retryable worker error; the row stays pending and River retries. No metric — nothing was attempted and no deletion occurred.
- **Transient error on the final attempt** → mirrors the POST-failure final-attempt path: the terminal `failed` write is the row's last chance (a discarded job would otherwise strand the row `pending` with a dead `job_id` the reconciler can't see), counted as `exhausted`/`none` — never `webhook_deleted` — so a sustained e2a-side outage burns the attempt-health SLI instead of hiding.

Also rides along: a timing-flake fix in `TestDeliverWorker_Metrics_FirstAttemptLatencyObservedDespiteBackoffSchedule` (merged in #676; the assertion now checks only a lower bound — the seeded age compares Postgres `now()` against host `time.Now()`, and Docker VM clock drift made any tight upper bound flaky).

## Review history

- **Round 1** (adversarial subagent): **SHIP**; the one medium finding (final-attempt stranding residual) is addressed here by the `exhausted` terminal treatment. Verified the sentinel contract is airtight (only `pgx.ErrNoRows` maps to it; the filters-JSON unmarshal error does not) and that no metric can fire for something that didn't happen.
- **Round 2** (external review, "approve with one requested fix"): the final-attempt terminal write stored the raw pgx error into `last_error`, which `GET /v1/webhooks/{id}/deliveries` returns **verbatim** — leaking internal hosts/IPs and DB identifiers. Fixed: `last_error` is now the constant `"internal error resolving webhook"` (the existing convention; the real error still reaches River's job-error log via the returned `err`), pinned by test. Docs nits applied: `status_class=none` now covers "no POST was ever made", and the pre-POST retry window's observability gap is documented in the attempt-health SLI section.
- **Accepted trade, surfaced by review:** a permanently-broken filters-JSON row (unmarshal error in the store) now retries 8× over the 29h21m envelope before terminating instead of failing fast. Better classified than before, slightly more wasteful — accepted.
- **Known residual, noted not fixed:** the `GetSubscriberDeliveryByID` row-load error path has the same discard-after-final-attempt stranding gap (pre-existing, separate from the misclassification this PR fixes).

## Test plan

- TDD: `TestDeliverWorker_TransientWebhookLookupErrorRetries` failed pre-fix exactly as predicted (`JobCancelError: webhook ... not found: connection reset by peer`); green after.
- New tests: transient error → non-cancel error + row pending + no metric; genuine sentinel → terminal + `webhook_deleted`; wrapped sentinel (`fmt.Errorf` %w) still terminal; transient error on final attempt → row failed + `exhausted`/`none`/no-duration + constant non-leaking `last_error`.
- Three pre-existing tests that exercised the terminal path via a generic error (i.e., via the bug itself) now use the sentinel; `worker_extra_test.go`'s unreachable-lookup case deliberately keeps a generic error.
- `make spec-check` — green (no `/v1` change).
- `E2A_TEST_DATABASE_URL=postgres://e2a:e2a@localhost:5433/e2a_test_whtrans?sslmode=disable go test -tags integration -p 1 ./...` (private DB) — green, gated on `go test`'s own exit code.
- Coverage gate (`vladopajic/go-test-coverage`, `.testcoverage.yml`) — exit 0; webhookdelivery sits at 90.9% vs the 88 floor.

### DIFF
--- a/docs/observability.md
+++ b/docs/observability.md
@@ -86,7 +86,7 @@ acceptance SLI below deliberately excludes them.
 
 | Metric | Type | Labels | Meaning |
 |---|---|---|---|
-| `e2a_webhook_attempts_total` | counter | `outcome`, `status_class` | Delivery attempts to subscriber endpoints: `delivered`, `retryable_failure`, `exhausted` (terminal after max attempts), `webhook_deleted`, `skipped_disabled`. `status_class` is the endpoint's response class, or `none` when no HTTP response was received (connect/DNS/SSRF-blocked). |
+| `e2a_webhook_attempts_total` | counter | `outcome`, `status_class` | Delivery attempts to subscriber endpoints: `delivered`, `retryable_failure`, `exhausted` (terminal after max attempts — including a delivery that exhausted its retries on a pre-POST infrastructure error, e.g. a sustained webhook-lookup outage), `webhook_deleted`, `skipped_disabled`. `status_class` is the endpoint's response class, or `none` when no HTTP response was received — connect/DNS/SSRF-blocked, or no POST was ever made. |
 | `e2a_webhook_attempt_duration_seconds` | histogram | — | HTTP POST duration per attempt. |
 | `e2a_webhook_first_attempt_latency_seconds` | histogram | — | Event→first-attempt latency per subscriber delivery (attempt start − the `webhook_events` row's `created_at`), observed **only on a first-delivery row's first HTTP attempt** (no recorded prior attempt — regardless of River attempt number, so a first POST delayed by pre-POST failures still observes). Retries, the no-POST outcomes (`webhook_deleted`, `skipped_disabled`), replay rows (their baseline would be the original event's age), eventless `/test` deliveries, and jobs that sat through a customer-disabled window (River snooze) never observe. |
 | `e2a_outbox_events_published_total` | counter | `type` | Events written to the outbox (fan-out input). |
@@ -238,6 +238,13 @@ snooze, and no POST happens:
 sum(rate(e2a_webhook_attempts_total{outcome="delivered"}[5m]))
 / sum(rate(e2a_webhook_attempts_total{outcome=~"delivered|retryable_failure|exhausted"}[5m]))
 ```
+
+One deliberate gap: a delivery retrying a *pre-POST* infrastructure error
+(e.g. a webhook-lookup DB outage) emits nothing until it exhausts — up to
+the 29h21m retry envelope — because no attempt happens to count. River's
+job-error logs and `e2a_queue_oldest_age_seconds{queue="webhook"}` cover
+the window; the first `exhausted` sample lands the SLI impact when the
+envelope runs out.
 
 **Webhook event→first attempt** — p95 latency from event creation to the
 first HTTP attempt (covers fan-out, queue wait, and worker pickup):
--- a/internal/webhookdelivery/worker.go
+++ b/internal/webhookdelivery/worker.go
@@ -175,8 +175,30 @@ func (w *DeliverWorker) Work(ctx context.Context, job *river.Job[WebhookDeliverA
 	}
 
 	wh, err := w.webhooks.GetWebhookByIDInternal(ctx, d.WebhookID)
+	if err != nil && !errors.Is(err, identity.ErrWebhookNotFound) {
+		// Anything but a genuine miss is an infrastructure error (Postgres
+		// blip, pool exhaustion, network reset) — retryable. Terminally
+		// cancelling here would permanently fail a deliverable row, mislabel
+		// it "webhook not found", and hide the loss behind the
+		// customer-fault webhook_deleted metric.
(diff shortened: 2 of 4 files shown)


jiashuoz (commented):
**Verdict: approve with one requested fix.** The core change is correct and well-targeted — it fixes a genuine delivery-loss bug, the sentinel contract it depends on is airtight, and the final-attempt handling correctly mirrors the existing POST-failure path. One new information-disclosure issue rides along.

## Verification performed

Worktree at `315c50b8`, private test DB:

- `go vet ./internal/webhookdelivery/...` clean.
- Full `./internal/webhookdelivery/...` integration suite green; **coverage 90.9%** vs the 88 floor in `.testcoverage.yml:42`.
- All 5 new/changed tests pass individually; the deflaked timing test green 8×.
- Adjacent `./internal/webhook/...`, `./internal/webhookpub/...`, `./internal/jobs/...` green.
- **Sentinel contract confirmed**: `internal/identity/webhooks.go:250` maps *only* `pgx.ErrNoRows` to `ErrWebhookNotFound`; the filters-JSON unmarshal error (`:255`) propagates raw. The PR's claim holds.
- **Blast radius confirmed**: the only production `WebhookReader` is `*identity.Store` (`webhookdelivery/jobs.go:56`, `testutil/server.go:206`), and there is a single enqueue site with `MaxAttempts: MaxDeliveryAttempts` (`jobs.go:145`), so `job.Attempt >= MaxDeliveryAttempts` genuinely is the final attempt.

## Requested fix

**[Medium] The new terminal write leaks internal infrastructure detail into a customer-facing API field.**

`internal/webhookdelivery/worker.go:190` writes `"webhook lookup error: "+err.Error()` through `markFailedReliably` → `webhook_subscriber_deliveries.last_error`, which is returned **verbatim** to the customer as `last_error` in `GET /v1/webhooks/{id}/deliveries` (`internal/httpapi/webhooks.go:378` and `:445`). A pgx connect failure formats as:

```
failed to connect to `host=10.0.1.5 user=e2a database=e2a`: dial tcp 10.0.1.5:5432: connect: connection refused
```

so internal hostnames/IPs, the DB user, and the database name reach the customer's delivery log. Note the sentinel branch immediately below writes the constant `"webhook not found"` — the constant-message convention is already established here, and the POST path's `out.Error` is customer-relevant (their endpoint), not internal.

Fix is one line: write a constant (`"internal error resolving webhook"`) and log the real `err` server-side alongside the existing CRITICAL log.

## Nits (non-blocking)

**Observability gap during the outage window.** Attempts 1–7 emit no metric at all, so a sustained lookup outage is invisible in `e2a_webhook_attempts_total` for up to 29h21m until the first `exhausted` lands. This is the deliberate choice ("nothing was attempted"), and River's job-error metrics cover the window — but worth a sentence in `docs/observability.md`.

**`docs/observability.md:89` is now slightly stale.** `status_class=none` is documented as "no HTTP response was received (connect/DNS/SSRF-blocked)"; it can now also mean "no POST was ever made" on an `exhausted` attempt — which the SLI denominator at `:239` counts.

**The timing deflake's stated rationale is likely wrong, and its lower bound is dead.** The band compares Postgres `now() - 90min` against Go `time.Now()`, so it measures host↔container clock skew, not "suite-load overhead between seeding and Work" — a minute-plus gap between two adjacent statements is implausible (locally the skew is 13ms and the value lands at ~90min+ε). Separately, `got < 88*time.Minute.Seconds()` can never fire: `start` is always after the seeded `created_at`, so the latency is ≥ 90 min by construction. A skew-free version would `RETURNING created_at` from the INSERT and assert against that. The widened band does work; just don't treat the root cause as diagnosed.

**Accepted trade worth naming in the PR body:** the filters-unmarshal error (permanently broken data) now retries 8× over 29h before terminating instead of failing fast. Better classified than before, slightly more wasteful — fine, but it's a behavior change the description doesn't mention.

PR #683, opened 2026-07-25
feat(api): add per-axis capabilities to the domain resource

jiashuoz:
## Summary

A domain has two independent capabilities: it can **receive** mail (ownership TXT + inbound MX) and it can **send** as its own address (the async SES sending identity — DKIM + custom MAIL FROM). Both are already tracked separately and reconciled on different schedules, but the public contract only names them as `verified` and `sending_status` — and `verified` reads as though it covered the domain as a whole while naming only the inbound axis.

This adds `capabilities: {inbound, outbound}` to `DomainView` so the two axes have one home whose name stays accurate as they diverge. It is **derived, never stored**: it reuses the exact locals the DNS records are already stamped from, so it cannot drift from `dns_records[].status` or the `sending_status` rollup. No new storage, no new DB reads, no new queries.

`outbound` is deliberately the all-or-nothing rollup — per-record detail stays in `dns_records[]`, so a domain with healthy DKIM but broken MAIL FROM reads `outbound: failed` while the `dkim` record still reads `verified`. That's pinned by a test with a comment, to stop a future "improvement" into a per-axis derivation reporting a capability the domain doesn't have.

`inbound` emits only `verified`/`pending` — it restates a boolean, so there is deliberately no inbound failure state: a missing or wrong MX leaves it pending indefinitely. The doc string says so, since "pending forever" is otherwise an easy trap for a polling client.

Motivating context: supporting outbound-only (send-only) domains later means a domain will legitimately report inbound-pending while sending fine. Landing this now means that work populates a field that already exists instead of contradicting one whose GA-frozen doc string says `verified` requires both the ownership TXT *and* the inbound MX.

## Client surface checklist

- [x] Go handler + tests — `internal/httpapi/domains.go`; `TestDomainViewCapabilities` covers a 6-case axis matrix, anti-drift assertions against `dns_records[].status` and the rollup, and the diverged-SES-axes case
- [ ] Migration — **N/A**: no schema change. The field is derived from existing columns.
- [x] OpenAPI spec + generated types refreshed — `make generate` clean; both codegen normalizer allowlists extended for the new model
- [x] TypeScript SDK — generated base regenerated (new `DomainCapabilities` model). No `E2AClient` change needed: the domains resource already exists and the field flows through the regenerated `DomainView` type.
- [x] Python SDK — generated base regenerated (new `domain_capabilities.py`). Same reasoning for `AsyncE2AClient`.
- [ ] CLI — **skipped: redundancy.** `doctor` already diagnoses domain DNS per-record, at finer granularity than this rollup, so a capabilities check would add no signal. (An earlier revision of this description claimed the frozen `e2a.doctor/v1` schema blocked it — that was wrong; `cli/src/commands/doctor.ts:31` explicitly allows new check IDs. The decision stands, the stated reason was incorrect.)
- [x] MCP — no new tool; the existing `get_domain` description now teaches `capabilities` and points polling at `capabilities.outbound`. No registry assertion needed since the tool set is unchanged (`check-mcp-tool-compat`: additive, 60/60).
- [x] Contract tests — `tests/contract/scenarios.yaml` asserts both axes at register and pins the post-verify transition through GET. The shared scenario file drives the Go, TS, and Python runners, so cross-SDK parity is pinned in one place.
- [x] Tests at each surface — Go, contract, and web (12 new: 7 card, 5 helpers), each with negative/regression paths

**Web dashboard** — not on the template's list, so called out explicitly: it reads both axes through one seam (`inboundCapability` / `outboundCapability` / `canReceive` in `onboarding/state.ts`, with fallback to the legacy fields for servers predating `capabilities`) and surfaces the outbound axis as a header chip. Every `domain.verified` read in the domain UI — the card, the get-started checklist phase, the domain-selector chip, and the inboxes dashboard stat — now goes through that seam.

## Operational risk

Low. Additive, read-only response field on three existing domain responses; no side effects, no data touched, no auth/billing/notification path involved. The oasdiff compat gate reports **no breaking changes**, and both legacy fields keep working unchanged. Rollback is a plain revert — no migration, no flag.

One judgment call worth a reviewer's eye:

**It lands as a stable field, not `beta`.** `DomainView` carries no `x-stability-level`, so `capabilities` is inside the GA freeze. It's the right call because the field is a pure restatement of two already-stable fields, so stable costs nothing and the shape has almost no room to be wrong. New string values can still appear (both axes are documented open sets). Marking it beta instead is a one-line change in `stability.go`.

*(An earlier revision argued this on the grounds that "clients ignore beta surface." That premise is false in this repo — MCP and web both consume beta fields such as `hold_reason`. Same conclusion, better reason.)*

## Known follow-up, not fixed here

**Version-skew policy is undocumented.** `capabilities` is `required` in the spec, so the generated Python model rejects a payload without it — an SDK carrying this field requires a server that emits it. This is **pre-existing and repo-wide**, not introduced here: removing `sending_ramp` from a payload raises the identical `ValidationError`, and Huma marks every non-`omitempty` field required, so every additive response field has this property. Required-at-freeze is also the correct contract choice (you cannot add to `required` post-GA without a major version). What's missing is the written policy — "SDK ≥ server" — in both SDK READMEs and `docs/api.md`. Worth resolving before the GA tag; out of scope for this PR.

## Test plan

- [x] `make test-unit` — exit 0
- [x] `make spec-check` — clean (no drift)
- [x] `make openapi-compat-check` — "No breaking changes to report"
- [x] `make generate` — spec + both SDK bases regenerated, committed
- [x] Go contract scenarios (`-tags integration`) — pass; the new `capabilities` assertions mutation-tested to confirm they actually bite
- [x] TS SDK — `build` + 206 tests + type tests pass
- [x] Python SDK — 466 passed, 4 skipped (run with `addopts` overridden; `mypy`/`pytest-cov` are not installed locally, so **the Python type gate ran only in CI**)
- [x] MCP — `build` + 278 tests + type tests pass
- [x] Web — 604 tests pass (up from 592), `lint` 0 errors, `build` static export succeeds
- [x] `validate-plugin`, `check-repository-text-integrity`, `check-mcp-tool-compat`, `sync-agent-docs --check` pass
- [ ] Post-merge: confirm `capabilities` appears on `GET /v1/domains/{domain}` against a deployed instance

### Pre-existing, not from this PR

`npx tsc --noEmit` in `web/` reports two errors about `DashboardAgent.id` in `AgentHeader.test.tsx` and `onboarding/state.test.ts`. Verified identical on baseline with this branch's changes stashed. They don't gate CI (the web gates are `lint`/`test`/`build`, and `next build` doesn't type-check those files), but bare `tsc` being red is worth its own cleanup.


jiashuoz:
# Launch review — GA-freeze lens (API contract / SDK parity / security / docs)

**Verdict: GO.** No blockers from any dimension. The contract is freeze-safe: open-set strings (no closed enums) in both SDKs, a single derived constructor that can't drift from `dns_records[].status` or the `sending_status` rollup, additive-only per the oasdiff gate, and a clean security pass. Two items should be resolved **before the GA tag** (not necessarily before merge), plus a handful of cheap fixes worth folding in now.

## Cross-corroborated finding (flagged independently by the API-contract and SDK-parity passes — highest confidence)

**Python SDK hard-fails against a server that predates `capabilities`** — MAJOR.
`capabilities` is in the spec's `required` list, and the generated Python model makes it a non-Optional pydantic field. Verified by execution: `DomainView.from_dict()` on an old-server payload raises `ValidationError` (`sdks/python/src/e2a/v1/generated/models/domain_view.py:34,121`), surfaced as a typed `E2AServerError invalid_response` through the ergonomic client (`client.py:180-195`). TS tolerates the missing key but leaves `capabilities === undefined` under a non-optional type (`DomainView.ts:20`) — so in the same version-skew window Python fails loudly at the call and TS fails silently later in user code.

This is the **pre-existing house pattern** (`sending_ramp` behaves identically), the server always emits the field from a single constructor, and required-at-freeze is the *right* contract choice (you can't add to `required` later without a major version; dropping it post-GA would itself be oasdiff-breaking). What's missing is the policy:

- **Fix:** document the version coupling (SDK ≥ server for `domains.*`) in both SDK READMEs/changelogs and `docs/api.md`. Decide this once, now — Huma marks every non-`omitempty` field required, so **every future additive response field post-GA re-triggers this exact break** in the Python SDK.

## Must-fix before the GA tag

1. **Version-skew policy** (above) — write it down; it's currently an undocumented freeze.

2. **Decide the `capabilities.inbound` vocabulary now** — MAJOR.
`inbound` restates the one-way `verified` bool, so it can only ever emit `verified`/`pending` (`internal/httpapi/domains.go:144-147`) — yet its frozen doc string (`domains.go:44`) documents `none` (semantically impossible for inbound — "not provisioned" only makes sense for the async SES identity) and `failed` (no inbound-failure or TTL mechanism exists anywhere in `internal/identity`, `internal/janitor`, or migrations). Not v2-forcing — it's an open set, so emitting `failed` later is additive and legal — which is exactly why it's not a blocker. But GA is the only cheap moment: a client polling `inbound == "pending"` polls forever on a domain whose MX is simply wrong, where the outbound axis would read `failed` with a `sending_error` reason. Cheapest honest fix: trim the doc to the two real values. Better fix: define and emit `failed` for stalled inbound verification so the axes mirror.

## Strongly recommended (cheap now, annoying later)

- **Web seam bypass** — the PR's own invariant in `web/src/app/components/onboarding/state.ts:14-19` ("Read every axis through these — never off `verified` directly") is violated by `web/src/app/(app)/get-started/_components/CustomDomainChecklist.tsx:25`, `DomainSelector.tsx:76,81`, and `web/src/app/(app)/inboxes/page.tsx:145`. Display-only and can't misreport today (new server: `inbound` restates `verified` by construction; old server: `verified` is the only signal), but these are exactly the reads that silently go wrong the day outbound-only domains land — the PR's own motivating scenario. One-line fixes onto `canReceive(domain)`.
- **No contract-test pin for the new field** — `tests/contract/scenarios.yaml` has no `capabilities` assertion and neither SDK suite mentions it (grep-verified). A one-line assertion per SDK pins TS↔Python parity against a real server forever; without it a silent regen regression ships undetected.
- **"Outbound pending" chip shown for `none`** — `web/src/app/(app)/domains/_components/DomainCard.tsx:184` collapses "not provisioned" (nothing in flight; the blocker is the user's own unverified inbound DNS) into "pending" (reads as "just wait"). The pre-existing `SendingStatusChip` makes the same collapse, so it's consistent-not-regressed — optional, but the new header chip was the cheap place to distinguish.
- **Plugin skill gap** — `plugins/e2a/skills/e2a/SKILL.md:142-146` still teaches "`verify_domain` returns `verified: true` ⇒ live ⇒ create agents" with no outbound-axis caveat. Pre-existing (it never mentioned `sending_status` either), but this PR created "the field to prefer" and updated the sibling MCP surface while leaving the skill that teaches the same journey untouched. Add one line: after inbound verifies, poll `get_domain` until `capabilities.outbound` is `verified` before expecting own-domain From.
- **Small stuff:**
  - `DomainCapabilities` schema has no object-level description in the spec (`api/openapi.yaml:1065`) — Huma doesn't lift the type doc; generated SDK reference docs will show a bare object.
  - The PR body's rationale for skipping CLI is slightly off: the frozen `e2a.doctor/v1` schema doesn't block adding a capabilities check (`cli/src/commands/doctor.ts:31` — "new IDs may be added; never renamed"). The real justification is redundancy — `doctor` already diagnoses per-record at finer granularity than the rollup. Fix the description, not the decision.
  - TS drops unknown subfields of `capabilities` at deserialize time while Python preserves them in `additional_properties` — pre-existing generator-wide asymmetry, not this PR's code, but `capabilities` (declared `additionalProperties: true`) is exactly the kind of object it bites. Worth recording in the parity ledger.
  - Observation only: the web types `DomainCapabilityStatus` as a closed Literal union over a documented open set (`web/src/app/components/onboarding/types.ts:33`). Runtime-safe (both chips fall through to generic renders; `canReceive` treats unknown as not-verified), and web is a private app pinned to its own server — but it's the opposite typing posture from both SDKs.

## Verified good (checked and passed)

- **Security: clean pass.** Tenant scoping verified on every emitter — list/get/register all resolve the domain against the authenticated user (`WHERE user_id` in `internal/identity/store.go:978,1187`); `domainView` has exactly one production caller and is only reachable with user-scoped rows, so `capabilities` cannot leak cross-tenant. No spoofing path: the derivation touches no request input, and structurally cannot drift from `dns_records[].status` (same locals, same row snapshot, same function). Web fallback is fail-closed (`canReceive` requires `=== "verified"`) with no XSS or prototype-pollution surface (plain optional-chaining reads, JSX text interpolation only, `encodeURIComponent` on the one `href`). `verification_token` on list/get is pre-existing and benign by design — it's a value meant for public DNS.
- **Producer↔consumer accuracy** — every `docs/api.md` and MCP-description claim traced to producer code and held, including the non-obvious "DKIM verified + MAIL FROM broken ⇒ `outbound: failed` while the `dkim` record reads `verified`" (`mapSESStatus` is all-or-nothing, `internal/senderidentity/ses.go:162-178`; per-record dkim follows its own axis; pinned by the diverged-axes test in `domainview_test.go:236-250`). The MCP "equivalently `sending_status`" hedge is byte-exact (same local). The docs accurately scope `capabilities` to list/fetch/register — `VerifyDomainView` omits it and the docs don't overclaim.
- **SDK parity** — 13/13 `DomainView` fields identical TS vs Python (required/optional and types match); open-set `string`/`StrictStr` on both sides with unknown values tolerated at runtime (verified by executing both deserializers); regeneration complete across `FILES`, `all.ts`, `ObjectSerializer`, and both `__init__.py`s; the "no hand-written layer change needed" claim confirmed — both ergonomic resources return the generated `DomainView` directly and no hand-written domain model duplicates it.
- **Web legacy fallback** — old-server JSON → helper → UI traced end-to-end (`state.ts:25-32`); cannot misreport in either direction, because the new server's `inbound` restates `verified` by construction.
- **Gates re-run on the branch** — `TestSpecGoldenNoDrift`, `TestDomainViewCapabilities`, and the open-enum/vocabulary suites all pass; PR-body spot-checks confirmed (12 new web tests, additive-only oasdiff result, `check-mcp-tool-compat` unaffected).

## On the two judgment calls flagged in the PR body

1. **Stable, not `beta` — agree, but for a different reason.** It's right because the field is a pure restatement of two already-stable fields, so stable costs nothing. The stated premise ("beta fields get ignored") doesn't hold in this repo — MCP/CLI/web all consume beta surface (`hold_reason`, `lifecycle_transitions`, protection). Ship stable; just don't lean on that premise.
2. **Keeping the inbound chip copy `Verified`/`Unverified`** — agree; the new outbound chip is deliberately distinct from "Sending enabled", and copy churn buys nothing.

## Version note

No bump needed in this PR (publishing is tag-triggered). When the SDKs next publish this is a **minor** for both — additive response field, no break for same-version servers. The one coupling to record in both changelogs is the version-skew note above: an SDK carrying `capabilities` requires a server that emits it.

---
*Review methodology: four parallel read-only review passes (API contract & evolvability, SDK parity, security & auth, docs/spec accuracy) over the branch at `021112d0`, with runtime execution of both SDK deserializers and the Go test gates; findings cross-corroborated and adversarially re-checked before inclusion.*

jiashuoz:
Thanks — validated every finding against the code rather than taking the summary at face value. Pushed `b4c51cc4`. Six applied, three declined with reasons, and one severity re-graded.

## Applied

**Web seam bypass** — confirmed at all three sites and fixed onto `canReceive`: `CustomDomainChecklist.tsx:25` (phase derivation), `DomainSelector.tsx:76,81` (per-domain chip), `inboxes/page.tsx:145` (dashboard stat). This was the sharpest finding — it violated the invariant three lines into the module the PR itself added.

**`capabilities.inbound` vocabulary** — trimmed to `verified`/`pending`. Rather than just deleting the impossible values, the doc now states the consequence you identified: a missing or wrong MX leaves the axis pending *indefinitely*, so `pending` means "not proven yet", not "in flight", and there is no inbound equivalent of `sending_error`. That's the part a polling client actually needs. Defining and emitting a real `failed` state is a larger change than this PR should carry.

**Contract-test pin** — added to `tests/contract/scenarios.yaml`: both axes asserted at register, plus a new post-verify GET step pinning that inbound moves to `verified` while outbound stays put. Since the shared scenario file drives the Go, TS, and Python runners, that's one edit pinning all three rather than one assertion per SDK. I mutation-tested it to confirm it bites (`step get_after_verify: path "capabilities.inbound" = verified, want SHOULD_FAIL`) — a pin that can't fail isn't a pin.

**Plugin skill** — added the outbound-axis step to the custom-domain flow.

**Both PR-body corrections** — you were right twice. `doctor.ts:31` does allow new check IDs, so the frozen-schema rationale was wrong (decision stands, reason was bogus — now says redundancy). And the "clients ignore beta surface" premise is false here: `mcp/src/tools/{review,legacy}.ts` and `web/.../reviews/_components/*` all consume `hold_reason`. Same conclusion, honest reasoning.

## Re-graded: version skew is real but not this PR's

Reproduced the `ValidationError` by execution. But the control run is decisive: **removing `sending_ramp` from the same payload raises the identical error.** So an SDK carrying `sending_ramp` already hard-fails against a server predating it — this PR adds one instance of a repo-wide property of Huma's required-field marking, it doesn't introduce the class. Your own text says as much; I'd just not rank it as this PR's headline MAJOR. The undocumented "SDK ≥ server" policy is a genuine pre-GA item and is now recorded in the PR body as a known follow-up.

## Declined

- **Object-level schema description** — only **7 of 145** component schemas have one; `DomainView`, `SendingRampView`, and `DNSRecord` all lack it. Adding it to mine alone *creates* the inconsistency. Worth a repo-wide pass or nothing.
- **Web closed `Literal` union over an open set** — deliberate: it matches `DomainSendingStatus` and `DNSRecordStatus` in the same file, and both chips plus `canReceive` fall through safely on unknown values. Changing only the new type would be the outlier.
- **"Outbound pending" for `none`** — as you noted, the pre-existing `SendingStatusChip` collapses identically; distinguishing one and not the other trades a small inaccuracy for an inconsistency.

## One you missed

`github-code-quality` flagged 5 unused imports in the generated SDK files, and it was right — my first instinct to wave it off as "generated code" was wrong. The normalizer takes an **explicit symbol/file allowlist**, and the precedent is this field's closest analogue: `HttpFile "$OUT/models/SendingRampView.ts"` plus `SendingRampView` in the three `types/*.ts` files and `re` for `sending_ramp_view.py`. I never extended it for `DomainCapabilities`, so the new files carried unused imports their peer doesn't. Fixed in both `generate-oag.sh` entrypoints and regenerated — not by editing generated output, which the next `make generate-sdk` would revert. Normalizer counts moved 20→24 (TS) and 9→10 (Python), matching the five new pairs.

Worth noting for future passes: four review dimensions over the diff didn't catch it, because it isn't visible in the diff's logic — only in the codegen pipeline's configuration.

PR #685, opened 2026-07-25
feat(contacts): account-level contact identity + bulk import (beta)

jiashuoz:
Contacts for e2a: a durable account-level record of the people an account
corresponds with, so an agent doing outreach at volume has somewhere to keep
who it is talking to. Two slices, one feature.

Design: `docs/design/2026-07-24-contacts-and-outreach-state.md` (included).

## Slice 1 — contact identity

- Migration `079_contacts.sql` — `contacts` keyed `(user_id, address)`, with
  provenance (`source`, `import_batch_id`) and an opaque `metadata` blob.
- `/v1/contacts` list / create / get / update / delete.

## Slice 2 — bulk import

- `POST /v1/contacts/import` — up to 1000 rows in one transaction, per-row
  outcomes, so one malformed line never rejects an upload.
- `DELETE /v1/contacts/imports/{batch_id}` — reverses an import, reporting what
  it removed and what it deliberately kept.

All operations are **beta** and account-scoped, exposed through both SDKs'
ergonomic clients.

## The modeling decision

Contact identity is **account-level**; per-agent outreach state (stage, next
action, reply status) is a separate resource in a later slice. This mirrors the
existing `suppressions` / `agent_suppressions` split and exists for the same
reason: consent and outreach stage are per-relationship, identity is not.
Nesting contacts under an agent would give one human a row per agent, which
makes "has anyone on our side already contacted this fund?" unanswerable — the
question a multi-person raise most needs.

## Contract notes

- Addresses are canonicalized with `NormalizeMailboxAddress` — **the same key
  suppression lookups use**. A spreadsheet's `Name <addr>` form and the bare
  address are one resource. Divergence would make a contact look sendable and be
  blocked at send time; `TestContactKeyMatchesSuppressionKey` pins it.
- **Omitted means "leave alone"** everywhere — `PATCH` and import rows alike. A
  narrower re-upload cannot erase columns it no longer carries.
- Import is inert: it records identity and sends nothing. Suppressed addresses
  are imported and **marked**, not dropped, so the count matches the file.
- Cursors pin their filter fingerprint; a continuation with changed filters is
  rejected rather than interleaving two queries.
- `metadata` is bounded on size, key count, key length, value size, and nesting.
- Deleting a contact leaves suppressions intact — consent outlives the record.

## Verification

- Go: 45/45 packages. TS SDK: 213 tests. Python SDK: 478 tests.
- `spec-check`, `generate-sdk-check`, `openapi-compat-check` (no breaking
  changes), SDK operation coverage 68/68 reachable from both SDKs.
- **Over-the-wire e2e** against a real binary + Postgres for both slices.

Three defects were caught only by driving a real instance, and none of them
could have been caught by the unit suite:

1. Handlers were registered but never wired into `BuildDeps` — every endpoint
   returned 501 in the real binary while all handler tests passed, because they
   inject fakes directly. `TestContactDepsAreWired` now guards it.
2. Re-importing without a `metadata` column wiped the metadata.
3. Re-importing without a `display_name` column wiped the name.

## Known gaps (documented, not hidden)

- `If-Match` compare-and-write is **not atomic** — two writers with the same
  valid ETag can both be accepted. The published description says so; tightening
  it to a conditional write is a pre-stable follow-up.
- Deleting a contact is **immediate and permanent**, unlike agents and messages
  which are trash-backed. Worth deciding whether contacts join the trash model
  before this leaves beta.
- **CLI and MCP surfaces are not yet implemented** for contacts.

## Follow-ups

- Slice 3: engagements (per-agent outreach state, `agent_email` + `stage` on
  import). Slice 4: `contact.due`.
- `Deps.CountContacts` is wired but unused — lands with `contact_limit_reached`.


jiashuoz:

PR #686, opened 2026-07-25
fix(logging): redact PII from process logs before centralized log shipping

jiashuoz:
## Why

Container stdout is about to ship to Google Cloud Logging (30-day retention, org-queryable). Today the server logs email **subject lines** and **external human email addresses** in plaintext, plus a few adjacent leaks (account-owner emails, free-text HITL reasons, unbounded third-party error bodies, full connecting IPs). Once logs are centralized, every one of those lines becomes replicated personal data outside the database's access controls. This PR removes/reduces them while keeping the operational signal. It is standalone — no structured-logging conversion, only value redaction in the existing `log.Printf` lines.

**Direction rule applied per-site:** on inbound, the sender is an external human (redact to domain) and the recipient is an e2a agent address (kept in full — our namespace, the tracing key). On outbound it flips: agent `from` kept, external recipients reduced to count + distinct domain set.

## New package

`internal/logredact` — `AddressDomain`, `AddressDomains` (distinct, sorted, capped at 10 + overflow marker), `IPNetwork` (/24 v4, /48 v6), `Truncate` (rune-safe). All panic-safe on hostile input (they run in the SMTP receive path); unit-tested including malformed/unicode/no-panic cases. `logguard_test.go` is a source-scan regression guard (house style of `internal/httpapi/response_enum_stance_test.go`): it fails if `subject=%q`/`%s`/`%v` logging reappears anywhere in non-test Go source, with an allowlist-with-reason escape hatch.

## Redactions, and what an operator loses at each

(line numbers are post-change)

**Subject lines → `subject_len=%d`** (full subject stays on the message row in Postgres). Lost: grepping logs by subject; you now pivot via `msg_` id → DB.
- `internal/relay/server.go:688` (every inbound message)
- `internal/agent/api.go:1124, 1314, 1378, 1492` — `internal/agent/hitl_api.go:154, 180`

**External addresses → domain only** (full addresses on the message/intake rows in Postgres). Lost: correlating one specific human sender across sessions from logs alone; domain + trace/session id remains.
- Inbound sender: `internal/relay/server.go:251, 258, 265, 270, 284, 288, 313` (MAIL FROM / RCPT / rejections / DATA), `:399` (auth line), `:667` (persist failure), `:688` (`from_domain=`). Recipient/agent addresses on these lines kept in full.
- Outbound recipients → `to_count` + `to_domains`: `internal/agent/api.go:1124, 1328, 1339, 1371, 1378, 1458, 1469, 1488, 1492`, `internal/agent/hitl_api.go:154, 180`, `internal/agent/hitl_magic_api.go:274, 308`, `internal/hitlworker/worker.go:342, 455`, `internal/outbound/smtp_relay.go:75` (retry line).

**Account-owner (login) emails → user id only** (email on the users row; for the deletion line, in the export the user could take before deleting).
- `internal/agent/user_data_rights_api.go:93` — the GDPR/CCPA deletion audit line no longer logs the deleted person's address.
- `internal/hitlnotify/notifier.go:162` — `owner=` is now the user id.

**Free-text human input → length (or 60-rune preview)**
- HITL/review reject reasons → `reason_len`: `internal/agent/hitl_api.go:374, 434`, `internal/agent/hitl_magic_api.go:357`. Full reason is stored (`rejection_reason` column) and delivered in the `email.review_rejected` event. Lost: reading the reason in the log stream.
- Feedback fallback `internal/agent/api.go:1651`: now `message_len` + 60-rune preview (was 200-rune dump). Honest caveat: when no delivery channel is configured (self-host dev), the log **was the only sink**, so content beyond 60 runes is genuinely lost there. Hosted prod configures a delivery channel, so nothing is lost in our deployment.

**Third-party error bodies → status code only**
- `internal/auth/oidc.go:305-319` — `*oauth2.RetrieveError` formats the provider's FULL raw HTTP response body; now logs HTTP status + RFC 6749 error code, other errors hard-truncated to 200 runes. Lost: the provider's freeform error prose; usually reproducible.
- `internal/agent/user_data_rights_api.go:131` — billing-hook error keeps the status, drops the sidecar body. The sidecar logs its own errors, so the detail still exists on the billing container.

**Connecting IPs → network prefix** (/24 / /48)
- `internal/relay/server.go:221, 224` (session open) and `:399` (auth line). Lost: exact-IP correlation from logs; prefix-level abuse/rate-limit debugging preserved. Exact IP is still used (not logged) by SPF evaluation.

**Duplicate removed instead of redacted twice**
- `internal/emailauth/checker.go:104`: the library-level "Email auth for X from Y" line duplicated relay `server.go:399` byte-for-byte in production (the relay is the only production caller via `srv.authenticate`) but without the trace id. Removing it beats maintaining redaction in two places; a comment marks the decision. Non-relay/library consumers of `CheckAuthentication*` lose that implicit log line — none exist in this repo.

## Judgment calls / found-in-flight

- **Extra leak found and fixed** (not on the audit list): `internal/hitlworker/worker.go:280` built an auto-reject reason containing the full suppressed recipient addresses, which flowed verbatim into the `reason=%q` auto-reject logs. Reason now carries count + domains. This also changes the *stored/emitted* rejection reason — the owner identifies the exact address via the message's stored recipient list or the suppression API. `suppression_test.go` updated to assert the new contract (and that the local part does NOT appear).
- **Deliberately NOT redacted:** agent addresses everywhere (our namespace, primary tracing key), including the loopback self-send line `api.go:1314` where `from`/`to` are both the agent itself; `comp.EnvelopeFrom` on the platform-send line `api.go:1492` (platform address at our domain); rejected-RCPT target addresses in `relay/server.go` (addresses in our MX namespace — the typo'd local part is the debugging signal); worker auto-reject `reason=%q` at `worker.go:577/581` (system-generated strings, now address-free after the suppression fix — though embedded `%v` compose/send errors are a residual theoretical path); the magic-link HTML responses showing the recipient (user-facing HTTP to the owner, not logs).
- **Ambiguity flagged:** `relay/server.go:399` also logs the connecting IP; the task listed that line for address redaction only, but with the checker.go duplicate removed this became the sole copy, so I applied the same /24 truncation there for consistency.

## Tests

- `internal/logredact`: full unit coverage incl. malformed/empty/unicode/no-panic; `TestNoSubjectContentInLogFormats` regression guard.
- Changed test: `internal/hitlworker/suppression_test.go` (asserted the old full-address reason).
- `go build ./...` clean; `go test ./...` green for every touched package.
- **Pre-existing, unrelated:** `internal/identity` DB-backed tests fail flakily (random FK-violation subsets, e.g. `domains_user_id_fkey`) on my local test DB — reproduced identically on a clean `origin/main` checkout, so not introduced here. Also pre-existing: `go vet` nits in agent test files and gofmt drift in ~20 untouched files.

### DIFF
--- a/internal/agent/api.go
+++ b/internal/agent/api.go
@@ -30,6 +30,7 @@ import (
 	"github.com/tokencanopy/e2a/internal/identity"
 	"github.com/tokencanopy/e2a/internal/inboundscreen"
 	"github.com/tokencanopy/e2a/internal/limits"
+	"github.com/tokencanopy/e2a/internal/logredact"
 	"github.com/tokencanopy/e2a/internal/oauth"
 	"github.com/tokencanopy/e2a/internal/outbound"
 	"github.com/tokencanopy/e2a/internal/piguard"
@@ -1120,8 +1121,8 @@ func (a *API) HoldForApprovalCore(ctx context.Context, agent *identity.AgentIden
 	}
 
 	slug, _, _ := strings.Cut(agent.EmailAddress(), "@")
-	log.Printf("[mail:%s] dir=outbound type=%s status=%s from=%s to=%v slug=%s conv_id=%s subject=%q approval_expires_at=%s",
-		msg.ID, msgType, msg.Status, agent.EmailAddress(), req.To, slug, req.ConversationID, req.Subject, msg.ApprovalExpiresAt.Format(time.RFC3339))
+	log.Printf("[mail:%s] dir=outbound type=%s status=%s from=%s to_count=%d to_domains=%v slug=%s conv_id=%s subject_len=%d approval_expires_at=%s",
+		msg.ID, msgType, msg.Status, agent.EmailAddress(), len(req.To), logredact.AddressDomains(req.To), slug, req.ConversationID, utf8.RuneCountInString(req.Subject), msg.ApprovalExpiresAt.Format(time.RFC3339))
 
 	a.publishPendingApproval(ctx, a.buildPendingApprovalEvent(agent, msg, req, msgType), msg.ID)
 	return msg, nil
@@ -1310,7 +1311,7 @@ func (a *API) DeliverOutbound(ctx context.Context, user *identity.User, agent *i
 		// on metering; the pre-check remains the quota gate.
 		a.recordLoopbackUsage(ctx, user.ID, agent)
 		slug, _, _ := strings.Cut(agent.EmailAddress(), "@")
-		log.Printf("[mail:%s] dir=outbound type=%s method=loopback status=sent from=%s to=%s slug=%s conv_id=%s subject=%q", outMsg.ID, msgType, agent.EmailAddress(), agent.EmailAddress(), slug, req.ConversationID, req.Subject)
+		log.Printf("[mail:%s] dir=outbound type=%s method=loopback status=sent from=%s to=%s slug=%s conv_id=%s subject_len=%d", outMsg.ID, msgType, agent.EmailAddress(), agent.EmailAddress(), slug, req.ConversationID, utf8.RuneCountInString(req.Subject))
 		return &OutboundResult{MessageID: outMsg.ID, SentAs: "own_address", Method: "loopback"}, nil
 	}
 
@@ -1324,7 +1325,7 @@ func (a *API) DeliverOutbound(ctx context.Context, user *identity.User, agent *i
 	// (email.sent / email.failed + metering). Missing queue wiring is a startup bug;
 	// fail closed here as defense in depth and never submit inline.
 	if a.outboundEnq == nil {
-		log.Printf("[api] outbound queue unavailable: agent=%s to=%v", agent.Domain, req.To)
+		log.Printf("[api] outbound queue unavailable: agent=%s to_count=%d to_domains=%v", agent.Domain, len(req.To), logredact.AddressDomains(req.To))
 		return nil, &OutboundError{Status: http.StatusInternalServerError, Code: "internal_error", Msg: "outbound delivery queue unavailable"}
 	}
 	comp, cerr := a.sender.ComposeForAccept(agent, req)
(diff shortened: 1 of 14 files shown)


jiashuoz:
## Review findings applied (b8fb8eab)

All seven applied. The two mediums share a theme: the first cut redacted in the **wrong layer**, degrading a customer-visible string to protect a log line.

| # | Sev | Finding | Fix |
|---|-----|---------|-----|
| 1 | MED | Suppression reason degraded to domains-only — but that string is stored (`messages.rejection_reason`), served (`RejectResultView.RejectionReason`), and emitted (`email.review_rejected`) | Reverted to the precise address list; suppression test reverted to assert `bob@external.test` |
| 2 | MED | `reason=%q` printed verbatim at two `autoReject` log sites | Both now `reason_len=%d` |
| 3 | MED | Unauthenticated RCPT probing injected third-party addresses into logs | Pre-resolution `to` redacted to domain; false comment corrected |
| 4 | LOW | `lastErr` raw — MTA rejections quote the recipient back | `Truncate(…, 200)` + comment |
| 5 | LOW | Guard regex case-sensitive, `=`-only | Case-insensitive, `[:=]` separator, conjunctive with a log-site match |
| 6 | LOW | `AddressDomain` returned display names via the RFC 5322 comment form | Hostname-shape validation + tests |
| 7 | LOW | OIDC `ErrorCode` un-truncated | `Truncate(…, 200)`, matching the sibling branch |

### Judgment calls worth a look

**#2 — `reason_len` rather than `Truncate`.** Truncation bounds *size* but does not *redact*: a 40-char address survives any sensible cap, so `Truncate` would have left the leak open while looking fixed. The exact reason stays recoverable from the message row via the message id already on both lines. This is what makes #1 safe — the reason is precise everywhere it is a contract, and reduced only in the log copy.

**#5 — the broadened regex needed a second predicate.** Adding `[:=]` + case-insensitivity alone produced two *real* false positives on `main`:

- `internal/hitlnotify/notifier.go:232` — `fmt.Fprintf(&b, "Subject: %s\n", msg.Subject)`, composing the notification email body
- `internal/piguard/gemini.go:213` — building the LLM prompt

Neither is a log. Rather than allowlist two whole files (which would blind the guard to any *future* real violation in them), the match is now conjunctive: subject-content pattern **AND** a log/error emission site on the same line. `TestSubjectLogViolationDiscrimination` pins both directions. Doc comments no longer oversell — it is a subject-only tripwire, and the package comment now says the rest of the policy is code-review's job.

**#3 — restructured rather than patched.** The old comment asserted the agent-namespace invariant unconditionally, which is exactly what made the leak look already-considered. The invariant is now stated as beginning *at* `resolveAgent` success; domain-not-verified and limit-exceeded keep the full address as before (verified agent rows).

**#6 — cut-then-validate.** Domain is truncated at the first non-hostname rune, then must be hostname-shaped or it degrades to `invalid`. IDN labels (`bücher.example`) still pass. New coverage: comment form (spaced and unspaced), display-name-plus-comment, stray brackets, address literals, quoted/control-character domains, leading/trailing dots, a display-name non-leak property test, and four more no-panic fuzz shapes.

### Also
Documented the two rules these findings imply in the package comment: an agent address is only log-safe once **resolved**, and redaction belongs at the **log sink** whenever the value is also stored, served, or emitted.

### Verification
`go build ./...` clean; `go test ./...` green across all 49 packages (exit 0); `gofmt -l` clean on every touched file. Pre-existing, untouched: `internal/auth/cli_login_test.go` gofmt drift (present on `origin/main`).

No slog/structured-logging work — still out of scope.

PR #689, opened 2026-07-25
feat(contacts): per-agent outreach state (beta) — slice 3

jiashuoz:
**Stacked on #685** — targets `feat/contacts`, since this builds on the contacts
table from slices 1–2. GitHub retargets to `main` when #685 merges.

This is the slice that turns contacts from an address book into the outreach
loop. Until now nothing could answer the question the feature exists for:

> which of my 300 investors haven't replied in 5 days and need touch 2?

Now that's one request.

## What's here

- **Migration `080`** — `contact_engagements` keyed `(user_id, agent_id, contact_id)`.
- **Store layer** — upsert / get / list / delete, plus `PurgeEngagementsForAgent`.
- **`/v1/agents/{email}/contacts`** — list, get, `PUT` enrol/update, `DELETE` un-enrol.
- Outreach methods on both SDK ergonomic clients.

## Why a separate table from contacts

Identity is account-level (one row per human); consent and outreach stage are
per-relationship. The same partner may be worked by `raise@` and `support@` with
independent history, and an unsubscribe from one is not an unsubscribe from the
other. Same two-level shape as `suppressions` / `agent_suppressions`.

## The authorization divergence

This is **the one contact surface an agent-scoped credential may reach**.
`/v1/contacts` is account-only — account-wide identity would expose every person
a sibling agent corresponds with. Engagements are different: they're the runtime
state the agent itself authors, and an agent that can't read its own due list
can't do its job.

It's safe because consent lives in `agent_suppressions`, which this surface only
ever reads through a join. An agent can move its own stage and can **never**
un-suppress itself. Both halves are tested: the agent *can* drive its own
outreach, and *cannot* reach a sibling agent or account-wide identity.

## The lifetime asymmetry

`agent_id` carries no FK, matching `agent_suppressions` — but with the opposite
lifetime, which is the subtle part. Because `agent_id` **is** the agent's email
address, a recreated agent inherits whatever is left behind. For consent that's
correct and deliberate. For outreach state it would be a misfire: a recreated
`raise@` inheriting last year's `stage: "touch3"` and past-due schedule would
wake up and mail investors it never contacted.

So engagements survive trash and are purged on hard delete, while suppressions
survive both. `TestPurgeEngagementsForAgentSparesConsent` pins exactly that.

## Contract notes

- **Derived fields are absent from the request struct** — a client cannot fake a
  reply or a send; attempting it is a 422, not a silent no-op.
- **`next_action_at` is tri-state**: omitted leaves the schedule alone, `null`
  clears it. This is what stops every stage advance from rewinding the schedule.
- **`replied` is computed**, not stored (`last_inbound_at > first_outbound_at`),
  so it can't drift. Comparing against *first* outbound makes it mean "replied
  to us" rather than "has ever written".
- **Suppression is joined, not copied**, so consent has one home and this view
  can't disagree with what the send path enforces. The reason is surfaced too —
  a dead address and someone who said no need different handling.
- **`last_outbound_before` is documented prominently** on the list.
  `last_outbound_at` is server-maintained, so including it in the sweep drops
  anyone just contacted *even when the client's own state write was lost*.
  Omitting it is a live double-send hole.

## Verification

- Go 45/45 packages · TS SDK 216 tests, function coverage 81.18% · Python 478 tests.
- `spec-check`, SDK operation coverage 72/72 reachable from both SDKs.
- Migration applied and verified idempotent.
- Store wired into `BuildDeps` **in this slice**, with the wiring guard extended —
  the 501-in-production failure mode from slice 1 can't recur silently.

## Not in this slice

- **`contact.due`** (slice 4) — the wake-up event. `notified_next_action_at` and
  the partial `ce_due_idx` are in the schema ready for it.
- **Derived-field maintenance** — the columns exist and are read, but nothing
  populates them from real message activity yet. That's the next piece of work
  after the due event, and until it lands `replied` and the counts stay at their
  zero values.
- `agent_email` + `stage` on import; CLI and MCP surfaces.


jiashuoz:

PR #690, opened 2026-07-25
feat(contacts): engagement activity recorder and reconciliation (beta)

jiashuoz:
**Stacked on #689** (which is stacked on #685).

Store-side maintenance for the derived columns the outreach query reads:
first/last contact timestamps, message counts, and last conversation.

## Status: half the slice

This PR contains the **recorder and the reconciliation sweep, with tests**. It
does **not** yet contain the hooks that call them from the live send and receive
paths — see "Deliberately not done" below. Until those land, `replied` and the
counts stay at their zero values, so the outreach query is structurally correct
but the data is not flowing.

## Design decisions worth reviewing

**Update-only, never create.** Neither recorder creates an engagement. An agent
sends mail for all sorts of reasons — replies, one-off notes, transactional
messages — and auto-enrolling every recipient would fill the due list with
people nobody is running a campaign against, which is the noise this feature
exists to remove. Enrolment stays an explicit act (`PUT` or an import).

**`first_outbound_at` is pinned.** Set once, never moved, because `replied` is
defined against it — moving it on a later send would silently un-reply everyone
who had already answered. Tested directly.

**Timestamps use `GREATEST`.** A retried or late-arriving record cannot move a
timestamp backwards. At-least-once delivery makes out-of-order arrival normal
rather than exceptional.

**Reconciliation reports, not just corrects.** Drift is the known cost of
materializing these counters instead of computing them on read (the class
`AGENTS.md`'s schema-change rule warns about). The sweep recomputes from
`messages`, corrects, and returns what it found — a non-empty result means a
hook was missed. It is idempotent, so that signal stays trustworthy; a second
pass returning nothing is asserted.

## Verification

Go 45/45 packages. Five new store tests covering: update-only for both
directions, `first_outbound_at` pinning (including that a later send does not
un-reply someone), out-of-order timestamps, per-agent scoping, and drift
detection + correction + idempotency.

## Deliberately not done in this PR

The hooks that call these from `MarkSent` (outbound) and the relay after
`resolveConversationID` (inbound). That edit lands in the live send and
inbound-receive paths — the highest-risk code in the system, where a mistake
means dropped or duplicated real mail. It deserves its own change with room to
test it properly rather than being appended here.

Also outstanding: the janitor periodic wiring for the reconciliation sweep, and
`contact.due` (slice 4).


jiashuoz:

PR #692, opened 2026-07-25
test(selftest): make WS stubs read so close handshakes complete

jiashuoz:
## The issue

`internal/selftest` took **10.6 seconds**, and almost all of it was pure waiting — every invocation of `scenarioWebSocketRoundTrip` cost ~5s inside its deferred `conn.Close`.

`nhooyr.io/websocket`'s `Conn.Close` writes a close frame and then waits up to **5s** for the peer's close frame (`close.go:211`, `waitCloseHandshake`). Control frames are only processed while something *reads* the connection. All three stubs in `ws_scenario_test.go` accepted the upgrade and then never read — so the close frame was never answered, and every `Close` burned the full timeout.

## It is a test-fidelity gap, not a product defect

I checked the real handler before assuming: `internal/ws/handler.go:377` runs an explicit read loop and classifies `websocket.StatusNormalClosure` as `"client_close"`. So the real server *does* answer the handshake, and production and the prober are unaffected. The stubs were simply less faithful than the thing they stand in for.

The fix makes them mirror it via `CloseRead` — the push-only equivalent: it answers close frames, leaves writes working, and expects no client messages, which is correct since this transport is server-push.

One trap worth recording, because I hit it: **`context.Background()`, not `r.Context()`.** `Accept` hijacks the connection so it outlives the handler, and the request context is cancelled the moment the handler returns — which tore the socket down before the push could be written (`failed to read frame header: EOF`).

## Results

| | before | after |
|---|---|---|
| `internal/selftest` package | 10.57s | **0.686s** (−94%) |
| `TestScenarioWebSocketRoundTrip` | ~5.0s | 0.00s |
| `TestScenarioWebSocketRoundTrip_Fail` | 5.20s | 0.23s |

**No assertion was removed or loosened.** The happy path still requires a pass and still pins that *both* message copies are trashed. The no-frame case still pushes nothing and still must fail on the probe's own round-trip timeout — and now genuinely is bounded by it, clearing that sub-case's 5s cap by an order of magnitude rather than running near it.

## Verification

- `go test ./internal/selftest -count=5` green; `-race -count=3` green
- `make test-unit` green (exit 0)
- One file changed, test-only: +22/−2

## Context

This came out of investigating why the Go coverage gate went red on `main` in a ~90-minute window on 2026-07-25. That window produced four failures in two classes:

- **Docker Hub unreachable** (2 of 4) — `docker pull` timeouts against `registry-1.docker.io`. Not a repo bug; see the follow-up note below.
- **Tests with no timing headroom** (2 of 4) — `TestTruncateAll_CleansInboundIntakeWithoutExclusiveTableLock` (fixed in #687) and `TestScenarioWebSocketRoundTrip`.

This PR does **not** fix the WS test's CI failure mode. That failure was the push not arriving within the probe's 200ms budget, which I measured to have large headroom locally (the scenario passes even at a 1ms budget), meaning CI hit goroutine starvation of >200ms on a saturated runner. That is worth hardening separately; this PR is about the 5s that was being burned on *every* run, pass or fail.

### Follow-up worth its own issue: 6 Docker Hub pull points

`test.yml` depends on Docker Hub in six places with no mirror or fallback — `postgres:16-alpine` in four jobs (`:90, :129, :273, :482`), `axllent/mailpit:v1.30.0` (`:146`), and `openapitools/openapi-generator-cli:v7.16.0` pulled by `make generate-sdk-check` (`:582`). The runner's built-in retry made three attempts and gave up. Mirroring these into GHCR — where the project already publishes — would remove the failure mode that caused half the observed red runs.

### DIFF
--- a/internal/selftest/ws_scenario_test.go
+++ b/internal/selftest/ws_scenario_test.go
@@ -55,6 +55,20 @@ func wsStub(t *testing.T) (*httptest.Server, *wsStubState) {
 			st.mu.Lock()
 			st.conn = c
 			st.mu.Unlock()
+			// Consume incoming frames, mirroring the real handler's read loop
+			// (internal/ws/handler.go). Control frames are only processed while
+			// something reads, so a stub that never reads never answers the
+			// client's close frame — and the scenario's deferred conn.Close then
+			// blocks for the library's full 5s close-handshake timeout, making
+			// every WS scenario invocation cost 5s of pure waiting. CloseRead is
+			// the push-only equivalent: it answers close, keeps writes working,
+			// and expects no client messages (this transport is server-push).
+			//
+			// NOT r.Context(): Accept hijacks the connection so it outlives this
+			// handler, and the request context is cancelled the moment the
+			// handler returns — which tore the socket down before the push could
+			// be written ("failed to read frame header: EOF").
+			c.CloseRead(context.Background())
 		case r.Method == http.MethodDelete:
 			parts := strings.Split(r.URL.Path, "/")
 			st.mu.Lock()
@@ -127,7 +141,11 @@ func TestScenarioWebSocketRoundTrip_Fail(t *testing.T) {
 		if strings.HasSuffix(r.URL.Path, "/ws") {
 			c, err := websocket.Accept(w, r, &websocket.AcceptOptions{InsecureSkipVerify: true})
 			if err == nil {
-				_ = c // hold open, push nothing
+				// Hold open and push nothing — but still read, so the deferred
+				// conn.Close is answered instead of burning the library's 5s
+				// close-handshake timeout. The scenario must fail on its own
+				// round-trip timeout, which is what this case asserts.
+				c.CloseRead(context.Background())
 			}
 			return
 		}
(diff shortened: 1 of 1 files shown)


jiashuoz:
Added `ccd06c15` — a deterministic regression test, because the original commit had a real gap: it made the suite 5s faster but **nothing failed if the fix were reverted.** CI would have stayed green while the cost silently came back.

The assertion is on behavior, not wall clock. The happy-path stub's read loop now records whether it observed the client's normal-closure frame, and the test asserts it did. That is the exact signal — `conn.Close` blocks until the peer answers or its 5s timeout expires, so a false reading means the handshake did not happen.

Deliberately **not** a duration bound. A wall-clock budget is the precise flake pattern this suite was just bitten by (and #687 before it); it would go red under load for reasons having nothing to do with the handshake. This assertion has zero timing sensitivity.

That is also why the stub uses an explicit read loop rather than `CloseRead`: the close frame needs to be both *answered* and *observable*. It mirrors `internal/ws/handler.go:377` more literally as a side benefit.

Red/green proof — reverting just the read loop:

```
ws_scenario_test.go:145: stub never observed the client's close frame: the close
    handshake did not complete, so conn.Close burned the library's full 5s timeout
--- FAIL: TestScenarioWebSocketRoundTrip (5.00s)
```

The assertion fires *and* the duration jumps 0.00s → 5.00s, demonstrating the cost it guards. Restoring the loop returns both to green.

`-count=3` and `-race -count=2` green locally; full CI green on this branch (27 checks).

PR #699, opened 2026-07-25
feat(contacts): call the engagement recorder from the send and receive paths

jiashuoz:
**Stacked on #690** (→ #689 → #685). This is the fourth and riskiest slice.

Wires the derived-field recorder from #690 into the live paths, so `replied`,
the timestamps, and the counts actually populate. Until this lands the outreach
query is structurally correct but reads all-zero data.

## The one question a reviewer should ask

**Can this fail a send or reject inbound mail?** It is designed so the answer is
no, and that is worth checking directly:

- **Outbound** — `outboundSendStore.MarkSent`, called *after* the terminal
  transaction commits. `finalizeSentTx` is the canonical finalizer, so the
  worker path is covered.
- **Inbound** — `internal/relay`, in the existing post-commit best-effort
  section alongside usage metering and protection events, where the conversation
  is already resolved.

Both are **outside their transactions and non-fatal**. That is deliberate and
differs from metering next door: metering is accounting and must roll back
rather than undercount, but these counters are derived convenience data with
`ReconcileEngagementCounts` behind them. Failing an already-submitted send, or
returning a 451 on real inbound mail, because a bookkeeping row would not update
is strictly worse than letting the sweep converge it. Doing it *inside* the
transaction would be worse still — a failed statement aborts the whole commit.

Update-only on both sides, so an agent's ordinary correspondence (replies,
one-off notes, transactional mail) does not silently enroll people into an
outreach list.

## Verification — and its limit, stated plainly

Go **48/48 packages** pass, and the recorder itself has five store tests from
#690 covering update-only, `first_outbound_at` pinning, out-of-order timestamps,
per-agent scoping, and drift correction.

**What is NOT proven here:** that the hooks actually fire against a live send and
a live inbound message. The build compiles and nothing regressed, but no test in
this PR drives a real message end to end and asserts the counters moved. That is
exactly the gap that hid the unwired-`BuildDeps` bug in slice 1 — a unit-tested
component wired to nothing.

Before merging this, one of:

1. An over-the-wire e2e: enrol a contact, send to it, assert `outbound_count`
   and `last_outbound_at` moved; deliver an inbound reply over SMTP, assert
   `replied` flips.
2. An integration test at the relay and outbound-worker seams asserting the same.

I would not merge on the current evidence.

## Not in this slice

- Janitor periodic wiring for `ReconcileEngagementCounts` (the sweep exists and
  is tested; nothing schedules it yet).
- The provider-accepted crash-window bridge (`FinalizeProviderAcceptedTx`) runs
  inside a caller's transaction and is not hooked — the reconciliation sweep is
  what covers that path.
- `contact.due` (slice 5).


jiashuoz:
## E2E verification done — and it found a real bug

I flagged this PR as not-mergeable without an end-to-end run. Done now, against
a live instance with Postgres and Mailpit. It caught a defect that made the
outbound hook a **complete no-op in production**.

### The bug (fixed in 839a6b4d)

The hook keyed the engagement update off `info.Message.Recipient`. The terminal
path never populates that field — `MarkOutboundSentTx` scans `ToRecipients` into
an array and leaves the scalar empty. So the `UPDATE` matched on `""` and
affected zero rows: no error, no log, counters frozen at zero while every test
passed.

No unit test could have caught it. They inject a fake `OutboundSentInfo` with
`Recipient` set — the field that is empty in production is populated in every
test. Same shape as the unwired-`BuildDeps` bug from slice 1: a well-tested
component wired to something that doesn't exist.

Iterating `ToRecipients` is also more correct than the scalar was: one send to
several people now advances each of their engagements.

### Verified end to end, with real mail

| Step | Result |
|---|---|
| Enrol contact | `out=0 in=0 replied=false` |
| Send a real message | `out=1`, `first_outbound_at` + `last_outbound_at` set |
| Real SMTP reply into `:2525` | `in=1`, `last_inbound_at` set, **`replied=true`** |
| Due sweep `?replied=false&next_action_before=…` | returns **only** the contact who hasn't replied |
| Mail a **non-enrolled** address | no engagement created — update-only holds |

Zero panics or 5xx in the server log. Test databases dropped, tree clean.

That last table is the motivating user journey working for real: send, get a
reply, and the person drops out of your follow-up list automatically.

### Still outstanding before merge

- Janitor periodic wiring for `ReconcileEngagementCounts` (sweep exists and is
  tested; nothing schedules it).
- `FinalizeProviderAcceptedTx` — the crash-window bridge — runs inside a
  caller's transaction and is not hooked; the reconciliation sweep is what
  covers that path.
- No automated regression test pins the `ToRecipients` behaviour. The e2e above
  proves it today, and the reconciliation sweep would surface a recurrence as
  drift, but neither is a unit-level guard.

jiashuoz:

PR #702, opened 2026-07-25
fix(contacts): purge engagements with the agent and schedule the drift sweep

jiashuoz:
**Stacked on #699** (→ #690 → #689 → #685).

Two capabilities that already existed, already had passing tests, and were
**never called**. Both are the same failure shape that has now bitten this
feature three times: a well-tested function wired to nothing.

## 1. Engagements were not purged with the agent

`PurgeEngagementsForAgent` had a green store test and no caller anywhere. The
invariant it guarded was therefore unenforced in production.

It matters because **`agent_id` IS the agent's email address**. Engagements left
behind by a hard-deleted agent are inherited by a recreated agent at that
address — which then wakes up holding last campaign's `stage: "touch3"` and a
past-due schedule, and mails investors it never contacted.

The delete now happens inside `PurgeDeletedAgents`' own transaction, beside the
message purge, so outreach state and the agent die together. **Suppressions are
deliberately not touched** — consent has to survive deletion and recreation.
That asymmetry is exactly why this is an explicit delete rather than a cascade.

`TestPurgeDeletedAgentsRemovesEngagements` drives the **real purge path**, not
the helper. That distinction is the whole point: the previous test called the
function directly, so it could never notice that nothing else did. I verified it
fails when the delete is removed:

```
--- FAIL: TestPurgeDeletedAgentsRemovesEngagements
    1 engagement(s) survived the agent purge — a recreated agent would resume a dead campaign
```

It also asserts the suppression still stands afterwards, pinning both halves of
the asymmetry in one test.

## 2. The drift sweep had no scheduler

`ReconcileEngagementCounts` existed and was tested but nothing ran it, so drift
in the materialized outreach counters would never have been detected.

It now runs in the janitor's maintenance sweep, bounded at 500 rows per pass so
a consistency check cannot dominate a maintenance run. It reconciles every
account (empty `userID` reconciles globally; a specific one scopes to a tenant).

Corrections are logged **individually and loudly**, and counted as a metric.
Each one means an activity hook did not fire — that is a defect to chase, not
expected churn, and it should read that way in the logs.

The reconciler is optional on the `Janitor` (nil when contacts aren't wired), so
its absence can't break cleanup; `cmd/e2a/main.go` is the one place it becomes
live.

## Verification

Go **48/48 packages**. The purge test verified to fail without the fix.

## Still outstanding

- No unit-level guard on the `ToRecipients` behaviour fixed in #699 — the e2e
  there proves it today, and this sweep would now surface a recurrence as
  drift, but neither is a unit test.
- `FinalizeProviderAcceptedTx` (crash-window bridge) remains unhooked; the sweep
  is what covers it — which is now actually scheduled, so that mitigation is
  real rather than theoretical.
- `contact.due` (next slice).


jiashuoz:

PR #704, opened 2026-07-25
feat(contacts): emit contact.due to wake an agent whose outreach is due

jiashuoz:
**Stacked on #702** (→ #699 → #690 → #689 → #685). This closes the loop.

An agent that is *not running* now gets woken when an engagement's
`next_action_at` passes. Everything before this assumed a live agent polling;
this is the piece that makes the feature work for a session that has ended.

## Exactly-once per schedule

The claim is atomic — the store marks `notified_next_action_at` in the same
statement that selects the row, with `FOR UPDATE SKIP LOCKED`. A retried or
concurrent sweep cannot wake an agent twice for one schedule. Writing a new
`next_action_at` re-arms it.

A failed publish therefore costs at most one **missed** wake-up rather than a
duplicate. That asymmetry is deliberate: a duplicate event invites a duplicate
email to a real investor, and the drift sweep from #702 plus the pull query are
both there to recover a miss.

## Two fail-closed exclusions

Both are joins, not stored flags, so they cannot go stale against what the send
path actually enforces:

- **Suppressed addresses never fire.** A due-event is an invitation to send.
  Waking an agent to mail someone who unsubscribed is the single worst thing
  this feature could do. Covers both account-wide and agent-scoped suppression.
- **Trashed agents never fire.** Deleting an agent stops its outreach *at once*,
  not 30 days later when retention expires. Restoring brings it back — tested
  both directions.

## Delivery and payload

Goes through the standard webhook outbox, inheriting durable at-least-once
fan-out, retries, HMAC signing, and the SSRF guard rather than a bespoke path.
The event id is deterministic per `(agent, address, schedule)`, so an
at-least-once redelivery is dedupable by subscribers too.

The payload carries stage, schedule, reply state, and the contact's identity and
metadata — everything needed to compose without a second round trip — and
deliberately carries **no suggested action or body**. e2a wakes the agent; the
agent decides and writes. Putting content here would be the first step toward
e2a owning the sequence, which is the line this design does not cross.

## Naming

`contact.due`, not `agent.contact_due`. Scope prefixes don't scale as namespaces
in a per-agent product (nearly everything here is agent-scoped), and subscribers
already narrow by agent via the `agent_ids` webhook filter — which frees the
namespace to name the *subject*, as `email.*` and `domain.*` do.
`agent.suppression_added` is the exception, justified by its `domain.*` pair.

## Verification

Go **48/48 packages**; `spec-check` and SDK operation coverage pass. Four store
tests: fires once per schedule and re-arms; skips both suppression scopes; skips
trashed agents and resumes on restore; ignores future schedules.

The drift gate (`TestWebhookEventEnumMatchesCatalog`) failed until `contact.due`
was added to `AllEventTypes`, `ExperimentalEventTypes`, and all three
request-side enum copies — working as intended.

## Not verified here

No end-to-end run yet proving a real webhook subscriber receives `contact.due`.
Given that an e2e caught a production-breaking defect in three of the previous
four slices, I would want that before merge: arm an engagement in the past,
subscribe a webhook, run the sweep, assert the delivery arrives — and assert a
suppressed contact produces nothing.


jiashuoz:

PR #725, opened 2026-07-26
fix(oauth): include issuer in authorization errors

jiashuoz:
## Summary

- append the RFC 9207 `iss` parameter to redirecting OAuth authorization errors
- accept explicit `response_mode=query`, matching authorization-server discovery
- normalize unsupported fragment/form-post error responses to the server's sole advertised query mode
- preserve fosite as the authority for request validation, OAuth error parameters, and redirect matching
- fail closed when a non-DCR client row contains a dangerous redirect scheme
- reject malformed OAuth issuer/API URLs during configuration validation
- leave direct errors for unknown clients or untrusted redirect URIs unchanged

## Root cause

e2a's authorization-server metadata advertised two capabilities that the implementation honored only partially:

1. `authorization_response_iss_parameter_supported: true`, while only successful authorization responses included `iss`.
2. `response_modes_supported: ["query"]`, while clients that explicitly requested `response_mode=query` were rejected because the stored client type did not implement fosite's response-mode interface.

This PR aligns authorization behavior with discovery. It does **not** claim to resolve the separate Codex MCP login failure: the pinned Codex client ignores `iss` and does not send `response_mode`, so that issue remains open for request-level diagnosis.

## Verification

- RED: denied-consent and invalid-scope redirects omitted `iss`
- RED: explicit `response_mode=query` returned `unsupported_response_mode`
- RED: fragment split `iss` from `error`/`state`; form-post returned HTML without `iss`
- RED: a database-seeded `javascript:` redirect emitted an executable `Location`
- RED: malformed `http.api_url` values were accepted during startup validation
- focused OAuth/config regression and control tests passed
- isolated `E2A_TEST_DATABASE_URL=... make test` passed across all Go packages
- real binary startup rejected an invalid `/` API URL before service initialization
- real local service:
  - explicit query mode returned `302` to login with `response_mode=query` preserved
  - a database-seeded dangerous redirect returned `400` with no `Location`
  - unsupported fragment/form-post errors returned `303` query redirects containing `error`, `state`, and `iss`
  - unknown client remained a direct `401` without a redirect

### DIFF
--- a/cmd/e2a/main.go
+++ b/cmd/e2a/main.go
@@ -568,7 +568,7 @@ func main() {
 	} else {
 		// Issuer = api_url (defaults to public_url). fosite stamps it into
 		// token `iss` and the RFC 9207 response, so it must match what
-		// discovery advertises and what agentAuthIssuer signs/verifies.
+		// discovery advertises and what oauthIssuer signs/verifies.
 		oauthProvider, err := oauth.NewProvider(oauthStorage, cfg.HTTP.APIURL, []byte(cfg.Signing.HMACSecret))
 		if err != nil {
 			log.Fatalf("[oauth] provider wiring failed: %v", err)
--- a/internal/agent/agent_identity.go
+++ b/internal/agent/agent_identity.go
@@ -30,13 +30,6 @@ import (
 // falling through to fosite's authorization_code/refresh_token handling.
 const jwtBearerGrantType = "urn:ietf:params:oauth:grant-type:jwt-bearer"
 
-// agentAuthIssuer is the iss/aud bound into every minted token — the API base
-// URL (apiURL, which defaults to publicURL), trailing slash trimmed so it's
-// byte-stable with the discovery doc's issuer. It signs AND verifies, so a
-// deployment that changes api_url re-keys its token audience: tokens minted
-// under the old issuer stop validating and clients must re-auth.
-func (a *API) agentAuthIssuer() string { return strings.TrimRight(a.apiURL, "/") }
-
 // agentAuthReady reports whether the agent-identity surface is usable: a signing
 // key AND a public URL (needed for iss/aud) must both be configured.
 func (a *API) agentAuthReady() bool {
@@ -83,7 +76,7 @@ func (a *API) handleAgentIdentity(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	assertion, exp, err := a.signer.SignIdentityAssertion(ag.ID, identity.ScopeAgent, ag.AssertionVersion, a.agentAuthIssuer())
+	assertion, exp, err := a.signer.SignIdentityAssertion(ag.ID, identity.ScopeAgent, ag.AssertionVersion, a.oauthIssuer())
 	if err != nil {
 		writeOAuthError(w, r, http.StatusInternalServerError, "server_error", "failed to mint identity assertion")
 		return
@@ -120,7 +113,7 @@ func (a *API) handleJWTBearerGrant(w http.ResponseWriter, r *http.Request) {
 		writeOAuthError(w, r, http.StatusBadRequest, "invalid_request", "assertion is required")
 		return
 	}
-	claims, err := a.signer.VerifyToken(assertion, agentauth.TypIdentityAssertion, a.agentAuthIssuer())
+	claims, err := a.signer.VerifyToken(assertion, agentauth.TypIdentityAssertion, a.oauthIssuer())
 	if err != nil {
 		writeOAuthError(w, r, http.StatusBadRequest, "invalid_grant", "identity assertion is invalid or expired")
 		return
@@ -137,7 +130,7 @@ func (a *API) handleJWTBearerGrant(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	token, exp, err := a.signer.SignAccessToken(ag.ID, identity.ScopeAgent, ag.AssertionVersion, a.agentAuthIssuer())
+	token, exp, err := a.signer.SignAccessToken(ag.ID, identity.ScopeAgent, ag.AssertionVersion, a.oauthIssuer())
 	if err != nil {
 		writeOAuthError(w, r, http.StatusInternalServerError, "server_error", "failed to mint access token")
 		return
@@ -160,7 +153,7 @@ func (a *API) resolveAgentAccessToken(r *http.Request, bearer string) (*identity
 	if !a.agentAuthReady() || !looksLikeJWT(bearer) {
 		return nil, false, nil
 	}
-	claims, err := a.signer.VerifyToken(bearer, agentauth.TypAccessToken, a.agentAuthIssuer())
+	claims, err := a.signer.VerifyToken(bearer, agentauth.TypAccessToken, a.oauthIssuer())
 	if err != nil {
 		// It parses as a JWT but isn't a valid e2a access token — reject
 		// rather than fall through (a tampered/expired token is a 401, not an
--- a/internal/agent/oauth_consent_test.go
+++ b/internal/agent/oauth_consent_test.go
@@ -284,6 +284,39 @@ func TestHTTP_Authorize_InvalidClient(t *testing.T) {
 			t.Errorf("unknown client should NOT redirect to consent; Location=%q", loc)
 		}
 	}
+	if loc := resp.Header.Get("Location"); loc != "" {
+		t.Errorf("unknown client must not redirect to an untrusted redirect_uri; Location=%q", loc)
+	}
+}
+
+// TestHTTP_Authorize_InvalidScope_RedirectIncludesIssuer covers an authorize
+// error after fosite has verified the client and redirect_uri. RFC 9207
+// requires the redirect to identify the authorization server just like a
+// successful code response does.
+func TestHTTP_Authorize_InvalidScope_RedirectIncludesIssuer(t *testing.T) {
+	f := newConsentFixture(t)
+	_, challenge := newPKCE(t)
+	q := authorizeParams(challenge, f.clientID, "s1s1s1s1s1s1s1s1")
+	q.Set("scope", "agent unknown")
+
+	resp := f.authorizeRequest(t, q, true)
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusFound && resp.StatusCode != http.StatusSeeOther {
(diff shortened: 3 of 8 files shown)


jiashuoz (commented):
## Review: sound fix, approve with follow-ups

Reviewed against the code, the fosite v0.49.0 source, and by running the tests — including a RED check rather than taking the description's word for it.

**Verification**

- **RED confirmed.** Reverting `writeAuthorizeError` back to `oauthProvider.WriteAuthorizeError` fails all three new/updated assertions: `iss=""` on invalid-scope and on denied consent, fragment mode putting `error`/`state` in `loc.Fragment`, and form_post returning a 200 HTML body instead of a redirect.
- **GREEN confirmed.** `go test ./internal/agent ./internal/oauth -tags integration -count=1` passes (52s / 5s); `go build ./...` clean. The `go vet` noise in `internal/agent` is pre-existing across the whole test tree and unrelated.
- The issuer string is `strings.TrimRight(a.apiURL, "/")` — byte-identical to what `handleOAuthDiscovery` advertises (`oauth_handlers.go:727`) and to the success path (`oauth_handlers.go:1239`). That's what RFC 9207 clients actually compare, and it matches.
- The decorator is correctly scoped: fosite's `!ar.IsRedirectURIValid()` branch (`authorize_error.go:23`) never sets `Location`, so `appendIssuer` no-ops and unknown-client stays a direct JSON error. Forcing `ResponseModeQuery` is safe because `DefaultResponseModeHandler.ResponseModes()` returns nil, so the `.Has(...)` check at `authorize_error.go:17` is false and fosite falls into the plain query branch rather than delegating to an extension.
- All five previous `WriteAuthorizeError` call sites were converted; one raw call remains, inside the new helper. Neither `issueOAuthCode` nor `issueOAuthCodeWithNewAgent` can write-then-error, so the wrapper never lands on an already-committed response.

Nothing below blocks merge, but (1) is the one I'd want landed before calling the strict-client story closed.

---

### 1. `response_mode=query` — the only mode we advertise — is rejected (medium; pre-existing, but this PR is the place)

`internal/oauth.Client` is a minimal `fosite.Client` and does not implement `fosite.ResponseModeClient`, so fosite's `validateResponseMode` rejects **any** explicit response_mode, including the one `response_modes_supported: ["query"]` promises. Probed on this branch:

```
response_mode=query → 303 …/callback?error=unsupported_response_mode
  &error_description=…The request has response_mode 'query'. set but
  registered OAuth 2.0 client doesn't support response_mode
  &iss=https%3A%2F%2Ftest.e2a.dev&state=…
```

This is the same failure class the PR set out to fix — a strict client reading discovery and echoing what it says. Error delivery is now correct, but such a client still can't complete a *successful* authorize. The new table test covers `fragment` and `form_post` and conspicuously omits `query`.

Fix is small: add `GetResponseModes() []fosite.ResponseModeType` to `internal/oauth.Client` returning `{ResponseModeDefault, ResponseModeQuery}`, plus a `query` row in the table asserting a successful consent redirect rather than an error.

### 2. The error path skips the dangerous-scheme denylist the success path applies (low-medium; pre-existing, now cheap)

`writeAuthorizeRedirect` refuses to emit `Location: javascript:` / `data:` / `blob:` … as defense-in-depth (`oauth_handlers.go:1161`), explicitly because a client row could arrive from a non-DCR path. fosite applies no such check — `IsValidRedirectURI` only requires a URL shape and an empty fragment — so an error redirect has always been exempt.

`appendIssuer` is now the single choke point for every authorize-error `Location`, which makes mirroring those four scheme cases a ~4-line change. Worth closing the asymmetry while we're in here.

### 3. Empty issuer is emitted as `iss=` (low)

`appendIssuer` sets `iss` unconditionally. If `apiURL` were ever empty, a strict client gets `iss=` — worse than absent, since it compares byte-for-byte. Not reachable today (the provider is only wired when `public_url` is set, and `apiURL` defaults to `publicURL` at `api.go:534`), so this is a guard rather than a bug.

Same latent hole in `writeAuthorizeRedirect`. `strings.TrimRight(a.apiURL, "/")` is now written three times — `oauth_handlers.go:1118`, `oauth_handlers.go:1239`, and `agentAuthIssuer()` in `agent_identity.go:38` — so one `a.oauthIssuer()` helper with the empty check inside would cover all three.

### 4. The requester wrapper silently drops `fosite.G11NContext` (low, latent)

`GetLang` isn't part of `AuthorizeRequester`, so it isn't promoted through the embedded interface. Compile-checked: `*fosite.AuthorizeRequest` satisfies `G11NContext`; `queryModeAuthorizeRequester` does not. So `getLangFromRequester` at `authorize_error.go:22` falls back to English.

Zero impact today — no `MessageCatalog` is configured — but if localization is ever wired it will silently not apply to authorize errors, and nothing will fail. A `GetLang()` passthrough plus `var _ fosite.G11NContext = queryModeAuthorizeRequester{}` makes it non-silent.

### 5. Nits

- In `TestHTTP_Authorize_InvalidClient`, the new `Location != ""` assertion subsumes the older `if resp.StatusCode == http.StatusFound { … contains "/oauth/consent" }` block, which can no longer trip. Collapse to the stronger one.
- Four assertions hardcode `https://test.e2a.dev`. An `issuer` field on `consentFixture` would keep them from drifting from `newConsentFixture`.
- `authorize_error.go:17` means that if a `ResponseModeHandlerExtension` is ever configured and claims `query`, the forced mode will route into that extension instead of the default branch. Worth a sentence in the `queryModeAuthorizeRequester` doc comment, since current correctness depends on that config being nil.

---

No spec or SDK work needed — `/oauth2/*` lives on the legacy mux, not the Huma `/v1` surface, so `spec-check` and `generate-sdk-check` are unaffected.

jiashuoz:
## Follow-up: the stated root cause doesn't hold for Codex — the underlying bug is still open

Went and read Codex's MCP OAuth client rather than reasoning from the symptom. The fix in this PR is correct on its own terms, but it should **not** be expected to unblock Codex, and the linked issue shouldn't be closed on the strength of it.

References pinned to `openai/codex@61a4488`, `codex-rs/rmcp-client/src/perform_oauth_login.rs`.

### 1. Codex never reads `iss`

`parse_oauth_callback` (L330) matches exactly four keys and discards everything else:

```rust
match key {
    "code" => code = Some(decoded),
    "state" => state = Some(decoded),
    "error" => error = Some(decoded),
    "error_description" => error_description = Some(decoded),
    _ => {}          // ← iss lands here
}
```

And its own test asserts the no-`iss` error callback works cleanly:

```rust
#[test]
fn parse_oauth_callback_returns_provider_error() {
    let parsed = parse_oauth_callback(
        "/callback?error=invalid_scope&error_description=scope%20rejected",
        "/callback",
    );
    assert_eq!(parsed, CallbackOutcome::Error(OAuthProviderError::new(
        Some("invalid_scope".to_string()),
        Some("scope rejected".to_string()),
    )));
}
```

So a missing `iss` cannot have suppressed the underlying OAuth error — Codex would have surfaced `error=<whatever>` and its `error_description` already.

### 2. Codex never sends `response_mode`

`start_authorization` (L667) delegates to `rmcp`'s `OAuthState` / `AuthorizationManager`; the only parameter Codex appends itself is RFC 8707 `resource` (L552), and only when `oauth_resource` is configured. So the response-mode normalization doesn't apply to it either.

### 3. The success redirect was already shaped the way Codex needs

Codex's success branch requires **both** `code` and `state` (L360). fosite's authorize-code handler adds `state` to the response params (`handler/oauth2/flow_authorize_code_auth.go:91`), `writeAuthorizeRedirect` copies all of them, and `TestHTTP_Consent_Allow_CreateNew` already asserts `code` + `state` + `iss` on main. Not the cause either.

### Where I'd look instead

Structurally, the things that differ per-client and would produce "works for one MCP client, fails for another":

- **redirect_uri exact-match.** Codex derives the callback path from the redirect_uri and can bind a callback ID to the server URL (`callback_id_from_server_url` → `/callback/<id>`), on a loopback port. fosite compares registered `redirect_uris` by exact string, so a DCR-registration / redirect_uri mismatch fits the symptom far better than anything in this PR.
- **`resource`.** If `oauth_resource` is configured, the appended RFC 8707 param hits `validateAuthorizeAudience` → `invalid_target`.
- **Scope.** Codex sends whatever scopes it's configured with; those must be a subset of `{agent, account}` (the legacy `mcp` scope is retired).
- **Opaque-failure trap.** Any callback that is neither (`code`+`state`) nor (`error`) returns `CallbackOutcome::Invalid` (L368) — a generic "invalid callback" with no provider detail, which is a good match for a vague client-side symptom.

The definitive answer is one log line: the actual `/oauth2/authorize` query string from a Codex attempt plus the fosite error it produced. Worth grabbing before assuming this is fixed.

### Bearing on this PR

None of the above argues against merging — e2a advertises `authorization_response_iss_parameter_supported: true` and should honor it on error responses too, and funnelling every authorize error through one writer is a good change. Two asks:

1. Reword the **Root cause** section so it doesn't attribute the failure to a specific client's `iss` enforcement; "e2a advertised RFC 9207 but only honored it on success responses" is accurate and sufficient on its own.
2. Keep the underlying Codex issue open.

Also note that review finding #1 (explicit `response_mode=query` — the only advertised mode — being rejected with `unsupported_response_mode`) is still a real bug worth fixing; it just isn't Codex's, since Codex doesn't send the parameter.

jiashuoz:
## Codex troubleshooting update

Reproduced the separate MCP login failure against production after addressing the review comments.

- `codex-cli 0.145.0` sends a valid authorization request, receives the browser callback, then fails with:

  ```text
  Authorization server response missing required issuer: expected https://api.e2a.dev
  ```

- The current Codex callback parser still discards `iss`, but Codex now ships `rmcp 1.8.0`. That rmcp release requires the callback issuer whenever discovery advertises `authorization_response_iss_parameter_supported: true`. Codex then calls the issuer-less `handle_callback(&code, &state)`, so rmcp receives `None` even when the authorization response contained `iss`.
- The same unchanged e2a MCP configuration succeeds with `codex-cli 0.142.5`, the last unaffected release.
- This is tracked upstream as [openai/codex#31573](https://github.com/openai/codex/issues/31573).

So the conclusion above still stands: this PR is valid e2a RFC 9207/discovery conformance work, but it does not fix the Codex 0.143–0.145 callback-forwarding regression. The server-side workaround would be to stop advertising issuer support, which would weaken the protection and misrepresent e2a's behavior; the correct fix is for Codex to preserve and pass the callback `iss` value to rmcp.

PR #765, opened 2026-07-30
feat(threading): materialize mailbox-local email thread identity

jiashuoz:
## Summary
- materialize a server-owned, mailbox-local email `thread_id` separately from caller-owned `conversation_id`
- make API replies inherit the referenced message's email thread whether or not `conversation_id` is supplied
- resolve inbound RFC `Message-ID` / `In-Reply-To` / `References` topology through exact, bounded, mailbox-local lookups with fail-closed ambiguity handling
- lazily adopt an exactly referenced predeployment anchor without bulk reconstruction or historical backfill
- group the dashboard inbox by `thread_id`, with legacy fallbacks for null rows
- expose optional, response-only `thread_id` as a beta field on existing message list/detail SDK models and CLI JSON

## Contract
- `conversation_id` remains application/agent metadata; it never overrides email topology
- new cold sends mint new threads, including sends that reuse a `conversation_id`
- replies with or without `conversation_id` inherit the referenced message's thread
- inbound replies that exactly reference an inbound or outbound wire anchor inherit that mailbox's thread
- unmatched, malformed, conflicting, ambiguous, or cross-mailbox evidence mints a new thread
- old rows remain null unless one is directly referenced and safely lazy-adopted
- caller-supplied `thread_id` is rejected; there is no request field, filter, `/threads` endpoint, MCP tool, event/WebSocket field, export field, or new CLI command

## Storage and safety
- add nullable `thread_id`, `thread_parent_id`, and canonical RFC key columns with no backfill
- add concurrent partial indexes and validate `pg_index.indisvalid` / `indisready` before recording no-transaction migrations
- keep migrations additive and no-backfill; retire the unreleased branch-only grouping index with migration 093 so fresh databases do not build it and intermediate databases clean it up
- bound parsing to 64 KiB per field, 998 bytes per identifier, and the nearest 32 identifiers per field (64 total)
- bound each exact anchor branch and deduplicated candidate to `N+1 = 17`; lock at most 1,024 initially threadless anchors per recipient
- use the exact production query under a forced generic plan and require mailbox + message-ID equalities inside all three intended index conditions
- bound retention parent selection and child-pointer detachment to 5,000 rows per transaction

## Validation
- real over-the-wire local flow: inbound root → API reply without `conversation_id` → external SMTP reply all shared one thread
- verified explicit `conversation_id` does not change email topology; cold send with a reused conversation gets a new thread
- verified caller `thread_id` rejection and same RFC anchor isolation across two mailboxes
- `go test -p 1 -tags integration -count=1 ./internal/identity` — pass (164.945s, uncached, before the final focused remediation tests)
- targeted real-Postgres migration retry, no-backfill, forced-generic plan, retention, high-fanout, and lock-safety regressions — pass
- RFC parser, relay malformed-header, and Prometheus metric tests — pass
- MCP projection test: 137 tests + type check (sandboxed full MCP HTTP tests require socket permissions; CI covers the full suite)
- web: 687 tests, lint, build
- CLI: 256 tests + build
- TypeScript SDK: 222 tests, types, build
- Python SDK: 499 passed / 32 skipped + mypy
- OpenAPI compatibility, generated-code freshness, operation coverage, contract scenarios, and repository text integrity — pass locally before final remediation; CI reruns them on this head

## Review
- specification reviewed in multiple Claude Code and Kimi rounds before implementation
- implementation received independent, adversarial, and protocol/PostgreSQL review
- latest Claude Code review: approve with follow-ups, no blocking defects
- fixed: malformed-header observability, MCP projection drift guard, migration rollout note, and forward/idempotency behavior documentation
- fixed: migration 086 is now a retained no-op and migration 093 drops its unused branch-only index, preserving migration history without ongoing write/storage overhead
- deferred: a `thread_id` filter/complete-thread endpoint is a separate product/API decision; cross-runner topology assertions remain covered by Go integration and wire e2e tests rather than expanding the contract DSL in this PR

## Rollout
- additive for new messages after deployment
- no old-message bulk backfill or global thread reconstruction
- no deployment or production data change is part of this PR
- first startup may spend time building the direction-aware concurrent indexes over existing rows; schedule normal migration headroom and monitor startup logs

One accepted low-priority UI follow-up remains: a previously loaded paginated legacy null parent can appear separately from a newly lazy-adopted child until reload. It self-heals on reload and does not affect persisted topology or mailbox isolation.


jiashuoz (commented):
# Multi-agent review — docs: specify mailbox-local email thread identity

Four parallel reviewers: ground-truth accuracy vs. code, design/RFC correctness, API surface, repo conventions. All read the full 1,114-line spec against `main`.

## Result: **pass with risks**

No blockers. The spec's claims about the current system are remarkably accurate — the ground-truth reviewer verified ~20 claims against the code (conversation resolution order in `internal/agent/api.go:1239`, wildcard provider-ID lookup in `internal/identity/store.go:4485`, the `hitlworker` forward trap at `internal/hitlworker/worker.go:642`, the SES Message-ID commit `c618738d`, all 14 cited file paths) and found zero factual errors. Migration plan matches the repo's real `-- e2a:no-transaction` + `CONCURRENTLY` conventions. The API shape would pass the oasdiff compat gate. What remains is spec-completeness debt — places where an implementer would have to invent contract-frozen choices.

## Findings (ordered by severity)

### Should-fix

1. **Outbound wire-anchor persistence race → permanent thread splits** (doc lines 319-324, 429-441). The doc allows `rfc_message_id_key` to be written only at provider acceptance, but an instant agent-to-agent reply can hit the inbound relay in that window, find no anchor, and mint a new thread — permanently, per the no-merge invariant. Remedy: e2a composes the Message-ID itself, so the canonical key is knowable at queue time — persist it at outbound row creation, or have inbound resolution consult pending outbound rows. Add a scenario-matrix row + e2e test.

2. **Root `CONTEXT.md` is orphaned.** Nothing references it (`git grep` → zero hits), it's not in any discovery path (`CLAUDE.md`→`AGENTS.md`, `sync-agent-docs.mjs` only mirrors `plugins/e2a/docs/`), and the design doc silently *depends* on it — "delivery twin", "application conversation", "agent mailbox" are defined only there, yet neither file links the other. Remedy: move to `docs/design/2026-07-30-email-thread-identity-glossary.md` and link it from the spec's Terminology section, or add a pointer from `AGENTS.md`.

3. **`thread_id` absent-vs-explicit-null not pinned** (lines 575-587, 754). The repo has both conventions live (`conversation_id` is omitempty; `envelope_from` is `type: [string,"null"]`). Once generated SDKs ship, switching is a breaking change the gate rejects. Remedy: pick one (recommend omitempty, matching `conversation_id`) and state it for REST, event payloads, and WS.

4. **Parser spec silent on RFC 5322 obsolete msg-id forms** (lines 274-287). §4.5.4 `obs-id-left` (quoted-string left portions) MUST be parsed by conformant parsers and appears in old References chains; the doc neither accepts nor rejects them, so two implementers will diverge and strict rejection silently splits legacy threads. Remedy: explicit rule — accept obs forms, preserve byte-for-byte, never generate — plus unit-test entries.

5. **Thread views mix trust levels without per-message verdicts** (lines 825-833, 670-688). A spoof-joined message renders inline beside DKIM-verified traffic in `ThreadDetailView`, and e2a's consumers are LLM agents that will read "in the thread" as a soft trust signal — for a product with piguard in its threat model. Remedy: include compact per-message auth/screening fields in thread-detail entries, or document explicitly that thread membership is not a trust signal.

6. **Frozen-at-first-commit API metadata unspecified** (lines 592-694): `operationId`, operation tag (determines generated SDK module grouping), and `x-stability-level` — neighbors diverge (`forwardMessage` is beta, `replyToMessage` stable), so "copy the neighbor" isn't a coherent default. Also: held-outbound treatment in thread aggregates is unspecified (lines 630-634) while `/messages` today shows held outbound; and the export `schema_version` bump consequence (line 583, currently pinned at `"4"` in `internal/identity/user_data_rights.go:27`) is ignored.

7. **Verification plan misses mandated harnesses** (lines 978-986): no mention of `tests/contract/scenarios.yaml` three-runner parity (required by AGENTS.md for every new `/v1` feature), and no mention of regenerating the eventpayload golden fixtures + `docs/events.md` field table for the webhook additions.

### Nits (selection)

- Duplicate same-Message-ID redelivery (alias loops, crash-before-ACK) splits threads — no matrix row (lines 803-804).
- Cycle audit's own traversal needs a visited-set/depth cap (lines 835-841).
- CONTEXT.md "Twins share a thread" contradicts the conditional platform-twin path (lines 399-418).
- "Anchor" vocabulary is load-bearing but never defined in Terminology.
- `payloads.go:71`'s comment already uses "thread" to mean *conversation* — sweep with the CLI/MCP help-text renames (CLI says `--conversation-id <id> Thread id`).
- PR body's `check-repository-text-integrity.sh` claim is near-vacuous for markdown (it greps conflict markers, doesn't lint docs).

## Verification summary

Independent of the PR's own claims: ground-truth reviewer read ~15 backend/dashboard files and found every checkable current-behavior claim correct (2 wording nits). RFC reviewer verified the msg-id grammar section against RFC 5322 full text (correct for modern forms; silent on obs forms). API reviewer checked the proposed shapes field-by-field against `api/openapi.yaml` and the compat-gate policy. Conventions reviewer ran `check-repository-text-integrity.sh` (passes) and byte-verified the reviewed files match PR head `d61eeabe`.

## Operational / composition risks

This PR is spec-only, so nothing deploys. The real risk is downstream: finding #1 (anchor race) is the one issue that would produce *wrong production behavior* if implemented as written, and #3/#6 are choices that become unchangeable the moment Phase-3 API code lands. All are cheap to fix now, expensive later.

## Recommended next slice

Revise the doc in place: fix the anchor-persistence race (§Assignment/§Data model), pin the null-vs-absent rule + operationIds/tag/stability, move/link the glossary, and add the missing scenario-matrix rows. Then it's ready to approve.

---

🤖 Generated with [Kimi Code](https://www.kimi.com/code) — 4-agent parallel review (ground-truth, design/RFC, API surface, conventions)

jiashuoz:
## Review: approve, with follow-ups — no blocking defects

The core design holds up under scrutiny. Things I specifically tried to break and could not:

- **Mailbox isolation** — every branch of `threadAnchorBatchQuery` carries `m.agent_id = $1`; the twin path additionally requires SPF pass + envelope-domain alignment + `message_type = 'test'` + own-message-ID equality before it will copy a thread.
- **Fail-closed ambiguity** — the `N+1 = 17` per-branch cap plus the post-union dedup cap means an attacker-controlled duplicate set is never materialized, and overflow degrades to an older `References` candidate rather than a merge.
- **No leakage** — `Message.ThreadID` / `ThreadParentID` / `RFCMessageIDKey` are `json:"-"`, so `ExportUserData` (which serializes `[]Message` directly) cannot acquire them; the contract scenarios pin absence on reviews, send results, and event payloads.
- **`postCommitTx`** savepoint semantics are correct: a nested commit transfers callbacks to the parent, rollback discards them, and only the outermost commit fires them.
- **Retention** — the detach-then-delete ordering cannot leave a dangling `thread_parent_id`, and the loop terminates (each round either detaches a pointer that cannot be re-detached, or deletes a row).

Verified locally against a real Postgres: `go build ./...`; `internal/{identity,relay,agent,httpapi,janitor,hitlworker}` with `-tags integration` (all pass); `make spec-check`. Also wrote a throwaway fuzz target for `ParseTokens` / `Canonicalize` — **17.7M executions, no panic, token/identifier bounds held, canonicalization idempotent**.

---

### Findings

**1. Migration 086 builds an index nothing reads** _(medium)_

`(agent_id, thread_id, created_at, id) WHERE thread_id IS NOT NULL` — grepping every SQL string in the tree, nothing filters or orders by `thread_id`. That is a `CONCURRENTLY` build plus permanent write amplification on `messages`, the hottest table, for a reader that does not exist. Drop it until there is a consumer, or land the consumer (see #2).

**2. `thread_id` ships with no way to use it** _(medium, product)_

`docs/api.md` states plainly that there is no filter, no `/threads` resource, and no complete-thread retrieval guarantee. So the only real consumer is the dashboard, grouping client-side over whatever page happens to be loaded — which is exactly why the accepted UI follow-up (legacy parent appearing separately until reload) exists. Adding `GET …/messages?thread_id=` is additive, cheap, and index 086 is already purpose-built for it. Otherwise this carries a beta compatibility surface that no API caller can act on.

**3. All-or-nothing header parsing can fragment threads invisibly** _(medium)_

`relay.parseThreadCandidates` discards the *entire* `In-Reply-To` / `References` field on any parse error, and one bad byte is an error — a stray `<>`, an unbracketed id, a non-ASCII/EAI msg-id. That is a correct fail-closed choice, but the metric vocabulary has `legacy_anchor_unmatched` and `ambiguous_anchor` and nothing for "field would not parse", so it lands as plain `no_anchor`. A fragmentation spike caused by parser strictness would be indistinguishable from ordinary cold sends. A `malformed_reference_header` source is the first metric anyone will want when a user reports "my replies stopped threading".

**4. Migration cost at startup deserves an ops note** _(medium)_

086/087/088/090 have predicates on the new all-NULL columns, so they build empty. But **089 and 092 index every historical inbound/outbound row** — their predicates are on existing data. Six `CONCURRENTLY` builds run under the advisory lock at binary startup, before the process serves. The PR body's "no deployment change is part of this PR" undersells that. Either pre-create out of band (the runner is idempotent and now verifies `indisvalid`) or budget for a long first startup in the rollout plan.

**5. MCP `list_messages` projection has no drift guard** _(low)_

Replacing the generic camel→snake conversion with an explicit allowlist is the right call for keeping beta fields out of MCP output. But a *new* `MessageSummaryView` field will now silently vanish from MCP output instead of automatically appearing. A test asserting the projection covers every non-beta field of the SDK model would lock it.

**6 & 7. Two unadvertised behavior changes** _(low)_

Both are correct fixes; both belong in the PR body / release notes rather than buried in a threading commit:

- `hitlworker.sendRequestFromStoredMessage` now gates `ReplyToMessageID` on `type == "reply"`, bringing the River worker into parity with the guard `agent.buildSendRequestFromMessage` already had. Approved **forwards stop emitting `In-Reply-To` / `References`** — a visible outbound header change.
- `HoldForApprovalCoreThreaded` now completes the idempotency key inside the hold transaction, so a crash after commit replays instead of creating a second held draft.

**8. Migration 091 is now a gap** _(low)_

`091_drop_messages_thread_parent_idx.sql` was added in `77977e61` and deleted in `94e49086` when 088 was retained as a fallback. Fine for fresh databases — but anyone who ran an intermediate commit of this branch has 091 recorded in `schema_migrations` and will get a `warnOrphans` WARN forever. Worth confirming no staging database is in that state.

**9. `validateConcurrentIndexPostcondition` mis-parses an unnamed index** _(low)_

`CREATE INDEX CONCURRENTLY ON messages (...)` makes `createConcurrentIndexTarget` capture `on`, producing `concurrent index on is missing or invalid`. Loud rather than silent, so not a bug today — but a future migration author will lose an hour to it.

**10. Contract scenarios prove the field, not the grouping** _(low)_

The added scenarios pin `thread_id` presence and absence correctly. But the behavior that justifies the feature — a reply inherits the referenced message's thread; a cold send reusing a `conversation_id` gets a new one — is proven only in Go integration tests. `AGENTS.md` asks for cross-runner proof of new `/v1` behavior, and both assertions are expressible with the scenario language already in use.

**Nits**

- Commit the fuzz target (~20 lines). It is a hand-rolled byte parser on attacker-controlled input.
- `local_delivery.go` mints and discards a thread ID via `freshInboundMessageThread(...).rfcMessageIDKey` where `canonicalRFCMessageIDKey(...)` is the direct call.
- Web's `m.thread_id != null` would collapse every row into one bucket if the server ever emitted `""`. `omitempty` prevents that today, but `m.thread_id ? … : …` costs nothing.

---

<sub>Automated review by Jace (Claude Code), posted from the repo owner's account.</sub>

PR #769, opened 2026-07-31
fix(contacts): safe import reversal, strict import validation, RFC 3339 timestamp parity

jiashuoz:
## Contacts correctness: safe import reversal, import validation, timestamp parity

Three focused fixes across the contacts/outreach surface. All are **beta**-surface changes; the public shapes are unchanged except that `ImportContactsRequest.contacts` is no longer nullable (see Compatibility).

### A. Safe contact-import reversal

`DELETE /v1/contacts/imports/{batch_id}` previously deleted every engagement tagged with the batch and every tagged contact without message history. A contact PATCH did not protect the contact, an engagement edit did not protect the engagement, and contact deletion relied on the `contacts → contact_engagements` `ON DELETE CASCADE`, which silently destroyed engagements created *independently* of the import.

**Provenance semantics (now documented in the code and the design doc §3.3.1):** `import_batch_id` is *origin provenance* — it records which batch created the row and never moves (a merge re-import keeps it pointing at the original batch). Reversal is **defensive, not tag-driven**: a tagged row is removed only when verifiably untouched —

- `updated_at = created_at` (no PATCH/upsert/activity-record/due-notification ever landed);
- engagements: plus no derived wire activity and no message history for that agent/address;
- contacts: plus no message history (sender **or To/Cc/Bcc**) and **no surviving engagement** — evaluated after the batch's own untouched enrolments are deleted, so an independent engagement retains its contact and the cascade is never reached with live state;
- all batch-tagged rows are locked `FOR UPDATE` (contacts first, matching writer lock order) before the guards are evaluated, closing the statement-snapshot race where a concurrent PATCH or enrolment could be deleted on a stale verdict (EPQ re-checks only the join qual).

Receipt semantics: `contacts_deleted + contacts_retained` accounts for every batch-created contact that still exists at reversal time. Tenant scoping by `user_id` is unchanged.

### B. Contact import request validation

`contacts` was `required` + `minItems: 1` but typed `["array", "null"]`, so `contacts: null` validated and produced a **successful zero-row import**. Now:

- the schema is non-nullable (`type: array`), so missing / `null` / `[]` all fail validation with the canonical `invalid_request` envelope (422);
- a runtime `len == 0` guard returns 400 `invalid_request` as defense in depth;
- the 1000-row cap behavior is unchanged;
- `api/openapi.yaml` regenerated; TS + Python generated bases regenerated (`contacts` is now a non-optional array in both).

### C. Contact timestamp parity (MCP + CLI)

- **MCP** (`mcp/src/tools/contacts.ts`): all contact timestamp fields (`created_after/before`, `next_action_before`, `last_outbound_before`, `next_action_at`) now share one `z.string().datetime({ offset: true })` schema, matching `messages.ts`/`apikeys.ts` — `Z`, `-07:00`, and `+05:30` are all accepted; date-only and offsetless values are rejected.
- **CLI**: the permissive `new Date(raw)` parsing in `commands/contacts.ts` is replaced by the strict RFC 3339 parser extracted from `--send-at` into `cli/src/time.ts`. An explicit UTC offset is required everywhere, and impossible dates (`2026-02-30`, `24:00`, `+24:00`) are rejected by component round-trip — modern V8 rolls those over instead of returning `NaN`, so the NaN backstop never fired.

## Compatibility

Tightening `contacts` to non-nullable is a request-side strictness change on a **beta** endpoint. `make openapi-compat-check` reports no breaking changes. A client sending `contacts: null` previously got a useless empty success; it now gets 422 `invalid_request`.

## Testing

New regression coverage (all three SDK contract runners + store/handler/unit):

- reversal: untouched contact reversed; PATCHed contact retained; edited engagement retained; later independent engagement survives (cascade regression); wire activity (outbound **and** recorded inbound) retains engagement + contact; cc-only history protects; `notified_next_action_at` / `last_conversation_id` predicates isolated; exact receipt counts on a mixed batch; stale re-pointed provenance cannot delete; two blocked-lock interleavings (concurrent PATCH, concurrent independent enrolment) pinned; tenant isolation (pre-existing test still green);
- validation: missing / null / `[]` / valid at the handler, plus all three rejected shapes in the shared contract scenario;
- timestamps: positive (`Z`, `-07:00`, `+05:30`, optional seconds, fractional) and negative (date-only, offsetless, impossible dates/offsets, non-leap-year Feb 29) at MCP and CLI levels; `--send-at` behavior unchanged.

Verification run locally (Postgres 5433): `make test-unit`; full `internal/identity` + `internal/httpapi` integration runs; Go contract runner; TS contract (18 passed) and Python contract (19 passed, CI flags) against a live `e2a-contract-server`; TS SDK 221; CLI 262; MCP 289; Python 490 unit + mypy; `make spec-check`; `make openapi-compat-check`; coverage floors pass (78.3% total). Over-the-wire e2e against the real server binary: import → PATCH → reverse receipt 1/1/0, edited contact survives, untouched 404s, second reversal 404s; null/missing/empty → 422 `invalid_request`; CLI exit 2 on ambiguous timestamps, exit 0 on explicit offsets; server log clean.

**One environmental caveat:** OrbStack was removed from this machine mid-session, so `make generate-sdk-check` (Docker-pinned generator) could not be re-run after the final doc-string wording change. The committed generated trees came from the pinned generator earlier in the session; the final wording delta was applied as a verbatim string substitution to the generated docstrings (exactly what the generator copies from the spec), and `TestSpecGoldenNoDrift` passes. CI's generated-code freshness job is the authoritative re-check.

## Reviews

Independent + adversarial review passes were run on the branch. Findings and dispositions:

| Finding | Severity | Disposition |
|---|---|---|
| Statement-snapshot race: concurrent PATCH/enrolment could be deleted on a stale guard verdict (probe-proven by both reviewers) | blocker | **Fixed** — `FOR UPDATE` lock-first ordering; two interleaving regression tests |
| Cc/Bcc-only correspondence not counted as message history | should-fix | **Fixed** — guards include `messages.cc`/`messages.bcc` with NULL-safe concat; regression test |
| Retention predicates `last_inbound_at`, `last_conversation_id`, `notified_next_action_at` untested | should-fix | **Fixed** — predicate-isolation + recorded-inbound tests |
| No shared contract scenario for the changed reversal behavior | should-fix | **Fixed** — `contacts_import` scenario now proves PATCH-protection, reversal, and counts in all three runners |
| CLI accepted impossible dates (V8 roll-over; MCP/zod rejects) — parity gap | should-fix | **Fixed** — component round-trip validation in `parseRfc3339`; tests |
| Receipt "reconciles against what the batch created" over-claims when contacts were manually deleted pre-reversal | nit | **Fixed** — wording corrected in code, spec, SDKs, design doc |
| No `engagements_retained` in the reversal receipt | nit | **Deferred** — additive response field on a beta surface; not required to close the data-loss hole; can ship separately if wanted |
| MCP contact timestamp schema had no offset-requirement description | nit | **Fixed** |
| CLI help still said `<ISO>` for now-strict flags | nit | **Fixed** — `<rfc3339>` |

Adversarial classes explicitly refuted with probes: case/canonicalization mismatch, tenant isolation, timestamp-witness forgery, NULL/trashed message handling, import body-shape games, idempotency replay interference, zod dangerous-accepts, exit-code contract drift.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -6923,7 +6923,7 @@ paths:
       x-stability-level: beta
   /v1/contacts/imports/{batch_id}:
     delete:
-      description: "Reverses the durable import batch. Requires ?confirm=DELETE. It removes only what is verifiably untouched: contacts the batch created that have not been edited, enrolled in surviving outreach, or corresponded with since, and per-agent enrolments the batch created that carry no later edit, message, or recorded activity. Pre-existing outreach and suppressions are never affected, and a contact with any surviving engagement is always retained. The response reports each category, and contacts_deleted + contacts_retained reconciles against what the batch created. Account-scoped credentials only. Beta: the contact import surface may change before it is declared stable."
+      description: "Reverses the durable import batch. Requires ?confirm=DELETE. It removes only what is verifiably untouched: contacts the batch created that have not been edited, enrolled in surviving outreach, or corresponded with since, and per-agent enrolments the batch created that carry no later edit, message, or recorded activity. Pre-existing outreach and suppressions are never affected, and a contact with any surviving engagement is always retained. The response reports each category; contacts_deleted + contacts_retained accounts for every batch-created contact that still exists. Account-scoped credentials only. Beta: the contact import surface may change before it is declared stable."
       operationId: deleteImportBatch
       parameters:
         - in: path
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -1130,13 +1130,13 @@ components:
           format: int64
           type: integer
         contacts_retained:
-          description: How many contacts from this batch were deliberately kept because they have correspondence history.
+          description: "How many contacts from this batch were deliberately kept: edited since the import, enrolled in outreach that survives, or carrying correspondence history."
           format: int64
           type: integer
         deleted:
           type: boolean
         engagements_deleted:
-          description: How many per-agent outreach enrolments created by this import were removed.
+          description: How many per-agent outreach enrolments created by this import were removed. Enrolments edited or used since the import survive and are not counted here.
           format: int64
           type: integer
       required:
@@ -2284,9 +2284,7 @@ components:
             $ref: "#/components/schemas/ContactImportRow"
           maxItems: 1000
           minItems: 1
-          type:
-            - array
-            - "null"
+          type: array
         on_conflict:
           default: merge
           description: What to do when the address already exists. merge (default) refreshes display_name and metadata and leaves provenance and any state hanging off the contact untouched — so re-uploading a corrected spreadsheet is safe. skip leaves the existing contact completely alone.
@@ -6925,7 +6923,7 @@ paths:
       x-stability-level: beta
   /v1/contacts/imports/{batch_id}:
     delete:
-      description: "Reverses the durable import batch. Requires ?confirm=DELETE. It removes untouched contacts created by the batch and per-agent enrolments the batch created, including enrolments on pre-existing contacts. Contacts with correspondence history are retained; pre-existing outreach and suppressions are never affected. The response reports each category. Account-scoped credentials only. Beta: the contact import surface may change before it is declared stable."
+      description: "Reverses the durable import batch. Requires ?confirm=DELETE. It removes only what is verifiably untouched: contacts the batch created that have not been edited, enrolled in surviving outreach, or corresponded with since, and per-agent enrolments the batch created that carry no later edit, message, or recorded activity. Pre-existing outreach and suppressions are never affected, and a contact with any surviving engagement is always retained. The response reports each category, and contacts_deleted + contacts_retained reconciles against what the batch created. Account-scoped credentials only. Beta: the contact import surface may change before it is declared stable."
       operationId: deleteImportBatch
       parameters:
         - in: path
--- a/cli/src/__tests__/contacts.test.ts
+++ b/cli/src/__tests__/contacts.test.ts
@@ -259,4 +259,63 @@ describe("contacts commands", () => {
     await expect(outreachList({ nextActionBefore: "not-a-date" })).rejects.toThrow("process.exit");
     await expect(contactsList({ limit: "0" })).rejects.toThrow("process.exit");
   });
+
+  it("accepts explicit RFC 3339 offsets on every contact time argument", async () => {
+    mockList.mockReturnValue({ toArray: vi.fn(async () => []) });
+    mockOutreach.mockReturnValue({ toArray: vi.fn(async () => []) });
+    mockSetOutreach.mockResolvedValue({ address: "partner@fund.vc" });
+    const { contactsList, outreachList, outreachSet } = await import("../commands/contacts.js");
+
+    // Z, a negative offset, and a positive half-hour offset all name the same
+    // kind of unambiguous instant and must reach the SDK as that exact Date.
+    await contactsList({
+      createdAfter: "2026-07-01T09:00:00-07:00",
+      createdBefore: "2026-08-01T12:00:00+05:30",
+    });
+    expect(mockList).toHaveBeenCalledWith({
+      source: undefined,
+      importBatchId: undefined,
+      createdAfter: new Date("2026-07-01T09:00:00-07:00"),
+      createdBefore: new Date("2026-08-01T12:00:00+05:30"),
+    });
+
+    await outreachList({
+      nextActionBefore: "2026-08-01T09:00:00-07:00",
+      lastOutboundBefore: "2026-07-01T12:00:00+05:30",
+    });
+    expect(mockOutreach).toHaveBeenCalledWith("bot@agents.e2a.dev", {
+      stage: undefined,
+      replied: undefined,
+      suppressed: undefined,
+      nextActionBefore: new Date("2026-08-01T09:00:00-07:00"),
+      lastOutboundBefore: new Date("2026-07-01T12:00:00+05:30"),
+    });
+
+    await outreachSet("partner@fund.vc", { nextAction: "2026-08-01T09:00:00+05:30" });
+    expect(mockSetOutreach).toHaveBeenCalledWith(
+      "bot@agents.e2a.dev",
+      "partner@fund.vc",
+      { stage: undefined, metadata: undefined, nextActionAt: new Date("2026-08-01T09:00:00+05:30") },
+    );
+  });
+
+  it("rejects date-only and offsetless contact timestamps", async () => {
+    const { contactsList, outreachList, outreachSet } = await import("../commands/contacts.js");
+    // Without an explicit offset the instant is ambiguous: the JS Date
+    // constructor reads a bare date-time in LOCAL time and a date-only value
+    // as UTC midnight, so the filter would shift with the runner's timezone.
+    for (const stamp of ["2026-08-01", "2026-08-01T09:00:00"]) {
+      await expect(contactsList({ createdAfter: stamp })).rejects.toThrow("process.exit");
+      await expect(contactsList({ createdBefore: stamp })).rejects.toThrow("process.exit");
+      await expect(outreachList({ nextActionBefore: stamp })).rejects.toThrow("process.exit");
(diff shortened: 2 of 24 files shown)


jiashuoz:
## Review

**Verdict: approve after one fix.** The core design is sound and well tested; there is one concrete defect that will fail CI and ship a typo to SDK users.

### Verification run (worktree on `50d92e75`, Postgres :5433)

| Check | Result |
|---|---|
| `go test -tags integration ./internal/identity ./internal/httpapi` | pass (141s / 5s) |
| `make test-unit` | pass, no failures |
| `go vet`, `gofmt -l` on changed files | clean |
| `make spec-check` | pass |
| `make openapi-compat-check` vs `origin/main` | "No breaking changes to report" |
| CLI vitest | 262 pass |
| MCP vitest | 289 pass |

`make generate-sdk-check` could not be run here (no Docker on this machine) — which is exactly where the one real problem is.

---

### 1. Blocker — malformed generated docstrings (`, ;`)

The hand-applied string substitution flagged in the PR description did go wrong. `api/openapi.yaml` says:

> …The response reports each category**;** contacts_deleted + contacts_retained…

but the committed generated trees say:

> …The response reports each category**, ;** contacts_deleted + contacts_retained…

10 occurrences across 5 files:

- `sdks/python/src/e2a/v1/generated/api/contacts_api.py` (×3)
- `sdks/typescript/src/v1/generated/apis/ContactsApi.ts`
- `sdks/typescript/src/v1/generated/types/PromiseAPI.ts`
- `sdks/typescript/src/v1/generated/types/ObservableAPI.ts`
- `sdks/typescript/src/v1/generated/types/ObjectParamAPI.ts`

This is not generator behavior — other generated descriptions carry `; ` through verbatim (e.g. `ContactsApi.ts` siblings in `AccountApi.ts:28`, "Mint a new API key; the plaintext key…"). So the `generated-code freshness` CI job will see a non-empty diff and fail, and until then the SDKs ship a visible typo. Fix is a straight `category, ;` → `category;` across those five files, then reconfirm with the pinned generator once Docker is back.

### 2. Should-fix — the contract scenario is no longer self-cleaning on failure

`tests/contract/scenarios.yaml`: cleanup now deletes only `second@imp.vc`, because `first@imp.vc` is expected to be removed by the `reverse_batch` step. If anything from `edit_protects_from_reversal` onward fails, `first@imp.vc` survives, and the *next* run's `created: 2` assertion becomes `created: 1 / updated: 1` — the scenario wedges permanently against a persistent contract server. AGENTS.md explicitly asks for self-cleaning shared scenarios. Low impact if CI always launches a fresh `e2a-contract-server` DB; real for local/staging reuse.

### 3. Should-fix — `waitForBlockedLock` isn't database-scoped

`internal/identity/contacts_reversal_test.go` polls `SELECT count(*) FROM pg_locks WHERE NOT granted` server-wide. With per-package test DBs on one server under `make test -p 4`, an unrelated blocked lock in another package satisfies the wait early; the racing commit then lands *before* the reversal reaches its lock point, and the test still passes — so the interleaving it guards can silently stop being exercised. Join `pg_stat_activity` on `datname = current_database()`, or filter to the reversal connection's pid.

### 4. Worth a look — reversal is now O(batch × agent messages)

The engagement delete was previously an index-only `WHERE user_id AND import_batch_id`. It now carries a correlated `NOT EXISTS` over `messages` with `unnest(to_recipients || cc || bcc)`, so a 1000-row batch against a busy agent mailbox does 1000 × (all that agent's messages). The contacts CTE already had this shape, so it isn't a new class — but this endpoint is synchronous and the new cost sits on the hot side of it. Worth an `EXPLAIN` against a realistically sized `messages` table before GA.

### 5. Nits

- **The batch record is consumed even when nothing was removed.** With retention now much stricter, "reverse too early → everything retained → batch row deleted → 404 forever" is far more reachable than before. Pre-existing code, newly reachable behavior; a doc line, or keeping the row when `retained > 0`, would help.
- **A racing enrolment can surface as a 500.** If the reversal wins the `FOR UPDATE` on the contact and deletes it, a blocked `UpsertEngagement` insert then fails with FK violation 23503; there is no `ForeignKeyViolation` mapping in `internal/identity` or `internal/httpapi`, so that caller gets a 500 rather than a mapped error. Very narrow race.
- **Lock-coverage gap in the race test.** `TestDeleteImportBatchWaitsForConcurrentEngagement` stages the enrolment with an explicit `UPDATE contacts` (mirroring `UpsertEngagement`'s `ON CONFLICT DO UPDATE`). The skip-mode import path (`ON CONFLICT DO NOTHING` + unlocked `SELECT id`) never takes that lock and relies purely on the FK's `KEY SHARE` at engagement insert — that variant is uncovered. It does still block correctly.

---

### Checked and confirmed correct

- **The locking is genuinely sufficient**, and for a subtler reason than the comment states: `UpsertEngagement` takes its contact lock via `ON CONFLICT … DO UPDATE SET updated_at = contacts.updated_at` — a real row lock that preserves `updated_at`, so it does not false-positive the untouched guard — and the skip path is covered by the FK's `KEY SHARE`, which conflicts with `FOR UPDATE`. Lock order (contacts → engagements) matches both writers; no deadlock cycle.
- **Guard predicates are NULL-safe against the real schema**: `last_conversation_id TEXT NOT NULL DEFAULT ''` (migration 081), so `= ''` cannot silently evaluate to NULL; `contacts.import_batch_id` has no FK to `contact_import_batches`, so deleting the batch row does not wipe provenance on retained contacts.
- **Both lock queries are index-backed** (`contacts_batch_idx`, `contact_engagements_import_batch_idx`).
- **The MCP/CLI parity claim holds.** Verified that zod 4.4.3's `.datetime({ offset: true })` does reject `2026-02-30T09:00:00Z`, `2027-02-29T09:00:00Z`, and `+24:00` — so the CLI's new component round-trip genuinely matches the MCP rule rather than overshooting it.
- `--next-action clear` is handled before `dateValue`, so the strict parser does not break the clear sentinel.

jiashuoz:
Thanks — all findings addressed or triaged in `d8bc6e8c`:

1. **Blocker (`, ;` docstrings)** — confirmed (10 occurrences, 5 files) and fixed; the generated trees now byte-match the spec wording (`category; contacts_deleted…`). The `, ` prefix came from my substitution replacing `and …` with `; …` and leaving the preceding comma behind. Still needs the Docker-pinned `generate-sdk-check` re-run (no Docker locally — OrbStack was removed mid-session); CI's freshness job is the authoritative check.
2. **Self-cleaning scenario** — fixed: a best-effort (expect-less) cleanup step now deletes `first@imp.vc` when a mid-scenario failure left it behind. All three runners treat a step without `expect.status` as status-agnostic, so a passing run's 404 there is accepted.
3. **`waitForBlockedLock` scoping** — fixed: joins `pg_stat_activity` filtered to `current_database()`, so a blocked lock from another package's derived DB under `-p 4` can't satisfy the wait early.
4. **O(batch × agent messages)** — checked: the probes are backed by `idx_messages_agent_created (agent_id, created_at DESC)`, so each probe is an index scan over one agent's mailbox, not the table; the pre-existing contact guard already had this shape. Agreed it's worth an `EXPLAIN` against a production-sized mailbox before GA — deferring the benchmark, not the index check.
5. **Nits**:
   - *Batch row consumed when nothing removed* — deferred with reason: retained rows stay retained permanently (history/edits don't un-happen), so a re-reversal could only matter in the "retained because of a surviving engagement, engagement later deleted" window. Changing the consume-the-receipt contract alters existing 404-on-second-reversal expectations; happy to do it as a follow-up if you want re-reversal supported.
   - *Racing enrolment → 500 FK violation* — deferred: window is the reversal committing between another transaction's contact upsert and its engagement insert; the caller's retry succeeds (fresh contact + enrolment). Mapping 23503 → 409 would be nice but no `ForeignKeyViolation` mapping exists anywhere in `internal/identity` today — out of scope for this PR.
   - *Skip-path lock coverage* — fixed: `TestDeleteImportBatchWaitsForKeyShareEnrolment` stages the enrolment with only the FK's KEY SHARE held and pins that the reversal waits and retains both rows.

Re-verified after the fixes: full `TestDeleteImportBatch*` suite (15 tests), Go contract runner, TS contract (18) and Python contract (19) against a fresh `e2a-contract-server` — all green.

PR #770, opened 2026-07-31
feat(outbound): fire-time per-agent rate limit + scheduled reply/forward proof

jiashuoz:
## Summary

Hardens scheduled sending in two slices.

**A. Fire-time per-agent rate limit.** The HTTP acceptance path enforces "max 60 sends/min/agent" (in-memory, per-process), but scheduled messages accumulate as River jobs and can burst past the advertised per-agent throughput at the upstream provider when they fire — and the in-memory limiter cannot coordinate multiple replicas. This adds `internal/sendrate`, a Postgres-backed per-agent sliding-window log (migration `094`), gated in `SendWorker.Work` immediately before provider submission:

- Over-budget jobs **release the send claim** and `river.JobSnooze` until the window frees capacity (+ deterministic per-message jitter so a deferred backlog doesn't re-wake in lockstep). Deferral burns no attempt, meters no usage, and emits no lifecycle/terminal event — the message simply fires late.
- Immediate, scheduled, reply, forward, HITL-approved, and test submissions share one per-agent bucket (all flow through the same worker).
- Gate store errors **fail closed** (release + fixed snooze, never an unthrottled submit); the existing 72h retry horizon bounds deferral loops; the acceptance-time limiter stays as request-abuse control.
- Correct across replicas/workers via `SELECT ... FOR UPDATE` row serialization; all timestamps from the DB server's `clock_timestamp()` (app-clock and txn-start-time skew never widen the budget). Crash semantics documented in the package doc: a crash between reserve and submit burns one slot conservatively; a crash after submit is settled by `provider_accepted_at` evidence with no second slot.
- Metrics/logs: `e2a_outbound_rate_deferred_total` (unlabeled) + one IDs-only log line per deferral; `docs/observability.md` updated.

**B. Scheduled reply/forward proof.** `send_at` already worked on reply/forward via the shared `deliver()` path but only send had coverage. Adds: handler tests (future → 202 `status=scheduled` + echo + `wait=sent` does not poll; past → immediate; >90 days → 400); a behavioral pin that future `send_at` on a direct loopback self-send is refused 400 (previously spec-text only); thin HTTP trash/restore coverage for a scheduled message; contract scenarios `scheduled_reply_fields` / `scheduled_forward_fields` / `scheduled_send_at_validation` (existing vocabulary only, self-cleaning); and TS + Python SDK serializer tests. MCP pass-through coverage was already complete and stays green.

No `/v1` surface change — no OpenAPI/SDK regeneration needed (`spec-check` + `generate-sdk-check` green).

## Verification (all fresh runs in the branch worktree)

- `go build ./...`, `make test-unit` — green
- Integration (`-tags integration`): `internal/sendrate`, `internal/outboundsend`, `internal/telemetry`, `internal/httpapi`, `internal/agent` — green (agent suite passes isolated in 264s; an earlier failure was the 600s default timeout under heavy parallel load, not a regression)
- `make spec-check`, `make generate-sdk-check`, `make cover-check` — green (new `sendrate` floor 80, actual 80.0%)
- Contract: Go runner full suite green; TS 18 passed/14 skipped; Python 19 passed/13 skipped (skips are the pre-existing `inject_message` no-store-access rule, which the 3 new scenarios follow)
- TS SDK unit 223 passed, MCP 287 passed, Python 502 passed + mypy clean
- **Local e2e against a real instance** (Postgres + Mailpit): scheduled send fired ~1s after `send_at` and `wait=sent` returned in 0.07s; >90d → 400; past → immediate; **burst proof**: 120 messages accepted across two acceptance windows with the same fire instant → exactly 60 `rate_limited` deferrals in the log, first 60 delivered at the fire instant, the deferred 60 delivered ~61s later, **all 120 delivered, zero failures**, no leaked River jobs, no recipient data in logs.

## Reviews

Independent + adversarial review passes (both read-only, adversarial ran live probes): no blockers. Resolved in this branch:

- **Jitter (adversarial MEDIUM)**: deferred backlog re-woke in lockstep each window → added deterministic FNV-per-message jitter (< window/4) with a determinism/spread regression test.
- **`clock_timestamp()` (adversarial LOW)**: `now()` is txn-start time and could stamp events early under lock contention, widening the budget ~`lock_wait/window` → switched both prune cutoff and event stamp.
- **`limit ≤ 0` latent edge** → denies all with a sane `RetryAt` (+ test).
- **`AGENTS.md` sync** (package bullet + coverage floor list) and observability doc nits.

Deferred with reason: FK-cascade unit test (cascade verified live by the adversarial pass), suppression-check-before-gate reordering (would reopen the race the current ordering closes), configurable window/limit (v1 hardcodes 60/min; constructor leaves the seam).

## Remaining risks

- A retryable provider error consumes a new slot per attempt (intended — each attempt is a provider submission), so an agent in a provider-error retry loop burns its own budget; fail-safe direction, visible via the new counter.
- A message suppressed during its snooze burns one slot before the terminal suppression failure (fail-safe, one slot per message).
- Contract/prober test servers run gate-less by design (nil = allow-all), so the limiter itself is not exercised by the contract runners — it is covered by DB-backed worker tests and the local e2e above.


jiashuoz:
## Review: fire-time per-agent rate limit + scheduled reply/forward proof

**Verdict: approve with two follow-ups.** The design is sound and the implementation matches its stated contract. Everything below was verified independently in a clean worktree at `053367a5` — no results taken from the PR description.

### Verification performed

- `go build ./...`, `go vet` on the touched packages — clean
- `make test-unit` — exit 0 (confirms no other implementer of the widened `telemetry.Metrics` / `outboundsend.Metrics` interfaces broke)
- `go test ./internal/{sendrate,outboundsend,telemetry}` — green, DB-backed
- `-tags integration` on the new `httpapi` scheduled tests and `TestSelfSend_FutureSendAtRejected` — green
- **FK safety**: `messages.agent_id` → `agent_identities(id)` (`migrations/001_init.sql:77`), so `SendJob.AgentID` is exactly the key migration 085 references — no orphan-insert failure mode.
- **Test hygiene**: `agent_send_rate_windows` has an FK into `agent_identities`, which is in `truncateAll`'s TRUNCATE list, so CASCADE reaches it — no repeat of the `inbound_intake` leak class.
- **Lock ordering**: `Reserve` takes the agent row (FK KEY SHARE, first insert only) *before* the rate row, the same order `DeleteAgent` uses — no deadlock cycle. A rate-deferred message sits at `accepted` with a NULL claim, so it doesn't trip `ErrSendInProgress` on agent deletion.
- Contract scenarios mirror the existing `scheduled_send_fields` shape exactly (202 `status=scheduled` / `delivery_status=accepted`).

### Findings

#### 1. `rateJitter` panics on a sub-4ms window — `internal/outboundsend/worker.go:628`

`maxJitter.Milliseconds()` truncates to `0` for any window under 4ms, and `h.Sum32() % 0` is an integer divide by zero. The `maxJitter <= 0` guard on line 623 only catches windows below 4ns. Confirmed with a probe (a `RateGate` returning `Window() = 2ms`):

```
panic: runtime error: integer divide by zero
  outboundsend.rateJitter(...) worker.go:628
  outboundsend.(*SendWorker).Work(...) worker.go:461
```

Not reachable in production (the window is hardcoded to 1 minute) and the existing tests use 400ms, but it is a panic inside a River worker reachable from any future short-window test or from a configurable window — which this PR explicitly leaves as a seam. One line:

```go
ms := maxJitter.Milliseconds()
if ms <= 0 { return 0 }
return time.Duration(h.Sum32()%uint32(ms)) * time.Millisecond
```

#### 2. Deferral re-drive cost is quadratic in backlog size (design note, not a defect)

Each deferral costs `ClaimSend` + `Reserve` + `ReleaseSend` — 3 transactions, 2 of them `UPDATE`s on `messages` — and every still-deferred job repeats that each window. Draining a backlog of N costs approximately `N²/(2·limit)` re-drives.

A realistic burst (an agent accepts 60/min for an hour, all scheduled for 9am → N ≈ 3,600) is ~108k wake-ups and ~216k `messages` row updates over the drain. The 259,200 ceiling named in `docs/observability.md` works out to ~560M. Those wake-ups also occupy outbound queue workers, so one agent's backlog adds latency to every other tenant's sends.

The new jitter fixes lockstep but not volume. The worker already holds `AcceptedAt`/`ScheduledAt`, so scaling the snooze by how long the message has already been deferring (capped well under the 72h horizon) would make the number of re-drives logarithmic with no schema change. Worth a follow-up issue rather than a blocker — the mechanism is correct; this is only the cost curve. Flagging it because scheduled bursts are precisely what the feature invites.

#### 3. The `sendrate` coverage floor has zero headroom

In-package coverage is **79.4%**; it reaches the declared floor of 80 only because `Window()` is exercised cross-package by `outboundsend`'s tests. Any refactor tips `cover-check` red for a non-reason. A one-line `Window()` assertion in `store_test.go` fixes it.

#### 4. Minor docs gap

`docs/observability.md` is updated well, but `docs/api.md`'s `send_at` section doesn't mention that a scheduled burst now drains at ≤60/min/agent. The existing "`scheduled_at` is a 'not before' bound" language technically covers it, but someone scheduling 3,600 messages for 9am gets an hour-long drain with no status change — one clause there would set expectations.

### Checked and found correct

- Fail-closed on gate error (release + fixed snooze, never an unthrottled submit); both the error and deferral paths are bounded by the 72h horizon.
- A deferral burns no attempt, meters no usage, and emits no lifecycle/terminal event — pinned by both unit and real-Postgres tests, including against the *real* usage tracker.
- `clock_timestamp()` over `now()` for both the prune cutoff and the event stamp is correct, and the reasoning about transaction-start time widening the budget under lock contention holds.
- Gate placement after the ramp reservation is the right trade-off: a rate slot regenerates every minute, a ramp reservation is per-day, so burning the cheap one on a message the other gate will reject would be the worse ordering.
- `RetryAt = kept[0] + window` is the exact sliding-window semantics, and a denied reservation consumes no slot.
- Slice B is test-only and adds no `/v1` surface; the contract scenarios use existing vocabulary and are self-cleaning.

---
🤖 Review by [Claude Code](https://claude.com/claude-code)

jiashuoz:
Thanks for the thorough pass, Claude — all four findings addressed in `88e075b3`:

1. **`rateJitter` sub-4ms panic** — real; my guard only caught `window/4 <= 0`, not the millisecond truncation. Fixed with the `ms <= 0` guard plus a regression test (`TestSendWorker_RateLimitedTinyWindowNoPanic`, 2ms gate window). Good catch.
2. **Quadratic re-drive cost** — agreed, and the 3,600-message example is exactly the shape this feature invites. Filed as follow-up issue #771 (elapsed-time-scaled snooze, no schema change) rather than expanding this PR.
3. **Coverage headroom** — added a `Window()` assertion in `store_test.go`; in-package coverage is now 82.4% (was 79.4%).
4. **`docs/api.md` `send_at` drain expectations** — added a clause: `scheduled_at` is a "not before" bound, bursts drain at 60/min/agent, with your 3,600-message example.

Verified after the fix: `go build`, `go vet`, unit + integration for `internal/sendrate` and `internal/outboundsend` — all green.

PR #776, opened 2026-07-31
test: pin the generated-header and CTE-snapshot bug classes

jiashuoz:
## Why

The first live staging run of the v1.5.0 conformance suite (e2a-ops release-pipeline run 30612956986) caught two bug classes that offline CI structurally missed:

1. **Generated optional headers hitting the wire unguarded** — the TS generator emitted `setHeaderParam` unconditionally for optional params, so an omitted etag arrived as the literal `If-Match: undefined` and every unkeyed `contacts.update` / `contacts.setOutreach` failed 412. Fixed in #774 (generation-pipeline guard + TS static audit + offline RequestContext tests + live contract tests).
2. **Data-modifying CTE reading the statement snapshot** — `UpdateEngagementIfUnchanged` read the row back via the outer SELECT of a `WITH updated AS (UPDATE …)` statement; in Postgres that outer query predates the CTE's own write, so an accepted conditional update returned the pre-update row and a dead ETag. Fixed in #775.

Those PRs fixed the *instances*. This PR pins the *classes*.

### A. Python SDK static header-guard audit (`sdks/python/tests/test_generated_header_guards.py`)
The Python generator already emits `if if_match is not None:` guards — but nothing pinned that. The generated-code freshness gate re-blesses whatever a new generator version emits, so a generator upgrade that dropped the guard would regenerate green and ship `If-Match: None`, failing only at the staging gate again. The audit (runs in the per-PR `python-sdk` job) fails if any generated api module assigns a param-sourced header without its own None guard, requires an explicit allowlist entry for any deliberate exception (empty today — unlike TS, Python's retry layer needs no Idempotency-Key stub), and carries non-vacuity + self-discrimination checks so emission-shape drift fails loudly instead of shrinking the denominator silently.

### B. CTE-snapshot guard (`internal/sqlguard`, runs in every plain `go test ./...`)
Repo-wide tripwire modeled on `internal/logredact`'s logguard. For every tracked non-test Go file it resolves each string expression **including `+` concatenation through package consts** — load-bearing: the original bug's outer `FROM contact_engagements` lived in the const `engagementFrom`, so a literal-only scan misses the actual incident. It then parses WITH clauses (paren-balanced, literal/comment-masked, tolerant of non-CTE `WITH TIME ZONE`/`WITH ORDINALITY`), classifies each CTE body, and fails when a data-modifying CTE's outer SELECT reads FROM/JOIN a modified table instead of the CTE's RETURNING output — with an error message explaining the snapshot hazard and pointing at `UpdateEngagementIfUnchanged`'s fix as the reference.

Kept conservative per #775's audit: read-only CTEs (webhook/webhookpub leases, `threadAnchorBatchQuery`, `detachThreadChildrenBatchTx`) and modifying CTEs read back through their RETURNING output (`GetPrincipalByAPIKey`, `GetMessageWithContent`) do not trigger. **The current tree passes with zero waivers.** Legitimate pre-write reads can waive with `// cte-snapshot-safe: <reason>` (stale waivers are themselves flagged).

### C. Live-contract coverage ratchet (`sdks/typescript/test/v1/contract-coverage.test.ts`)
Assessment answer: the per-PR ts-contract job had **no coverage denominator** — `contract.test.ts` interprets scenarios.yaml over raw HTTP, `contract-client.test.ts` exercised ~7 ergonomic methods, and nothing failed when a new SDK method shipped with no live per-PR test (exactly how the If-Match bug reached staging first). The cheap gate was feasible, so it's included: denominator = runtime introspection of the built client (`test/coverage/introspect.ts`, the same walker the staging-side live gate uses); numerator = direct `client.x.y(...)` call sites in `contract-client.test.ts`; every method must be exercised or carry an explicit `NOT_CONTRACT_TESTED` entry. The 66-entry seed is the honest inventory of today's gap (4 with specific can't-run reasons, 62 marked seeded), pruned in both directions — phantom and newly-covered entries fail the gate — so the list can only shrink truthfully. Wired into `npm run test:contract` (per-PR).

Residual gaps flagged as follow-ups (deliberately not built here):
- The numerator is a static call-site scan, not runtime recording — indirect calls aren't seen, and a skipped test still counts its sites. Upgrading to recorder-based coverage (recorder.ts machinery already exists) is straightforward if the seed list ever shrinks enough to make it worth it.
- The **python-contract** job has no equivalent: `tests/test_contract.py` is the scenario interpreter only; there is no Python analogue of `contract-client.test.ts`, so the Python ergonomic client has zero per-PR live coverage. Same ratchet pattern would apply if/when a Python contract-client suite exists.

### D. Conventions (AGENTS.md → Testing strategy)
Three new bullets: new SDK surface needs live coverage in per-PR CI, not only staging; generated clients are pipeline artifacts — never hand-edit, and the load-bearing emission shapes are pinned by static audits; data-modifying CTEs must not re-select from the modified table.

## Evidence the tree passes

- Python: `pytest tests/` under 3.12 with the coverage gate — **508 passed, 37 skipped** (includes the new audit); `mypy` clean; mutation check: stripping one real `if if_match is not None:` guard from `contacts_api.py` fails the audit.
- Go: `go build ./...`; `go test ./internal/sqlguard/` — all guards + discrimination suites pass, **zero waivers needed**; mutation check: restoring the pre-#775 `engagements.go` flags exactly `internal/identity/engagements.go: CTE "updated" modifies table "contact_engagements" but the outer SELECT reads FROM/JOIN "contact_engagements"`.
- TS: `npm run typecheck` clean; `test:unit` **233 passed**; full `npm run test:contract` against a locally built `cmd/e2a-contract-server` (CI-equivalent, Postgres-backed) — **3 files / 22 passed, 14 skipped** (seed-gated scenarios, same as CI), including the ratchet's 3 tests green on the seeded allowlist.
- `scripts/check-repository-text-integrity.sh` passes.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -314,6 +314,29 @@ manually on every API change even though the template won't remind you.
   feature, extend all three interpreters in parity, add runner-level regressions
   in every language, and prefer dynamic placeholders over fixtures that age or
   collide under concurrent runs.
+- **New SDK surface needs live coverage in per-PR CI, not only staging.** The
+  post-merge staging conformance gate must never be the FIRST live exercise
+  of a method — that is how `If-Match: undefined` shipped (#774). The
+  ts-contract job enforces a denominator: every ergonomic client method must
+  be exercised by `sdks/typescript/test/v1/contract-client.test.ts` or carry
+  an explicit allowlist entry in `contract-coverage.test.ts` (a ratchet —
+  shrink it; grow it only with a reason the contract server can't run the
+  path).
+- **Generated clients are pipeline artifacts — never hand-edit them.** Any
+  behavior fix belongs in the generation pipeline (`make generate-sdk`, e.g.
+  the TS post-processing step `scripts/guard-optional-header-params.py`), so
+  regeneration cannot undo it. Static audits pin the load-bearing emission
+  shapes per PR — optional header params must stay guarded in both generated
+  clients (TS: `test/v1/optional-header-params.test.ts`; Python:
+  `tests/test_generated_header_guards.py`) — because the freshness gate alone
+  re-blesses whatever a new generator version emits.
+- **Data-modifying CTEs must not re-select from the table they modified.** In
+  Postgres the outer query runs on the statement snapshot, which predates the
+  CTE's own writes — `WITH x AS (UPDATE t … RETURNING …) SELECT … FROM t`
+  returns the pre-update row (this broke `UpdateEngagementIfUnchanged`'s ETag
+  chain, #775). Read FROM the CTE's RETURNING output, or split write and read
+  into two statements in one transaction. Guarded repo-wide by
+  `internal/sqlguard` (waiver: `// cte-snapshot-safe: <reason>`).
 - **Schema-change rule**: when changing a table shape, add/update DB-backed
   tests in every package that writes direct SQL against that table — the
   idempotent migration runner will not catch drifted runtime SQL.
--- a/internal/sqlguard/cte_snapshot_discrimination_test.go
+++ b/internal/sqlguard/cte_snapshot_discrimination_test.go
@@ -1,0 +1,215 @@
+package sqlguard
+
+import (
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"strings"
+	"testing"
+)
+
+// TestAnalyzeSQLDiscrimination pins the matcher's own behaviour: the shapes
+// it must catch (the reason it exists) and the shapes it must not flag (the
+// reason it can hold a zero-waiver tree). The clean cases are distilled from
+// the five real in-tree CTE sites audited in PR #775.
+func TestAnalyzeSQLDiscrimination(t *testing.T) {
+	hazardous := map[string]string{
+		"outer SELECT re-reads the UPDATEd table (the UpdateEngagementIfUnchanged bug)": `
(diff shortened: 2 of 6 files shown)


jiashuoz (commented):
P2 — `internal/sqlguard/cte_snapshot_guard_test.go:286` does not pin the whole snapshot-read bug class it claims to guard. The reference regex only sees a modified table immediately after `FROM` or `JOIN`, so legal PostgreSQL comma joins pass silently:

```sql
WITH updated AS (
  UPDATE messages SET status = $1 WHERE id = $2 RETURNING id
)
SELECT m.* FROM updated u, messages m WHERE m.id = u.id;
```

That outer `messages` read uses the same pre-write statement snapshot as the #775 failure, but `analyzeSQL` returns no finding. Please recognize comma-separated relations in the outer `FROM` list and add this exact hazardous shape to `TestAnalyzeSQLDiscrimination`.

Two non-blocking hardening notes while touching the guard: `analyzeSQL` returns after the first syntactically valid `WITH`, so a later hazardous statement in the same resolved string is missed; and the Python header audit only recognizes a plain-identifier RHS, so a wrapped param such as `str(if_match)` is invisible while the aggregate non-vacuity threshold can stay green. Tests at 30fd9802: `go test ./internal/sqlguard -count=100` and the isolated Python header-guard suite pass.

PR #780, opened 2026-07-31
fix(api): input validation and error mapping

jiashuoz:
Four input-validation / error-mapping defects found by an adversarial edge-case sweep against **live staging, candidate `sha-32ce45dc`**. Each answered a permanent client error with a 5xx or a silently wrong success, and each is fixed with regression tests at the layer that owns the rule. A fifth candidate finding — strict `If-Match` comparison — was investigated and **deliberately not adopted**; see below.

## Defect 1 (HIGH) — NUL byte in client strings → 500 instead of 400

Postgres text columns cannot store `U+0000`, so any caller string carrying one reached the driver and came back as `internal_error`. Five authenticated-caller-reachable vectors on staging:

| Vector | Staging request id | Was |
|---|---|---|
| `POST /v1/agents/{email}/messages`, NUL in `subject` | `req_e8636da9e69ea87cc227e85f` | 500 `internal_error` "failed to accept message for send" |
| `POST /v1/contacts`, NUL in `display_name` | `req_6264c471b04b1f8f1f1091f1` | 500 "failed to create contact" |
| `POST /v1/contacts`, NUL in a metadata **value** | `req_8935d9a809f83acf898dfe2c` | 500 |
| `POST /v1/contacts`, NUL in a metadata **key** | `req_0f538473c1df984b86fb3d97` | 500 |
| `PUT /v1/agents/{email}/contacts/{address}`, NUL in `stage` | `req_d42087f5fc0b7fc7072c0e9e` | 500 "failed to save outreach state" |

### The rule chosen

> **No client-supplied string in a request may contain `U+0000` — anywhere in the JSON body (at any depth, including object *keys*), or in a path, query, or header parameter. Violation → `400 invalid_request` with `error.details.fields[].location` naming the field.**

Blanket, not per-field, on purpose. Guarding only the five columns that happened to be reachable leaves the sixth for the next caller to find, and "reject only what is persisted" is unknowable from outside the API — a caller cannot tell which of `subject` / `text` / `html` / `filename` lands in a text column and which is only composed into MIME. One rule, stated once, is the only version a client can reason about, and it costs nothing: `U+0000` is not meaningful in an email address, a display name, a subject line, or a MIME header. In JSON a NUL can only arrive as the six-character `\^@` escape.

### The seam

`registerOp` (`internal/httpapi/request_content.go`) wraps `huma.Register` and runs the guard on the **bound** input struct, between Huma's schema validation and the handler. All 72 operations were rewired to it, and a drift test (`TestEveryOperationGoesThroughRegisterOp`) fails the build if a future operation calls `huma.Register` directly — that opt-out would otherwise be invisible until someone found the 500 in production, which is exactly how these five were found. No partial writes before or after: the guard runs before any store call, and the tests assert the row does not exist afterwards.

## Defect 2 (MEDIUM) — CR/LF in `attachments[].filename` / `content_type` → 500 instead of 400

The security control **worked** — the composer refuses the header-injection attempt — but the refusal escaped as a 5xx, so SDK retry logic would hammer a permanent client error. Evidence: `req_e7be6c6d4b2ed074056b2717` (CR in filename), `req_5221059f93a505a8755e18f6` (LF), `req_d80a0151080139d2ac01d320` (content_type).

Now validated at the API edge in `validateAttachments`, before compose → `400 invalid_attachment`: the same answer the 1024-byte filename cap already gives, and the same treatment CR/LF in `subject` already got. The composer's guard stays as defence in depth. The error message does not echo the offending filename (it contains the CR/LF); the location rides in `details.fields[].location` instead.

## Defect 3 (MEDIUM) — `importContacts` broke its own per-row isolation for `maxLength`

The endpoint promises "one malformed row never rejects the rest of the upload", and that held for every app-level check. It did **not** hold for schema-level `maxLength`: a 400-char `display_name` in a 3-row batch returned `422` with `location: body.contacts[1].display_name`, **no `batch_id`, nothing imported** (`req_d7f946efd74b61aa9be5e00f`); same for a 321-char address. A 1000-row CSV with one long name lost the whole upload.

Fix: the item-level `maxLength: 320` on `ContactImportRow.address` / `.display_name` moves out of the request schema and into the handler's per-row prevalidation, alongside the address parse and metadata-bounds checks that already fail one row at a time. The bound is unchanged (320 code points) — only the layer enforcing it moved. Over-long address → row `status: "failed"`, `code: "invalid_recipient"`; over-long `display_name` → `code: "invalid_request"`.

`maxItems: 1000` stays a whole-request 4xx, as do the request-level `agent_email` / `stage` caps. The operation description and `docs/api.md` now state the split explicitly:

- **Row-level** (fails one row, batch imports): unparseable or over-long address, over-long `display_name`, metadata outside the per-contact bounds.
- **Request-level** (rejects everything): malformed JSON, unknown field, missing `address` key, a NUL anywhere in the body, >1000 rows, body >20 MiB, bad `agent_email`/`stage`.

Note the deliberate seam with defect 1: a NUL is *not* a row content bound — it is a byte the request document may not contain at all, the same class as malformed JSON — so it rejects the whole request. `TestImportRejectsNULAtRequestLevel` pins that decision.

## Defect 4 (LOW) — empty `If-Match` silently degraded a guarded write to unconditional

Confirmed on the wire over a raw TLS socket, so not a client artifact: `If-Match: ` (present, no value) + `PATCH /v1/contacts/{addr}` → **200 with a new ETag** (`req_91af2b0f6fe551c8df501fda`). Same on the engagement PUT (`req_d7bcfbf956f33b6064bc2601`). A client interpolating an unset variable into the header would overwrite a concurrent edit and get a 200 telling it the guard had held.

RFC 9110 §13.1.1 defines the field value as `"*" / 1#entity-tag` — at least one member — so an empty value is a **malformed request (400 `invalid_request`)**, not a precondition that failed (412). The quoted-empty form `""` is a syntactically valid validator and keeps its existing 412. Go collapses a present-but-empty header to the same `""` an absent one produces, so presence is checked against the raw request.

Fixed at all three layers:
- **Server** — `emptyIfMatchError` on `PATCH /v1/contacts/{address}` and `PUT /v1/agents/{email}/contacts/{address}`.
- **CLI** — `cli/src/commands/contacts.ts`: an empty (or whitespace-only) `--if-match` is now a usage error instead of a truthiness check that silently dropped the option.
- **MCP** — `mcp/src/tools/contacts.ts`: `if_match` is `z.string().min(1).optional()` on both conditional tools.

## Considered and deliberately NOT adopted — strict `If-Match` comparison

The sweep also flagged that `etagMatches` strips the `W/` prefix and compares the rest, so `W/"<current>"` matches and the write lands, where RFC 9110 §13.1.1 specifies *strong* comparison for `If-Match`. An earlier revision of this branch "fixed" that. **It has been reverted, and the code carries an expanded comment so nobody fixes it again.**

The deleted comment ("Weak prefixes are tolerated on input because intermediaries add them") was a Chesterton's fence, and the operational facts back it:

- `api.e2a.dev` is Cloudflare-proxied and the edge **is** actively transforming responses — `content-encoding: br` with `server: cloudflare` confirmed live on `/.well-known/oauth-protected-resource`.
- Our origin does not compress (no `encode` directive in either Caddyfile), so the transformation is the edge's own.
- Cloudflare converts strong ETags to weak whenever it transforms a response. "Respect Strong ETags" is Enterprise-only; e2a.dev is on the Free plan.

So under strict comparison a client would `GET` `ETag: W/"abc"`, echo back exactly what it was handed, and receive a **permanent 412 that no retry clears** — breaking every conditional write on the contacts/outreach surface, **in prod only**. The staging host is DNS-only (not proxied), so the conformance gate structurally cannot observe this: a green gate would have proved nothing. Tolerance is what makes optimistic concurrency work through a transforming intermediary; RFC purity loses to that.

The guard is not weakened by the tolerance: the compared body is the full validator, which moves on every accepted write, so a stale tag never matches however it is dressed. `etagMatches`'s body is byte-identical to `origin/main` on this branch — only comments, docs, and tests changed.

Two related findings kept as documentation rather than code changes:

- **`If-Match: *` was audited and is already correct.** It means "if any current representation exists": both call sites load the resource first, so it succeeds on an existing record and is refused on a missing one — on the engagement PUT it returns 412 rather than enrolling (`req_00d8492fdb57aee8a32da11c`). Now pinned by tests and stated in the header docs.
- **The reported "W/ matches standalone but 412s inside a comma list" inconsistency does not exist.** The prefix strip has always been inside the loop, so both paths behave identically. See the methodology note below for what actually produced that reading. A test now pins standalone/comma-list parity so the two cannot drift apart.

**Follow-up:** once v1.5.0 is on prod, check a real contact ETag through the edge — `curl -sD- -H "Authorization: Bearer $KEY" https://api.e2a.dev/v1/contacts/<address> | grep -i ^etag` — and revisit only if it comes back strong.

## User-visible behaviour changes

Two, both deliberate:

1. **NUL is now rejected where it was previously accepted.** `subject` and the four contact/engagement vectors changed 500 → 400 (strictly better). But NUL in `text`, `html`, and `attachments[].filename` was **accepted with 202** on staging and is now a 400. That is the cost of one consistent rule, and it is the right trade: a NUL in a mail body or MIME filename has no legitimate use, the alternative is a contract no client can predict, and the previously-accepted path was writing a byte the storage layer may reject elsewhere later.
2. **An empty `If-Match` is now 400 instead of an unconditional write** (defect 4), including the CLI usage error and the MCP schema rejection.

Defects 2 and 3 are not behaviour regressions: 2 turns a 500 into the documented 400, and 3 only widens what is accepted. `If-Match` matching behaviour is unchanged from `main`.

## Spec / SDK

`make spec` regenerated `api/openapi.yaml` (never hand-edited). `make openapi-compat-check` against `origin/main`: **"No breaking changes to report."** The only structural change is the *removal* of two item-level `maxLength` constraints, which widens accepted input.

That removal matters downstream: the generated Python SDK enforced `max_length=320` client-side with pydantic, so a long name would have raised before the request left the process — defeating defect 3 for Python users. Both SDK bases were regenerated with `make generate-sdk` (pinned OpenAPI Generator v7.16.0); the diff is the dropped constraint plus description text.

## Methodology note — the sweep's stateful probes produced a phantom finding

Worth recording, because it affects how the next wire sweep should be read.

The `p3_etag.py` probe sequence mutates the resource it is measuring, and every successful conditional write rotates the ETag. Log line 115 (`wk-W/ + CORRECT etag`) **succeeded**, rotating the validator from `88eba810…` to `f400b8d0…` — the probe's own `wk-re` re-reads show it. The lowercase-`w/` probe at line 119 and the comma-list probe at line 121 then both reused the now-stale `88eba810…`, so both returned 412 **for the wrong reason**: the tag was stale, not weak-prefixed.

That is where the "comma-list inconsistency" came from — it was an artifact, not behaviour. The lowercase-`w/` result was over-attributed the same way: it failed for two reasons at once (stale tag *and* a prefix that genuinely is not `W/`, since RFC 9110 spells the weak prefix in uppercase), and the sweep could not distinguish them. Both are resolved by the fixed-`current` unit table in `TestETagMatchesToleratesTheWeakPrefix`, which is the right instrument for a pure matching question.

**Recommendation for future wire sweeps:** re-read the validator between probes, or reset state between them. Otherwise every result after the first successful write is suspect, and a mutating probe silently turns into a test of staleness rather than of the thing it names.

## Additional review fixes in this branch

Three items from the adversarial review, applied here rather than deferred:

- **The `registerOp` drift guard was itself bypassable, twice over.** It grepped only for the literal `huma.Register(`, but huma v2.39.0 also exports `Get` / `Post` / `Put` / `Patch` / `Delete` / `AutoRegister`, all of which register internally. Worse, every one of them is **generic**, so explicit type instantiation is valid Go — `huma.Post[In, Out](api, "/x", handler)` compiles and contains no `huma.Post(` substring at all. Either form would have skipped the NUL guard *and* passed the test; the second hole predates this branch, since the original `huma.Register(` literal had it too. Matching is now the regex `huma\.(Register|AutoRegister|Get|Post|Put|Patch|Delete)\s*[\[(]` over a recursive `filepath.WalkDir`, with `TestHumaRegistrationPatternCatchesBothCallForms` pinning both call forms and a set of must-not-match cases (`registerOp(`, `huma.Operation`, `humanErrorMessage`) so the guard stays quiet enough to be trusted. Verified end-to-end by injecting each form into the package and confirming the walk fails.

  Its limits are now stated in the test comment rather than assumed: this is a regex over source text, not a type-aware analysis, so it does **not** catch an aliased import (`h.Register(`), a dot-import (bare `Register(`), or a sub-package that wraps huma and is called through the wrapper. Closing those needs `go/analysis`; the guard's job is to stop an ordinary mistake by someone who does not know `registerOp` exists, not to defeat someone routing around it deliberately.
- **The walker built location strings eagerly.** Every element, key, and field allocated a path string that a clean request throws away. Locations now come from a path stack reused across the walk and are rendered only when a violation is found. Measured on a clean 1000-row import: **350 KB / 14.7k allocs → 64.5 KB / 4k**. The path machinery is now effectively free — the same body without metadata objects costs 438 B / 12 allocs — and the remainder is `reflect`'s own map-iteration cost over the 1000 metadata maps. (A linked list of path nodes was tried first and does *not* work: the recursion makes escape analysis heap-allocate every node, trading string allocs for struct ones at no gain. The comment records that so it is not re-attempted.)
- **`docs/api.md` overclaimed the NUL rule.** It stated the rule absolutely, but the guard sits on the 72 Huma `/v1` operations; the legacy chi `/api/*` dashboard surface, the public unsubscribe handler, and the inbound SMTP path are outside that seam. The sentence is now scoped to `/v1` and names the entry points that validate independently.

## Test evidence

All suites run locally, all green.

| Suite | Result |
|---|---|
| `go test -short ./internal/httpapi/` | ok |
| `go test ./internal/outbound/...` | ok |
| `go test -short ./internal/identity/ ./internal/agent/` | ok (Postgres on :5433) |
| `make test-unit` (9 packages) | all ok |
| `make spec-check` | ok |
| `make openapi-compat-check` | no breaking changes |
| CLI `npm test` | 268 passed (18 files) |
| MCP `npm test` + `test:types` | 292 passed (13 files) |
| TS SDK `npm run build` + `npm test` | 233 passed |
| Python SDK `uv run pytest` | 506 passed, 37 skipped, 94% coverage |

New regression tests:

- `internal/httpapi/request_content_test.go` — all five NUL vectors with their staging request ids, the newly-rejected `text` / `html` / `filename` / recipient cases, query and header params, a clean-request control, direct walker coverage (nil pointers, byte slices, `json:"-"`, nested slices/maps, map keys), the import request-level seam, the `registerOp` drift guard, and `BenchmarkScanCleanImportBody` for the allocation budget.
- `internal/httpapi/outbound_test.go` — CR / LF / CRLF in filename and content_type at both the `validateAttachments` unit level and over HTTP, plus a clean-content_type control.
- `internal/httpapi/contacts_import_test.go` — over-long `display_name` and address each fail one row while neighbours import and a `batch_id` is still returned; a row exactly at the bound still imports; the `maxItems` cap stays request-level.
- `internal/httpapi/contacts_test.go` / `engagements_test.go` — empty `If-Match` → 400 with the row untouched, quoted-empty still 412, omitted still unconditional; a CDN-weakened *current* validator still completes the write while a weakened *stale* one still 412s; `*` pinned on both existing and missing resources; a table over `etagMatches` covering strong / weak / weak-in-a-list / mixed lists / lowercase-`w/` / wildcard / garbage. **`TestETagMatchesToleratesTheWeakPrefix` is a FENCE, not a regression test** — it passes against unmodified `main`, because nothing about the matching behaviour changed. Its job is to fail loudly for whoever next reads §13.1.1 and tightens the comparison; it should not be cited as evidence that a bug was fixed. Its comment says so.
- `cli/src/__tests__/contacts.test.ts`, `mcp/tests/tools.test.ts` — empty `--if-match` / `if_match` rejected, real validators still forwarded, omission still unconditional.

---

## Review fixes (2026-08-01)

Four validated review findings applied on top:

- **Shared contract coverage for the schema change** (`tests/contract/scenarios.yaml`, new `contacts_import_row_length_isolation` scenario): this PR removed `maxLength: 320` from `ContactImportRow` in the published spec — over-long rows now take the per-row isolation path instead of a whole-request 422. Per `AGENTS.md`, a changed public /v1 wire schema needs a shared scenario, not just handler tests. The scenario places an over-long `display_name` row between two valid rows and asserts HTTP 200, the offending row `failed`/`invalid_request`, both neighbours imported, and no contact left behind; self-cleaning via batch reversal + cleanup deletes. Verified green through the live Go, TypeScript, and Python contract runners.
- **Bounded error reflection** (`internal/httpapi/request_content.go`): the NUL guard echoed caller-controlled map keys verbatim into `error.message` and `details.fields[0].location` — with the 20 MiB import cap and body binding preceding auth, a huge key reflected ~2x its size, unauthenticated. Keys are now truncated at render time via the existing `truncateForError` idiom (32 chars + `…`); regression test asserts the response stays bounded for a 64 KiB key.
- **Drift-guard gaps closed** (`internal/httpapi/request_content_test.go`): `huma.Head` (exported by huma v2.39.0) added to the registration-pattern alternation with fixtures for both call shapes; the guard no longer skips `request_content.go` wholesale — it now allows exactly the one `huma.Register` call inside `registerOp`. The package-scope limit of the walk is documented (repo-wide grep confirms no registration outside `internal/httpapi`).
- **Untracked coverage artifact** (own commit): `sdks/python/.coverage` (pytest-cov SQLite DB embedding absolute local paths, first committed in v1.3.0) removed from tracking and ignored via `sdks/python/.coverage*`.


jiashuoz (commented):
The team found several merge-blocking gaps at `cc932895`:

1. **P2 — MCP still rejects the whole import before the API can provide per-row isolation.** `mcp/src/tools/contacts.ts:134-140` keeps row-level Zod validation (`address: contactAddress`, `display_name.max(320)`, bounded metadata). A three-row MCP import with a 321-character middle display name, malformed address, or out-of-bounds metadata imports nothing. The REST/CLI/SDK contract introduced here is the opposite: return 200, fail that row, and import its neighbors. Make the import-specific MCP row schema permissive for content while retaining request-level shape/cardinality checks, and add a mixed-row MCP regression.

2. **P2 — the new scanner creates an unauthenticated large-error amplification path.** At `internal/httpapi/request_content.go:213`, a clean caller-controlled map key is appended verbatim to the location. On a terminal NUL value, that path is copied into both the top-level error message and `details.location`. `template_data` can consume nearly the 40 MiB outbound body limit, and this scan runs before handler authentication/rate limiting, so a request with a huge key can allocate/encode roughly two extra copies and send an enormous response. Report the containing map or use a bounded/sanitized key segment; add a long-key regression.

3. **P2 — the documented blanket `/v1` rule is Huma-only.** Raw documented routes at `internal/httpapi/httpapi.go:527-567` (WebSocket, capability attachment download, magic-link handlers) never traverse `registerOp`. For example the WS route decodes `%00`, passes it into `WSHandle`, and the store lookup turns the PostgreSQL error into 404 rather than the promised 400. Apply equivalent validation through `/v1` middleware/raw-route helpers, or narrow the public contract explicitly; add a raw-route regression.

4. **P2 — invalid UTF-8 still produces the same permanent-client-error-as-5xx class.** `request_content.go:177-180` checks only byte zero. Huma v2.39.0’s query parser ignores `url.QueryUnescape` errors, so `GET /v1/contacts?import_batch_id=%FF` binds a Go string containing `0xff`; PostgreSQL rejects it and `contacts.go:331-334` maps it to 500. Reject `!utf8.ValidString` for values and keys with 400 and add an over-the-wire `%FF` test.

5. **P2 — the shared live contract required by AGENTS.md was not updated.** `tests/contract/scenarios.yaml` is unchanged; the existing import scenario only covers an unparseable address. Reintroducing item `maxLength` (and the whole-request 422) would leave Go/TS/Python contract runners green. Add an overlong row between valid neighbors and assert 200 + failed row + imported neighbors; cover representative NUL/empty-If-Match behavior or extend the scenario DSL symmetrically.

6. **P3 — the registration drift guard omits an exported registrar.** `internal/httpapi/request_content_test.go:293` omits `huma.Head`, exported and generic in pinned Huma v2.39.0. A normal `huma.Head(...)` or `huma.Head[I,O](...)` bypasses `registerOp` while the test stays green. Add `Head` plus both fixtures; also avoid skipping the entire wrapper file when only one direct call is legitimate.

7. **P3 — remove the mutable coverage artifact.** This PR rewrites tracked `sdks/python/.coverage`, a 53,248-byte SQLite pytest-cov DB whose delta is absolute local worktree paths. Prefer deleting it and ignoring scoped `.coverage*`; at minimum restore it before merge.

Validation at the pinned head: `go test -short ./internal/httpapi -count=1` passed, targeted CLI/MCP tests passed (153 total), `git diff --check` is clean, and GitHub CI is green. Those suites do not exercise the gaps above.

PR #781, opened 2026-07-31
fix(limits): guard cache fills against a racing invalidation

jiashuoz:
## The bug

`internal/limits/enforcer.go` caches resolved `account_limits` in-process for
`limits.cache_ttl_seconds` (60s in the hosted config). `cachePut` had no
generation guard, so this interleaving silently defeated an explicit
invalidation:

1. A request calls `Get` → cache miss → reads the **OLD** `account_limits` row.
2. The billing sidecar commits **NEW** limits and POSTs
   `/api/internal/limits/invalidate` (HMAC-authed) → the cache entry is deleted.
3. The in-flight `Get` from step 1 then `cachePut`s the **pre-write** limits with
   a **fresh** TTL.

**Customer-visible symptom:** a user who just upgraded their plan keeps hitting
spurious `402 limit_exceeded` for up to a minute. It self-heals on TTL, but it
defeats the one mechanism billing has to make an upgrade take effect
immediately — exactly the moment a paying customer is watching.

## The fix: a process-wide invalidation epoch

*(This section describes the current design. An earlier revision of this branch
used a per-user generation tombstone; review found that design retains one map
entry per distinct userID ever passed to `Invalidate` for the process lifetime —
an unbounded-growth vector (~283 bytes/entry, HMAC-authed endpoint, 2 GB VM with
no container mem_limit). The epoch design below addresses that finding.)*

A single **process-wide invalidation epoch** (`DBEnforcer.gen`, guarded by the
existing mutex):

- `cacheGet` returns the current epoch alongside the (missing) entry, sampled
  under the same lock acquisition as the lookup.
- `Get` carries that epoch across the DB read (no lock is held across it).
- `cachePut` stores the fill only if the epoch has not advanced; a losing fill
  is **dropped outright** (still returned to its caller, just not cached), so
  the next `Get` re-reads the post-write row — one extra query, no staleness.
- `Invalidate` **increments the epoch and deletes the entry**. No tombstone:
  the epoch survives on the enforcer itself, so an in-flight fill can still
  discover it lost even when the user had no cached entry — which is precisely
  the racing case, and there is a dedicated test pinning it.

**Why global is safe:** the epoch drops a strict superset of the fills the
per-user counter dropped, so it cannot reintroduce the stale-fill race. The
cost is an occasional cross-user false drop (one extra DB query) when an
invalidate lands inside another user's fill window; prod invalidates are
Stripe-event-driven — a handful per day — versus millisecond fill windows, so
the collision rate is noise.

**Why this bounds memory:** `delete` is now the only map effect of
`Invalidate`, so the cache is again bounded by users actually cached by real
fills — the pre-branch bound — while keeping the race guard. The earlier
"needs a TTL sweep" caveat is resolved by this design.

A counter rather than a timestamp because two events inside one clock tick
must still compare different.

### `isUserID` stays, as defense-in-depth

`handleInvalidateLimits` still 400s a malformed `user_id`. With the epoch
design this is no longer the memory bound (nothing the caller sends can grow
the map) — it stays because it keeps the endpoint contract tight and makes a
misbehaving caller fail loudly. **Note:** the original review suggested a UUID
parse — that would have taken the endpoint down. Identity user ids are **not**
UUIDs; `identity.generateID` emits 16 random bytes hex encoded (32 lowercase
hex chars), so a UUID parse would reject every real user. The check matches
the real shape.

## Deployment topology: a real gap this PR does NOT close

**Correcting an earlier version of this description, which claimed the hosted
blue/green setup was unaffected. That claim was wrong.** The corrected analysis,
verified against `e2a-ops`:

The sidecar posts to `http://haproxy:8080` and `be_http` does have the inactive
slot `disabled` — but **only in the durable config at rest**
(`haproxy/haproxy.bootstrap.cfg:123-124`). During a deploy,
`scripts/deploy-bluegreen.sh:289` sets the candidate `ready` **while the old
slot is still ready and serving**, and the old slot isn't drained until after a
health-wait loop bounded at 30×2s (typical overlap 2-10s). There is **no
`balance` directive**, so HAProxy defaults to roundrobin across *both* UP
servers in that window. The invalidate is a single POST with no retry and no
fan-out (`billing/internal/accountlimits/writer.go:115-142`; failure is logged
and swallowed).

The damaging interleaving: during the overlap a customer request lands on the
cold candidate and reads pre-write limits; the invalidate POST roundrobins to
the **old** slot; the candidate caches the stale value with a fresh TTL — and
then survives the deploy as the active slot. That's a **~60s (one TTL) spurious-
402 window that the generation guard cannot fix**, because the invalidate never
reached that process.

To be clear about scope: this is **pre-existing, and neither introduced nor
fixed by this PR**. Probability is very low — it needs a plan change to coincide
with a seconds-long cutover window that occurs once per deploy — and it
self-heals on TTL. The genuine fix is fanning the invalidate out to both slots,
or having the sidecar retry across them: ops/billing work, out of scope here.

The generically-correct scope note on the endpoint (`limits_api.go`) stands: one
POST clears the cache of exactly the process that served it, so any
multi-process deployment must fan out or accept the TTL.

Also confirmed there is only **one** cache layer for this data:
`internal/identity` reads `max_contacts` / `max_templates` / `max_webhooks`
straight from `account_limits` on every call, uncached.

## Tests

A `fakeStore.onGet` seam fires between the DB read and the `cachePut`, so the
race is reproduced **deterministically** — no sleeps, no flakes.

| Test | What it pins |
|---|---|
| `TestEnforcer_InvalidateDuringFillIsNotCached` | the core race: a fill that raced an invalidate must not be cached |
| `TestEnforcer_InvalidateDuringFillIsNotCached_Concurrent` (real goroutine + handshake, for `-race`) | same, with real mutex interleaving |
| `TestEnforcer_InvalidateAdvancesGenerationForUncachedUser` | the trap: an invalidate for a user with no entry must still advance the epoch |
| `TestEnforcer_InvalidateNeverCachedUsersRetainsNothing` | **new** — the memory bound: 10k invalidates for never-cached (well-formed) ids leave `len(cache) == 0`, and the race guard still holds afterwards |
| `TestEnforcer_InvalidateEvictsOnlyTheNamedUsersCachedEntry` | rewritten from `InvalidateIsPerUser`: an unrelated user's invalidate must not evict another user's **cached** entry (it may drop an in-flight **fill** — intentionally conservative under the global epoch; costs one DB read) |
| `TestEnforcer_ZeroValueInvalidateDoesNotPanic` | nil-map invariant; epoch now sampled via the normal `cacheGet` path |
| `TestInvalidateLimits_RejectsMalformedUserID` / `AcceptsWellFormedUncachedUserID` | the edge shape check (defense-in-depth) and its counterweight |
| `MissFillIsCached`, `CacheStillWorksAfterInvalidate` | counterweights — a guard that dropped *every* fill, or poisoned a user permanently, would pass the race tests |

```
$ go test -race ./internal/limits -count=1                                        # ok (all pass)
$ go test -race ./internal/agent -run "TestInvalidateLimits|TestIsUserID" -count=1 # ok
$ gofmt -l / go vet                                                                # clean on touched files
```

No API surface change, so no `make spec-check` implications.

### DIFF
--- a/internal/agent/limits_api.go
+++ b/internal/agent/limits_api.go
@@ -106,14 +106,14 @@ func (a *API) handleInvalidateLimits(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, "user_id is required", http.StatusBadRequest)
 		return
 	}
-	// Bound what can enter the enforcer's cache map. Invalidate now records a
-	// generation tombstone for whatever key it is handed — it MUST, because a
-	// user with no cached entry is exactly the case the invalidate/fill race
-	// guard exists to cover: the generation has to advance so an in-flight
-	// fill can see it lost. That makes an unvalidated user_id an unbounded
-	// map-growth vector (the map has no TTL sweep), so the bound belongs here,
-	// at the edge, rather than in Invalidate where it would silently
-	// reintroduce the race.
+	// Shape-check the user_id before it reaches the enforcer. This is
+	// defense-in-depth, not the memory bound: the enforcer's process-wide
+	// invalidation epoch means Invalidate only ever DELETES map entries
+	// (see limits.DBEnforcer.Invalidate), so even an unvalidated user_id
+	// could not grow the cache. Rejecting garbage here still keeps the
+	// endpoint's contract tight and makes a misbehaving caller (a sidecar
+	// bug sending the wrong field) fail loudly instead of silently
+	// no-oping.
 	if !isUserID(req.UserID) {
 		http.Error(w, "user_id is malformed", http.StatusBadRequest)
 		return
@@ -132,7 +132,7 @@ const userIDLen = 32
 // userIDLen lowercase hex characters. A shape check, not an existence check —
 // confirming existence would mean a DB round trip on a path whose whole purpose
 // is to avoid one, and a well-formed id for a deleted user is harmless (it
-// tombstones a key nothing will ever read).
+// deletes a cache key nothing will ever read).
 func isUserID(s string) bool {
 	if len(s) != userIDLen {
 		return false
--- a/internal/agent/limits_api.go
+++ b/internal/agent/limits_api.go
@@ -60,6 +60,16 @@ type invalidateLimitsRequest struct {
 // it's an internal seam between the OSS server and its operator's
 // provisioner. Self-hosters who don't run a provisioner simply leave
 // InternalAPISecret empty and the endpoint 503s.
+//
+// Scope caveat: the cache this busts is PER PROCESS. One POST clears it
+// in exactly the server process that handled the request. An operator
+// running several server replicas behind a load balancer must fan the
+// call out to every replica (or accept that non-targeted replicas keep
+// serving the old caps until limits.cache_ttl_seconds expires); a
+// single POST through the load balancer only reaches one of them. The
+// enforcer's generation guard makes each individual invalidation
+// reliable — it cannot make one invalidation reach processes it was
+// never sent to.
 func (a *API) handleInvalidateLimits(w http.ResponseWriter, r *http.Request) {
 	if a.internalAPISecret == "" {
 		http.Error(w, "internal api not configured", http.StatusServiceUnavailable)
--- a/internal/agent/limits_api.go
+++ b/internal/agent/limits_api.go
@@ -106,9 +106,44 @@ func (a *API) handleInvalidateLimits(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, "user_id is required", http.StatusBadRequest)
 		return
 	}
+	// Bound what can enter the enforcer's cache map. Invalidate now records a
+	// generation tombstone for whatever key it is handed — it MUST, because a
+	// user with no cached entry is exactly the case the invalidate/fill race
+	// guard exists to cover: the generation has to advance so an in-flight
+	// fill can see it lost. That makes an unvalidated user_id an unbounded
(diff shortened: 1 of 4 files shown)


jiashuoz (commented):
P2 — The new shape check does not actually bound tombstone cardinality. `internal/agent/limits_api.go:117` accepts a 2^128 keyspace, and every distinct HMAC-valid 32-character lowercase-hex ID reaches `internal/limits/enforcer.go:153`, which retains a tombstone for the process lifetime. A signed caller can therefore POST `000…001`, `000…002`, etc. and grow the map without bound; this is a regression from the old delete-on-unknown behavior. The HMAC requirement lowers exploitability, but a leaked credential or buggy trusted provisioner can still turn this Internet-reachable route into monotonic heap growth/OOM, and the PR claims to close this exact vector.

Please use a real cardinality bound: for example, a process-wide invalidation epoch (safe but may discard a few unrelated concurrent fills), a user-existence check on this rare internal path, or lifecycle/refcount/eviction that removes a tombstone after its in-flight fills finish. The lowercase-hex check can remain defense in depth, but it is not the bound. Add a regression that sends many distinct syntactically valid IDs and proves retained state stays bounded without restoring the stale-fill race.

Targeted evidence: `go test -race ./internal/limits` and `go test -race ./internal/agent -run "TestInvalidateLimits|TestIsUserID" -count=1` pass at aeae9cb2.

Issue #144, opened 2026-05-25
Audit follow-up: defense-in-depth findings (M2, M3, M4, M5, M8, M11)


jiashuoz:
Tracking the deferred audit findings from PR #142. These are real but either defense-in-depth or larger refactors that didn't earn a same-PR fix when batched with the H-tier work. Listing here so they don't get lost.

For full context on the audit that surfaced these, see the analysis trail in PRs [#140](https://github.com/Mnexa-AI/e2a/pull/140) and [#142](https://github.com/Mnexa-AI/e2a/pull/142).

---

## M2 — Pagination cursor unsigned (DoS surface)

**Where:** [`internal/agent/api.go:1786-1825`](https://github.com/Mnexa-AI/e2a/blob/main/internal/agent/api.go#L1786)

The opaque pagination token is just `base64url(json)` — no HMAC. Any authenticated caller can edit `CreatedAt` / `ID` in the JSON, e.g. set `CreatedAt` to `0001-01-01` to force a full-table-scan-then-truncate against `messages`. The `cursor.AgentID` check prevents cross-tenant reach, but a single bearer can repeatedly request expensive windows. Bounded by partial indexes today; integrity gap is real.

**Fix sketch:** HMAC the cursor JSON with `cfg.Signing.HMACSecret` (same pattern as the approval-token signer in `internal/approvaltoken/token.go`). Reject on tamper.

---

## M3 — Magic-link approval tokens replay-able within TTL

**Where:** [`internal/approvaltoken/token.go`](https://github.com/Mnexa-AI/e2a/blob/main/internal/approvaltoken/token.go), used by [`internal/agent/hitl_magic_api.go`](https://github.com/Mnexa-AI/e2a/blob/main/internal/agent/hitl_magic_api.go)

A signed approval / reject token is valid for the entire message TTL plus a 10-minute grace. The status-race guard at approve/reject time blocks repeated state changes once the message is finalized, but if a token leaks (mail archive scrape, Slack-channel paste, Gmail link-tracking) before the reviewer acts, an attacker who reaches the link can race them.

**Fix sketch:** Either (a) track `magic_link_consumed_at` on the messages table (new column + migration + index + handler check), or (b) maintain a single-use `jti` table that the handler INSERTs into and lets the unique constraint reject replay.

---

## M4 — DKIM private keys stored unencrypted at rest

[...]

PR #782, opened 2026-07-31
fix(api): bind pagination cursors to the endpoint and account that issued them

jiashuoz:
## The bug

Pagination cursors are HMAC-signed, so tampered / garbage / truncated / SQLi /
traversal cursors correctly return `400 invalid_cursor` on all list endpoints —
that part was already solid.

What was **not** rejected: a **validly-signed cursor minted by a different
endpoint**. Every account-level list shares the same `{"c":…,"i":…}` payload
shape, so a foreign cursor decoded cleanly and was applied as a real keyset
anchor.

**Affected: 10 endpoints.** The original report named 8; a review pass found two
more, both confirmed against a clean `origin/main`:

| Endpoint | |
|---|---|
| `/v1/agents` `/v1/events` `/v1/domains` `/v1/webhooks` `/v1/templates` `/v1/account/api-keys` `/v1/account/suppressions` `/v1/reviews` | the originally-reported 8 |
| `/v1/webhooks/{id}/deliveries` | **newly found** — with no status filter, a foreign cursor's empty status matched |
| `/v1/starter-templates` | **newly found** — a foreign cursor's absent alias decoded as `""`, silently re-serving page 1 |

Sharpest live evidence (staging `sha-32ce45dc`): a **contacts** cursor
`{"c":"2026-07-31T19:14:24Z","i":"cnt_5526…"}` replayed on `/v1/events?limit=3`
returned `200 {"items":[]}` while the same query without the cursor returned 3
events — so a client silently concludes "no events".

**No cross-tenant leakage** — results stayed account-scoped throughout. This is a
correctness/DX bug, not a security bug. The failure mode is a silent wrong
answer, which is nastier to debug than a 400 but is not data exposure.

The agent-scoped lists (messages, conversations, engagements, agent
suppressions) and `/v1/contacts` already rejected foreign cursors, because they
pin a **filter fingerprint** a foreign cursor can't match. The account-level
lists have no filters to fingerprint, so nothing bound them.

## The fix

**1. Resource binding.** Every cursor carries the minting collection inside its
signed payload:

```
envelope = {"r": "<resource>", "p": <the resource's own cursor struct>}
cursor   = base64url(envelope) "." base64url(hmac_sha256(secret, base64url(envelope)))
```

`EncodeCursor` / `DecodeCursor` take a `CursorResource` argument, so the binding
is **impossible to forget** — a new list endpoint can't mint an unbound cursor,
because the resource is a required parameter rather than a struct field someone
has to remember to set. It's inside the HMAC, so a client can't re-target a
cursor by editing it.

**2. Parent binding on `/v1/webhooks/{id}/deliveries` (review finding M1).** The
resource discriminator only proves a cursor came from *some* deliveries list.
That endpoint is parameterized by `{id}`, so a resource-bound cursor minted on
webhook A was **still** accepted on webhook B and A's anchor handed to
`ListDeliveries(ctx, B, …)` — the identical failure mode this PR exists to fix,
one scope level down. `deliveriesCursor` now pins `WebhookID`. Every other
parameterized list already pinned its parent (messages/conversations →
`AgentID`, engagements/agent suppressions → `AgentEmail`, message_lifecycle →
`AgentID`+`MessageID`); deliveries was the lone gap. Both webhooks are
caller-owned, so this was never a leak either.

**3. Loud view assertion (review finding L1).** `decodeKeyset` now rejects a
cursor carrying `Deleted=true`. Unreachable today — agents is the only
trash-view collection and it uses `decodeKeysetView` — but the resource argument
is compile-enforced while the view argument is not, so a future endpoint that
grows `?deleted=` and keeps calling `decodeKeyset` would silently lose its view
binding with no compile error.

A foreign cursor gets `400 invalid_cursor` with a specific message.
`ErrCursorResourceMismatch` wraps `ErrInvalidCursor`, so callers branching on the
broad sentinel keep working. All 16 cursor-minting handlers route through one
`s.decodeCursor(resource, …)` helper; existing filter-fingerprint checks are
preserved verbatim, just moved after the decode.

## ⚠️ In-flight cursors are invalidated

The payload shape changed, so a cursor issued by a previous build no longer
decodes. **A client mid-pagination gets one `400 invalid_cursor` and has to
restart the query.** Keeping them working was possible but would have meant a
legacy path that keeps the bug alive for exactly the cursors most likely to be
replayed. This is also the same trade #144 M2 already made when it hard-rejected
pre-signing cursors on the reasoning "cursors are ephemeral, so a client
mid-pagination simply restarts the query". Flagging it so it's a decision, not a
surprise.

## Tests

`internal/httpapi/cursor_binding_test.go` wires all 11 cursor-minting
account-level lists, mints a **real** cursor from each through the real handler,
and replays it on every other — a **110-pair matrix** (11 × 10), plus a test
pinning the exact staging repro (contacts cursor → `/v1/events`).

Verified against a clean `origin/main` worktree:

| | pre-fix | post-fix |
|---|---|---|
| foreign-cursor pairs accepted with `200` (of the original 72) | **64** | 0 |
| all 110 pairs | — | all `400 invalid_cursor` |
| `TestCursor_ContactsCursorOnEventsIsRejected` | **FAIL** (200, empty items) | pass |
| `TestDeliveriesCursor_ForeignWebhookRejected` | **FAIL** (200, webhook B's rows under A's anchor) | pass |

The 8 pairs that already behaved correctly pre-fix are exactly the ones targeting
`/v1/contacts` — the one collection with a filter fingerprint.

Counterweights, all passing: every endpoint still walks **its own** cursor with
disjoint pages; webhook A's own deliveries cursor still walks only A's rows; the
resource is proven signed rather than merely tagged; `decodeKeyset` still accepts
a live cursor; and the existing agent-scoped `"different filters"` rejections plus
the trash-view binding are unchanged (whole `internal/httpapi` suite green).

```
$ go build ./... && go vet ./internal/httpapi/     # clean
$ go test ./internal/httpapi/                      # ok  7.193s
$ make spec-check                                  # ok — no OpenAPI drift
```

No OpenAPI surface change: `cursor` stays an opaque string and `invalid_cursor`
was already the documented code for this class. `docs/api.md` now states the
binding covers the collection, its **path parameters**, and its filters —
wording that is only accurate once M1 is fixed.

## Accepted / deferred (on the record)

- **No version field in `cursorEnvelope`.** The next payload-shape change will
  repeat today's hard break. A `"v"` int would allow dual-decode across one
  release. Deliberately deferred — adding it now costs a byte and buys nothing
  until there *is* a second shape, and this PR already takes the one break.
  Worth doing whenever the envelope next changes.
- ~~**Matrix covers the 11 account-level endpoints only.**~~ **Closed in round
  2**: the matrix now covers all 16 cursor-minting collections and varies the
  principal as well as the endpoint (see below).
- **No `iat`/expiry field.** Considered and rejected: `cursorEnvelope` decodes
  via `encoding/json`, which ignores unknown fields and zero-fills missing
  ones, so adding `iat` later is backward-compatible in both directions and
  needs no invalidation. Only the principal binding is genuinely breaking, so
  only it must ride this window. `docs/api.md` no longer claims cursors are
  "short-lived" — nothing enforces an age; it now gives the client guidance
  the code actually guarantees (treat cursors as ephemeral; on
  `invalid_cursor`, restart the query).

## Round 2: bind the minting ACCOUNT via a derived MAC key

External review found the remaining axis: the envelope carries no principal and
the MAC key was the deployment-global `config.Signing.HMACSecret`, so **account
A's `/v1/events` cursor verified for account B on `/v1/events`**. B's query
stays hard-scoped by `user_id` — still **no cross-tenant leakage** — but the
foreign anchor silently mispositioned B's page (typically an empty `200`), the
same wrong-answer class this PR exists to eliminate.

**Why it lands now:** this PR already hard-rejects every pre-envelope cursor and
is unmerged, so no cursor in the new format exists in the wild. Binding now
costs zero additional breakage; binding after release would mass-invalidate
persisted cursors a **second** time, against SDKs whose `.page()` docstrings
recommend persisting them for checkpoint/resume.

**Mechanism — the account goes in the MAC KEY, not the envelope:**

```
key    = HMAC-SHA256(secret, "e2a/cursor/v1|" + accountID)
cursor = base64url(envelope) "." base64url(hmac_sha256(key, base64url(envelope)))
```

- **Key derivation, not a plaintext field**: the envelope is signed but NOT
  encrypted, and cursors travel in query strings, access logs, and referrer
  headers — a plaintext account ID would leak. The envelope schema is
  untouched.
- **Bound to `Principal.User.ID`** — the level every list query is already
  scoped at. Not the API key ID (rotation would break in-flight pagination),
  not the scope/agent (an account-scoped key must be able to resume a cursor an
  agent-scoped key minted over the same rows).
- **Required parameter, threaded explicitly** from each handler's authenticated
  principal through `EncodeCursor`/`DecodeCursor` and the `decodeCursor` /
  `encodeKeyset` / `encodeKeysetView` wrappers — never pulled from the request
  context — so a missed site is a compile error, the same mechanism that makes
  the resource binding impossible to forget. Applied uniformly to all sixteen
  collections, **including the global-catalog starter templates** (the handler
  is auth-gated, so there is no anonymous case; a blanket rule stays auditable).
- **Rotation preserved**: `DecodeCursor` derives per candidate secret inside
  the existing constant-time `hmac.Equal` loop.
- **No new error surface**: a foreign account's cursor fails plain MAC
  verification and falls out as the existing `400 invalid_cursor` —
  deliberately indistinguishable from a forgery, so nothing reveals whether a
  stolen cursor was "well-formed". No OpenAPI change (`make spec-check` clean).

**Tests (round 2):**

- `TestCursor_ForeignAccountCursorRejected`: a second account (own agent, own
  rows) joins the fixture; for **every** collection, A's cursor in B's hands is
  `400 invalid_cursor` while the A→A and B→B continuations keep walking.
- The matrix now covers **all 16 collections** — the five agent-scoped lists
  (messages, conversations, engagements, agent_suppressions,
  message_lifecycle) are added. They were already foreign-cursor-safe via their
  mandatory agent/filter fingerprints, so that part is coverage, not a fix.
  The N×N foreign-endpoint matrix is now 16 × 15 = 240 pairs.
- `TestCursor_AccountBindsViaMACKeyNotEnvelope`: the account changes the MAC
  but the payload segments stay byte-identical (nothing leaks into the
  unencrypted envelope); a foreign-account decode fails as plain
  `ErrInvalidCursor`, not a distinguishable mismatch; rotation verifies an
  old-secret cursor through the derived key.
- **Shared contract scenario** (`tests/contract/scenarios.yaml`
  `cursor_binding_foreign_cursor_rejected`, per the AGENTS.md rule that public
  `/v1` behavior changes need cross-SDK scenario coverage): a genuinely minted
  `/v1/contacts` cursor replayed on `/v1/agents` → `400 invalid_cursor`; a
  forged cursor → `400 invalid_cursor`; the own-endpoint continuation still
  walks; self-cleaning. **DSL limitation, on the record:** the contract runners
  share a single test account, so a true two-account replay cannot be staged in
  the scenario language; the account axis is deliberately indistinguishable
  from the forged-cursor case the scenario does assert, and its two-account
  matrix lives in the handler tests above.

```
$ go test ./internal/httpapi/ -count=1                 # ok  4.980s
$ go test -tags=integration ./tests/contract/ -count=1 # ok  8.747s (incl. new scenario)
$ gofmt -l … && go vet ./internal/httpapi/             # clean
$ make spec-check                                      # ok — no OpenAPI drift
```

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -39,7 +39,11 @@ MCP tool surface), see:
   plus-tagged addresses (`a+tag@x.com`). The official SDKs encode this for you;
   hand-rolled clients must do it themselves.
 - **Pagination.** List endpoints return `{ items, next_cursor }`; pass
-  `next_cursor` back as `?cursor=…` to page forward. The SDKs auto-page.
+  `next_cursor` back as `?cursor=…` to page forward. The SDKs auto-page. A
+  cursor is bound to the endpoint that issued it and to the filters it was
+  minted under: replaying one on a different list, or with different filters,
+  is a `400 invalid_cursor` rather than a silently wrong page. Cursors are
+  opaque and short-lived — on `invalid_cursor`, drop it and restart the query.
 - **Idempotency.** Nine mutating operations honor an opt-in `Idempotency-Key`
   header: `sendMessage`, `replyToMessage`, `forwardMessage`, `approveReview`,
   `createWebhook`, `rotateWebhookSecret`, and `createApiKey`, plus the beta
--- a/docs/api.md
+++ b/docs/api.md
@@ -40,12 +40,13 @@ MCP tool surface), see:
   hand-rolled clients must do it themselves.
 - **Pagination.** List endpoints return `{ items, next_cursor }`; pass
   `next_cursor` back as `?cursor=…` to page forward. The SDKs auto-page. A
-  cursor is bound to the exact request that issued it: the collection, its path
-  parameters (the agent, webhook, or message it was listed under), and the
-  filters it was minted with. Replaying one on a different list, a different
-  parent, or with different filters is a `400 invalid_cursor` rather than a
-  silently wrong page. Cursors are opaque and short-lived — on
-  `invalid_cursor`, drop it and restart the query from the first page.
+  cursor is bound to the exact request that issued it: the account that minted
+  it, the collection, its path parameters (the agent, webhook, or message it
+  was listed under), and the filters it was minted with. Replaying one on a
+  different list, a different parent, with different filters, or under a
+  different account's credentials is a `400 invalid_cursor` rather than a
+  silently wrong page. Cursors are opaque; treat them as ephemeral — on
+  `invalid_cursor`, drop the cursor and restart the query from the first page.
 - **Idempotency.** Nine mutating operations honor an opt-in `Idempotency-Key`
   header: `sendMessage`, `replyToMessage`, `forwardMessage`, `approveReview`,
   `createWebhook`, `rotateWebhookSecret`, and `createApiKey`, plus the beta
--- a/docs/api.md
+++ b/docs/api.md
@@ -40,10 +40,12 @@ MCP tool surface), see:
   hand-rolled clients must do it themselves.
 - **Pagination.** List endpoints return `{ items, next_cursor }`; pass
   `next_cursor` back as `?cursor=…` to page forward. The SDKs auto-page. A
-  cursor is bound to the endpoint that issued it and to the filters it was
-  minted under: replaying one on a different list, or with different filters,
-  is a `400 invalid_cursor` rather than a silently wrong page. Cursors are
-  opaque and short-lived — on `invalid_cursor`, drop it and restart the query.
+  cursor is bound to the exact request that issued it: the collection, its path
+  parameters (the agent, webhook, or message it was listed under), and the
+  filters it was minted with. Replaying one on a different list, a different
+  parent, or with different filters is a `400 invalid_cursor` rather than a
+  silently wrong page. Cursors are opaque and short-lived — on
+  `invalid_cursor`, drop it and restart the query from the first page.
 - **Idempotency.** Nine mutating operations honor an opt-in `Idempotency-Key`
   header: `sendMessage`, `replyToMessage`, `forwardMessage`, `approveReview`,
   `createWebhook`, `rotateWebhookSecret`, and `createApiKey`, plus the beta
--- a/internal/httpapi/account.go
+++ b/internal/httpapi/account.go
@@ -167,8 +167,8 @@ func (s *Server) handleListSuppressions(ctx context.Context, in *listSuppression
 	var afterAddress string
 	if in.Cursor != "" {
 		var cur suppressionsCursor
-		if err := DecodeCursor([]string{s.deps.CursorSecret}, in.Cursor, &cur); err != nil {
-			return nil, NewError(http.StatusBadRequest, "invalid_cursor", "invalid pagination cursor")
+		if err := s.decodeCursor(cursorAccountSuppressions, in.Cursor, &cur); err != nil {
+			return nil, err
 		}
 		afterCreatedAt = cur.CreatedAt
 		afterAddress = cur.Address
@@ -189,7 +189,7 @@ func (s *Server) handleListSuppressions(ctx context.Context, in *listSuppression
 	var nextCursor string
 	if hasMore {
 		last := list[len(list)-1]
-		nextCursor, err = EncodeCursor(s.deps.CursorSecret, suppressionsCursor{CreatedAt: last.CreatedAt, Address: last.Address})
+		nextCursor, err = EncodeCursor(s.deps.CursorSecret, cursorAccountSuppressions, suppressionsCursor{CreatedAt: last.CreatedAt, Address: last.Address})
 		if err != nil {
 			return nil, NewError(http.StatusInternalServerError, "internal_error", "failed to build pagination cursor")
 		}
--- a/internal/httpapi/account.go
+++ b/internal/httpapi/account.go
@@ -167,7 +167,7 @@ func (s *Server) handleListSuppressions(ctx context.Context, in *listSuppression
 	var afterAddress string
 	if in.Cursor != "" {
 		var cur suppressionsCursor
-		if err := s.decodeCursor(cursorAccountSuppressions, in.Cursor, &cur); err != nil {
+		if err := s.decodeCursor(user.ID, cursorAccountSuppressions, in.Cursor, &cur); err != nil {
 			return nil, err
 		}
 		afterCreatedAt = cur.CreatedAt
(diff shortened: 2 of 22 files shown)


jiashuoz (commented):
Team review at exact head 5c374da65046626113bb9a980d4ed70ddcc4431d found one blocking correctness gap.

[P2] Bind account-scoped cursors to the authenticated account, not only the collection.

The new signed envelope contains only `{resource,payload}` (`internal/httpapi/pagination.go:106-109`) and uses the deployment-global HMAC secret (`pagination.go:240-242`). Neither the shared `keysetCursor` nor the custom payloads for the account-level lists carry `user.ID`. Consequently, a real cursor minted for account A on `/v1/events` verifies under account B on `/v1/events`; `events.go:278-294` accepts A’s position/filter tuple and applies that anchor to B’s otherwise correctly scoped query. If the anchor falls outside B’s rows, B gets a silent truncated or empty 200 page—the same wrong-answer class this PR fixes across endpoints. The issue also applies to agents, domains, webhooks, templates, API keys, reviews, account suppressions, and contacts. Starter templates are global, and parent-bound lists are already protected. There is no cross-tenant row leakage.

The 110-pair matrix cannot catch this because it uses one principal (`cursor_binding_test.go:89-96`) and varies only the endpoint. Please make authenticated subject/account scope a required signed component in the shared encode/decode path (or equivalently derive a per-account cursor MAC key), then add a two-token/two-account same-resource replay matrix: A-to-B must return `400 invalid_cursor`, while A-to-A and B-to-B walks remain valid.

Verification: focused cursor/pagination tests passed five consecutive exact-head runs; the full short `internal/httpapi` suite passed; `git diff --check` passed; all GitHub checks are green.

jiashuoz (commented):
Follow-up from the client/contract reviewer:

[P2] Add the required shared client contract scenario for this public behavior change.

The PR changes the observable API contract from `200` with a wrong page to `400 invalid_cursor`, but the exact-head diff does not change `tests/contract/scenarios.yaml` or any SDK/MCP/CLI contract test. The repository contract requires public API behavior changes to be exercised through the shared Go/TypeScript/Python harness; handler tests are not a substitute. Add a self-cleaning foreign-cursor scenario and assert that each client surfaces `invalid_cursor`. The new Go matrix also covers only 11 of the 16 cursor collections; messages, conversations, engagements, agent suppressions, and message lifecycle should either join the matrix or be protected by an exhaustive uniqueness/binding guard.

[P3] Correct the new “short-lived” documentation claim.

`docs/api.md:47-48` says cursors are short-lived, but `cursorEnvelope` contains no issued-at/expiry and `DecodeCursor` performs no age check. They remain valid until the signing secret changes or their anchor ceases to be meaningful. Remove that wording or implement/document an actual lifetime policy.

PR #783, opened 2026-07-31
fix(store): return authoritative post-write state from write paths

jiashuoz:
## The class

PR #775 fixed the severe form of one bug: a data-modifying CTE whose outer SELECT ran on the pre-update statement snapshot, so an accepted conditional engagement update **committed the new state but returned the PRE-update row** — and therefore a permanently dead ETag. Its fix recipe was: do the write and the authoritative read **inside one transaction**, or **RETURN the full row from the write**.

That recipe was applied to one function. An audit found **six weaker cousins** still on `main`, all the same shape: *write, commit, then re-read from the pool*. None can permanently break a validator the way #775 did, but each leaves a real window another transaction can commit into — so the response either describes **somebody else's write**, or reports a **spurious error on a write that actually succeeded**.

All six now follow the #775 recipe.

## The six sites

| # | Site | Concrete wrong response | Approach |
|---|------|------------------------|----------|
| 1 | `identity.UpsertEngagement` (unconditional PUT path) | Commits, then `GetEngagement` on the **pool**. A `DeleteEngagement`, contact delete or import reversal in the gap → `ErrEngagementNotFound` → the handler answers **412 `precondition_failed`** (`httpapi/engagements.go`) on a request that sent **no If-Match**. `RecordInbound/OutboundActivity` (fires on every message touching that contact) can also repaint the row before a 201-created response describes it. | `RETURNING id` + read the joined row **in the same transaction** — literally the shape of `UpdateEngagementIfUnchanged`. (#775's fix comment cited this function as its model, but #775 reads inside its tx and this one did not; they agree now.) |
| 2 | `identity.UpdateAgentProtection` + handler `GetAgent` | Plain UPDATE, then a pool re-read. The PUT is a **full replace**, so a concurrent PUT hands the caller the **other writer's entire posture** as the authoritative-looking result of their own request. Concurrent trash → **500 "failed to reload agent"** after a committed write. | UPDATE + in-transaction re-read; the store now returns the agent. JOIN on `domains` for `domain_verified` rules out plain `RETURNING`. |
| 3 | Agent **PATCH** and **restore** (`httpapi/agents_write.go`, "Re-read for the authoritative post-update state") | Concurrent rename → response shows the **other** name. Concurrent trash → **500** after a committed rename. Concurrent re-trash → **500** after a committed restore. GA surface. | `UpdateAgentName` / `RestoreAgent` return the agent, read in-transaction. Restore keeps the **LIVE** projection, so it still proves the agent is visible again — the property the old post-commit re-read asserted. |
| 4 | `identity.UpdateWebhook` | `UPDATE … RETURNING id` (autocommit), then a separate `GetWebhookByID`. Concurrent PATCH → other writer's config in the response. Concurrent delete → **404 after a committed update**. | `UPDATE … RETURNING <full row>` — atomic by construction. The column list is now one shared const + `scanWebhook`, so read and RETURNING projections cannot drift. |
| 5 | `identity.UpdateTemplate` | Identical to #4; concurrent delete → **404 `template_not_found` after a committed update**. | Same: `UPDATE … RETURNING templateSelectColumns`. Alias-collision and not-found translations unchanged. |
| 6 | `identity.RestoreMessage` + handler `GetMessage` | Concurrent re-trash or purge → **500 "failed to reload message"** after a committed restore. | The restore transaction builds the detail view itself, using the **same** projection (read-marking included), so behaviour is unchanged apart from the race. |

## User-visible behaviour change

**Yes — and it is the point: the spurious 404 / 412 / 500s disappear.** A write that commits now always answers with the state it wrote, instead of occasionally answering `precondition_failed`, `not_found`, or `internal_error`. Responses also stop echoing a concurrent writer's configuration as if it were the caller's own result.

Nothing else changes:

- **No error-mapping changes for legitimate cases.** Every 404/412/409/500 that a genuinely-absent, genuinely-stale, or genuinely-live resource produces is byte-identical; only the *spurious* ones caused by the torn read are gone.
- **No response shape changes** — `make spec-check` green.
- **Lock ordering preserved.** Every added read is a plain `SELECT` taking no row locks, so the "exactly one row lock, then plain SELECTs" property #775's reviewer confirmed still holds. Checked against `DeleteImportBatch`'s contacts→engagements `FOR UPDATE` order (untouched — `UpsertEngagement` still locks agent → contacts → engagements) and against `DeleteAgent`.

Internal-only signature churn: `Deps.UpdateAgentName` / `Deps.UpdateAgentProtection` now return `*identity.AgentIdentity`; `RestoreAgent` / `RestoreMessage` move to new returning types `AgentRestoreOp` / `MessageRestoreOp`.

## Tests

New `internal/identity/post_write_read_test.go` — one DB-backed regression per site, using the blocking-interleaving technique from `contacts_reversal_test.go`:

1. a holder transaction takes the lock the write path needs;
2. the write path is started and confirmed blocked on it;
3. a **racer** is started and confirmed blocked **behind** it (lock waits are FIFO) — identified by a marker comment in its own query text, so "queued behind" is checked on that exact backend rather than a global waiter count;
4. the holder commits, so the racer lands **exactly** in the window the buggy shape leaves open.

Each racer carries no bind parameters, so pgx sends it over the simple protocol as one implicit transaction: once woken it runs *and commits* entirely server-side, while the buggy re-read still owes a client round-trip and loses. Each test also asserts the racer really landed, so a run that failed to interleave cannot pass silently.

**Pre-fix evidence** (against `origin/main`, for the three sites whose signatures did not change, 8 runs each):

| Test | Fails pre-fix | Observed failure |
|---|---|---|
| `TestUpsertEngagementReturnsTheRowItCommitted` | **8/8** | `UpsertEngagement returned engagement not found on a write that committed` |
| `TestUpdateWebhookReturnsTheRowItWrote` | **7/8** | `UpdateWebhook returned description "racer-won", want "mine"` |
| `TestUpdateTemplateReturnsTheRowItWrote` | **5/8** | `UpdateTemplate returned name "racer-won", want "mine"` |

The other four sites changed the store signature, so their tests only compile against the fix; they use the same interleaving and assert the same property.

**Post-fix:** all seven pass deterministically (6 consecutive `-count` runs, no flakes — with the fix the racer is *structurally* blocked until the read is done). `go build ./...`, `go vet ./...`, `go test ./internal/identity/ ./internal/httpapi/` green, plus `./internal/agent/ ./internal/webhook/ ./internal/usage/ ./internal/hitlworker/ ./internal/apiserver/ ./internal/contactdue/ ./internal/relay/` and `internal/e2e` compiling under `-tags integration`. `make spec-check` green.

## Also noted, NOT fixed (different class)

`UpdateWebhook`'s **re-enable cooldown is check-then-act**: a separate `SELECT auto_disabled_at` runs before the UPDATE (`internal/identity/webhooks.go`). Two racing re-enables can both pass it — harmless, since they converge on the same state — but the real exposure is the auto-disable worker setting `auto_disabled_at` **between** the check and the UPDATE, whose `enabled = true` branch then clears a cooldown that had just been armed. Folding the predicate into the UPDATE's `WHERE` needs a new not-found-vs-cooldown disambiguation branch (and interacts with the dynamic SET builder and the no-op-PATCH path), so it is not a trivially-safe change and is left for its own PR.

### DIFF
--- a/internal/agent/selfsend_screening_test.go
+++ b/internal/agent/selfsend_screening_test.go
@@ -27,7 +27,7 @@ func protectedSelfAgent(t *testing.T, store *identity.Store, label string, cfg i
 	t.Helper()
 	ctx := context.Background()
 	user, ag := selfAgent(t, store, label)
-	if err := store.UpdateAgentProtection(ctx, ag.ID, user.ID, cfg); err != nil {
+	if _, err := store.UpdateAgentProtection(ctx, ag.ID, user.ID, cfg); err != nil {
 		t.Fatalf("UpdateAgentProtection: %v", err)
 	}
 	refreshed, err := store.GetAgentByEmail(ctx, ag.EmailAddress())
--- a/internal/agent/test_send_async_test.go
+++ b/internal/agent/test_send_async_test.go
@@ -273,7 +273,7 @@ func TestSendTestCore_ScreeningReviewHolds(t *testing.T) {
 	ctx := context.Background()
 	user, ag := selfAgent(t, store, "testreview")
 
-	if err := store.UpdateAgentProtection(ctx, ag.ID, user.ID, identity.ProtectionConfig{
+	if _, err := store.UpdateAgentProtection(ctx, ag.ID, user.ID, identity.ProtectionConfig{
 		InboundGatePolicy:       "open",
 		InboundGateAction:       "review",
 		InboundScanSensitivity:  identity.SensitivityOff,
--- a/internal/agent/thread_identity_integration_test.go
+++ b/internal/agent/thread_identity_integration_test.go
@@ -215,7 +215,7 @@ func TestHITLReplyHoldCommitsOneIdempotentThreadDecision(t *testing.T) {
 	cfg.OutboundGatePolicy = "allowlist"
 	cfg.OutboundAllowlist = []string{"trusted@example.net"}
 	cfg.OutboundGateAction = "review"
-	if err := store.UpdateAgentProtection(ctx, ag.ID, user.ID, cfg); err != nil {
+	if _, err := store.UpdateAgentProtection(ctx, ag.ID, user.ID, cfg); err != nil {
 		t.Fatalf("enable outbound review: %v", err)
 	}
 	ag, err = store.GetAgentByEmail(ctx, ag.EmailAddress())
@@ -329,7 +329,7 @@ func TestHumanApprovedSelfSendPreservesHeldThreadAcrossLocalTwins(t *testing.T) 
 	cfg.OutboundGatePolicy = "allowlist"
 	cfg.OutboundAllowlist = []string{"trusted@example.net"}
 	cfg.OutboundGateAction = "review"
-	if err := store.UpdateAgentProtection(ctx, ag.ID, user.ID, cfg); err != nil {
+	if _, err := store.UpdateAgentProtection(ctx, ag.ID, user.ID, cfg); err != nil {
 		t.Fatalf("enable outbound review: %v", err)
 	}
 	ag, err := store.GetAgentByEmail(ctx, ag.EmailAddress())
--- a/internal/hitlworker/loopback_screening_test.go
+++ b/internal/hitlworker/loopback_screening_test.go
@@ -25,7 +25,7 @@ func TestWorkerAutoApproveSelfSendScreensInboundLeg(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if err := store.UpdateAgentProtection(ctx, ag.ID, owner.ID, identity.ProtectionConfig{
+	if _, err := store.UpdateAgentProtection(ctx, ag.ID, owner.ID, identity.ProtectionConfig{
 		InboundGatePolicy: "allowlist", InboundAllowlist: []string{"trusted@friend.example.com"}, InboundGateAction: "review",
 		InboundScanSensitivity: "off",
 		OutboundGatePolicy:     "open", OutboundGateAction: "flag", OutboundScanSensitivity: "off",
--- a/internal/httpapi/agents_write.go
+++ b/internal/httpapi/agents_write.go
@@ -163,13 +163,16 @@ func (s *Server) handleUpdateAgent(ctx context.Context, in *updateAgentInput) (*
 	if s.deps.UpdateAgentName == nil {
 		return nil, NewError(http.StatusInternalServerError, "internal_error", "update unavailable")
 	}
-	if err := s.deps.UpdateAgentName(ctx, ag.ID, ag.UserID, *in.Body.Name); err != nil {
+	// The store returns the authoritative post-update row, read INSIDE the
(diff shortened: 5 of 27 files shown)


jiashuoz (commented):
Team review at exact head 6e6a07b908be57ab393ea6e4bd41569606a2ddbc found the following issues.

[P2] Map the new persistence failures to generic 500s, not raw 400 validation errors.

`UpdateAgentName` now has Begin, Exec, joined read, and Commit failure paths (`internal/identity/store.go:1692-1713`), but `internal/httpapi/agents_write.go:171-174` turns every returned error into `400 invalid_request` with `err.Error()`. `UpdateAgentProtection` has the same regression (`internal/identity/protection.go:154-192`, `internal/httpapi/protection.go:241-244`). A timeout, pool outage, scan/schema failure, or ambiguous commit is a server failure; the current mapping suppresses retry/5xx telemetry and can expose pgx/Postgres details. Before this change, the post-write reload failure was generic 500. Please distinguish typed validation/not-found errors from operational errors, expose only the former as 4xx, and add dependency-error handler regressions for both paths.

[P2] Do not return success for a live-only mutation after the target has entered trash.

Both writes match only `(id,user_id)` and deliberately reload with `includeDeleted=true` (`internal/identity/store.go:1698-1708`, `internal/identity/protection.go:160-187`). If the handler resolves a live agent and a concurrent trash commits before the store call obtains the row, PATCH/PUT updates the trashed row. Name PATCH then returns 200 with `deleted_at`, even though `internal/httpapi/operations.go:68-72` says trashed agents appear only on trash/restore/permanent-delete surfaces; protection returns 200 for an inactive resource. No serial API outcome matches the name response: delete-first should make PATCH not found, while PATCH-first should return a live view. Add `deleted_at IS NULL`, return/map a typed not-found result, and cover a trash-before-write interleaving. The current tests queue only another rename/protection update.

[P2 test reliability] Make the webhook/template regression guards deterministic.

`post_write_read_test.go:331-356` and `392-417` release the holder and race the post-commit pool read against the competing writer without a barrier. The PR evidence itself reports that the reverted webhook/template implementations failed only 7/8 and 5/8 runs, so a single CI run can accept the old bug. Add a deterministic seam/barrier or structural assertion so the reverted implementation fails every run.

[P3] The all-writer timestamp invariant has only one writer under regression coverage.

The new test proves monotonicity for Upsert, while the invariant covers seven production `contact_engagements.updated_at` writers. A future reversion in activity, due-claim, review-release, or restore remains green. Please add table-driven coverage or a reliable drift guard for every writer.

[P3] `webhookColumns` is not yet the stated projection SSOT.

The comment in `internal/identity/webhooks.go:188-195` says every full-webhook read uses it, but `ListWebhooksByUser` and `ListEnabledWebhooksForRouting` still duplicate the projection/scanner at roughly 265-305 and 317-349. Reuse the shared projection/scanner so future fields cannot drift on list/routing paths.

Verification: all eight new DB race/timestamp tests passed three consecutive exact-head runs against isolated local Postgres; focused handler/spec tests passed; `git diff --check` passed; GitHub checks are green.

PR #789, opened 2026-08-01
feat(api): experimental quote_history flag on reply

jiashuoz:
## Summary

Adds an **experimental** opt-in `quote_history` boolean to `POST /v1/agents/{email}/messages/{id}/reply`. When `true`, the server composes the referenced message beneath the caller's body as mail-client-style quoted history — an `On <date>, <sender> wrote:` attribution line, the original text `>`-prefixed, and (when an `html` body is supplied) the original HTML in a blockquote. Default `false`: bodies remain byte-verbatim, today's behavior.

Why: API-sent replies currently look terse to human correspondents and give newly added thread participants no context. No peer email API offers this — they all require the caller to hand-roll the quote block, which is a long-standing complaint against the largest of them — so it is cheap differentiation for the agent-as-correspondent use case. The composition generalizes the existing forward quote path (`internal/outbound/forward.go`) into `internal/outbound/quote.go`; both share `ForwardContext` extraction.

Design notes:

- **Accept-time composition**: review holds, idempotency replay, and the stored outbound row all carry the final quoted content — a HITL reviewer approves exactly what ships (verified live).
- **Symmetric round trip**: our own inbound parser (`mailparse.stripQuotedReplies`) recognizes the emitted `On … wrote:` marker, so agent-to-agent mail with the flag on adds no noise to the receiving agent's model-facing `parsed.text`. Verified live against a local instance + Mailpit across a 9-scenario matrix: human recipient, deep-thread nesting (`>` → `>>`), text-only, HTML-only parent, reply-to-own-outbound, HITL hold/approve, idempotent retry, control.
- **Both alternatives carry the same history.** Only the body parts the caller supplies are quoted (a text-only reply stays text-only), and that no longer depends on the parent's MIME shape: a parent carrying only `text/html` is rendered to plain text (`mailparse.HTMLToText`) for the quoted text body, while its own markup is quoted verbatim into the HTML body. The parent's own `text/plain` part still wins when it has one — it is the lexically faithful copy.
- **Degrades silently by design.** A parent with no readable body (unparseable MIME, or markup that renders to nothing) sends the reply verbatim rather than emitting an attribution line over an empty quote block. The `202` is identical either way, so callers must not read it as proof that history was attached — stated in `docs/api.md`.
- Marked `Experimental` in the OpenAPI doc (house precedent: the `send_at` and `unsubscribe` beta markers); additive only, oasdiff-clean.

## Client surface checklist

- [x] Go handler + tests (`internal/httpapi/outbound_reply_quote_test.go`, `internal/outbound/quote_test.go`, `internal/mailparse/parse_test.go`)
- [x] OpenAPI spec + generated types refreshed
- [x] TypeScript SDK — generated base regenerated; `ReplyInput` flows the generated `quoteHistory` field per the `sendAt` precedent (no hand-written mapping layer exists for reply fields), unit + type tests added
- [x] Python SDK — generated base regenerated **and** `quote_history` kwarg on `reply` (sync mirror is automatic), 3 tests added
- [x] CLI — `--quote-history` on `e2a reply`, wired into the `checkFlags` allowlist + help (marked experimental), tests added
- [x] MCP tool — `quote_history` input on `reply_to_message` (plain optional, no default — no MCP field has one), mapping + schema-description tests added
- [x] Tests at each surface above (positive + negative/omission case per surface)
- [x] Contract scenario `reply_quote_history` in `tests/contract/scenarios.yaml` — proven through the Go runner; TS/Python runners skip it by established rule (uses `inject_message` store setup, same as `scheduled_reply_fields`)
- [x] Web dashboard checked manually: no reply composer exists in `web/`, so no dashboard change applies

> Migration row skipped — no schema change (composition is request-time; no new columns).

## Scope calls worth reviewing

- **`forward` keeps its existing behavior.** `BuildForwardBody` has the same text/HTML asymmetry this PR fixes on the reply path — forwarding a message that carries only `text/html` produces a text part with the header block and no body. Reusing the new helper there is a one-line change, but forward is GA, non-flagged surface, so changing its wire output belongs in its own PR rather than riding along behind an experimental flag.
- **The HTML-only branch is not in the shared contract scenario.** `inject_message` hardcodes the parent body and accepts only `agent_email` / `from` / `subject`, so covering it would mean extending the scenario language and all three interpreters in parity. That is a large change for a composition detail whose wire shape is pinned by unit tests (`internal/outbound`) and handler tests (`internal/httpapi`, against a bare `Content-Type: text/html` fixture). The existing scenario still proves the cross-SDK contract: the flag exists, composes at accept time, and stays off by default.
- **`maxExtractMIMEDepth` also hardens the pre-existing forward path**, so reverting the flag does not revert that bound.

## Operational risk

Low. Opt-in flag, default off; no schema change; no new external calls. The composed body passes through the existing outbound size ceilings (`413 payload_too_large`, scope `composed_message`, unchanged) — note that the quoted parent counts toward the 10 MiB ceiling, so a small reply to a very large parent can now `413`. Held replies show reviewers the final composed content — behavior reviewers should be told about, not a risk per se. Rollback = revert; no data migration to unwind.

Known caveats (why it is marked experimental):

- **Read-back is lossy.** The message detail view surfaces the quote-stripped `parsed.text` for outbound rows too, so an agent that sends with `quote_history: true` and then `GET`s its own message sees the body without the quote block. `raw_message` remains the canonical copy, and the exact wire bytes are pinned by unit tests.
- **Attribution residue on a Date-less parent.** `mailparse.quoteMarker` matches `On <date> … wrote:` but not a bare `<sender> wrote:`, so when the parent has no `Date` header the attribution line survives into the receiving agent's `parsed.text` (the contract scenario pins exactly this residue). Real mail almost always carries a `Date`. Loosening the parser pattern would false-positive on human prose, so this is left as-is.
- **Inline images in the quoted parent do not resolve.** Unlike forward, a reply does not carry the parent's attachments, so a quoted parent whose HTML references `cid:` parts renders those as broken images.
- **Encoded-word display names render raw** (`=?UTF-8?Q?…?=`) in the attribution — the deliberate lexical-fidelity choice documented on `ForwardContext`, inherited from forward.
- **Gmail quote-collapse** verified structurally (Mailpit) but not yet against a live Gmail account — worth one real send before promoting past experimental.

## Test plan

- [x] `make test-unit`, `make spec-check`, `make openapi-compat-check` (additive, no breaking changes), and the repository text-integrity check all clean
- [x] Go contract runner: full suite green, including `TestScenarios/reply_quote_history` (Postgres-backed)
- [x] TypeScript SDK 242 passed + type tests; CLI 285 passed; MCP 311 passed; Python SDK client suite 201 passed
- [x] Live local matrix (server + Mailpit): 9 scenarios incl. HITL hold→approve and agent-to-agent round-trip strip
- [x] Rebased onto current `main` and re-verified: the conflicts were `internal/httpapi/outbound.go` (multi-address `reply_to`, and the rewritten `send_at` doc now that scheduling survives a review hold), `cli/src/commands/send.ts` (same `reply_to` change), and an append collision in `tests/contract/scenarios.yaml`
- [ ] Post-merge: one real send to a Gmail inbox to confirm quote-collapse rendering


jiashuoz:
Adversarial review round completed before handing off (findings + fixes in eaef777a):
- **Contract scenario reworked to the scheduled-send pattern** — the original version would have been the suite's first immediate provider submission, sending real mail to alice@example.net on every staging seed run and risking suppression-list poisoning; it also raced cleanup DELETE against the send worker's lease (409 send_in_progress flake). Both replies now use a future `send_at` (composition happens at accept, so the assertions are unchanged) and trash-cancel cleanly.
- **MIME depth bound added to `extractBodyParts`** (mirrors `mailparse.maxMIMEDepth`) — pre-existing unbounded recursion on the forward path, newly hot via reply quoting; now cut at 32 levels with a nested-200-levels regression test.
- **Docs**: noted the quoted parent counts against the 10 MiB composed ceiling (413 `payload_too_large`, scope `composed_message` — verified enforced post-composition).
- **CLI test isolation**: the new `--quote-history` test no longer depends on a prior test's exitCode reset.

Cleared during review (verified, no change needed): CRLF/header injection via parent headers (net/mail unfolds; QP encoding canonicalizes), HTML injection (same contract as forward's blockquote), idempotency (dedup hashes raw request bytes, so key reuse with a different flag value correctly 422s), scenario determinism (6× green).

jiashuoz (commented):
**Review verdict: approve-worthy.** Read the full diff at eaef777a and verified the load-bearing claims against the surrounding code:

- **Accept-time composition holds.** The quote rewrite in `handleReply` (`internal/httpapi/outbound.go:578`) runs before `s.deliver`, and `composedMessageSizeError` fires inside `deliver` on the post-quote body — so the documented 413 `composed_message` behavior genuinely applies to the quoted total at accept time, and HITL holds / idempotency replay carry the final composed content.
- **TS SDK wiring is automatic** — `ReplyInput = Omit<ReplyRequest, "unsubscribe"> & {…}` picks up `quoteHistory` from the regenerated model with no `client.ts` change; wire tests pin both the on and off shapes.
- **Python kwarg is copy-on-write** when `body` is already a `ReplyRequest`, with a regression test pinning that the caller's model isn't mutated.
- **Contract scenario is safe** — future `send_at` means no provider submission and trash-cancel can't race a lease, matching the `scheduled_*` pattern.
- **`maxExtractMIMEDepth`** closes the deep-nesting DoS on the pre-existing forward path too, not just the new reply path.
- **Stability marking is convention-consistent** — prose "Experimental" per-field, same as `send_at`/`unsubscribe` (`x-stability-level` is schema-granular).

Minor, non-blocking observations:

1. **"Zero noise" is only exact for dated parents.** `mailparse.quoteMarker` matches `On .+ wrote:` but not a bare `<from> wrote:`, so a Date-less parent's attribution line survives into the receiving agent's `parsed.text` (the contract scenario pins exactly this residue). Real mail almost always has a Date — suggest softening the PR-body claim rather than touching the parser (a bare `.+ wrote:$` pattern would false-positive on human text).
2. **Encoded-word display names render raw** (`=?UTF-8?Q?…?=`) in the attribution — the deliberate lexical-fidelity choice documented on `ForwardContext`, inherited from forward. Consistent; just inherited surface area.
3. **Parent HTML is embedded verbatim** in the blockquote — same contract as forward and standard mail-client behavior, but it does round-trip the original sender's markup (links, pixels) into the thread.
4. **Web dashboard surface unchecked.** Per `AGENTS.md`, `web/` must be manually considered on every API change since the PR template doesn't list it. Not exposing an experimental flag in the dashboard composer is defensible — but worth stating as a deliberate decision.

Also verified the round-trip claim empirically against a live instance: a Gmail reply's quoted history is stripped from the model-facing `parsed.text` at the `On … wrote:` marker and retained verbatim in `parsed.html` — exactly the format this PR emits, so agent-to-agent mail with `quote_history: true` arrives clean at the receiving agent. (One caveat confirmed in the same test: the inbound stripper cuts at any line *starting* with `>`, so the emitted quote format must keep its lines `>`-prefixed — which `BuildReplyQuoteBody` does.)

PR #772, opened 2026-07-31
release: prepare v1.5.0 — client version bumps, changelogs, beta docs

jiashuoz:
## Summary

Release-prep for the additive **v1.5.0** product release (contacts/outreach beta, scheduled sending beta, webhook delivery observability, thread identity beta, MIME/DKIM hardening). Version bumps + changelogs + agent-guidance and public-doc updates only — no code changes. (No API surface changes, so the client-surface checklist is omitted.)

**Update (per Josh's direction):** the core `/v1` surface is now presented as **stable/GA with the beta set explicitly enumerated**, grounded in the spec's machine-readable `x-stability-level` markers. Verified against `api/openapi.yaml` in this tree: **43 GA operations, 29 beta operations** (note: the templates+starter-templates beta group is 8 ops, not 9 — the 29 total holds). Changes in the second commit (`1eeb187c`):
- `docs/api.md` — new prominent **"Stability: GA and beta surface"** section near the top: GA/beta matrix by resource group, the beta-fields-on-GA-operations list (scheduled sending, `thread_id`, template refs, managed unsubscribe, review-hold projections, export interior schemas), and the beta events, cross-referencing `x-stability-level` as the source of truth and the existing "Beta operations" table.
- `README.md` — the stale "release candidates / GA planned by July 31, 2026" banner replaced with the GA statement + enumerated beta list, linking the matrix.
- `AGENTS.md` — the same one-sentence release-candidate claim fixed (it would otherwise keep telling agents the API is unstable).
- Changelogs (TS 5.5.0 / Py 5.5.0 / CLI 2.2.0) — one-line pointer to the stability matrix.
- `cli/README.md` — `e2a contacts` marked beta (MCP README already was).

## Version bumps

| Package | From | To | Where |
|---|---|---|---|
| `@e2a/sdk` (TS) | 5.4.0 | 5.5.0 | `sdks/typescript/package.json` |
| `e2a` (PyPI) | 5.4.0 | 5.5.0 | `sdks/python/pyproject.toml` |
| `@e2a/cli` | 2.1.0 | 2.2.0 | `cli/package.json` |
| e2a plugin | 0.5.0 | 0.6.0 | `.claude-plugin`/`.codex-plugin`/`.cursor-plugin` plugin.json + `.claude-plugin`/`.cursor-plugin` marketplace.json |

[...]

PR #791, opened 2026-08-01
feat: suppressions across web, CLI, and MCP surfaces

jiashuoz:
## Summary

The suppression system (account-level bounce/complaint blocks + beta per-agent unsubscribe/manual blocks) was fully implemented server-side and in both SDKs, but invisible in the dashboard and absent from the CLI and MCP server. This closes the parity gap on all three surfaces. **No wire-API changes** — no Go, no OpenAPI, no migrations, no generated-SDK churn.

- **Web — per-inbox Suppressions tab** (Inbox · Outreach · **Suppressions** · Trash · Settings): lists the agent's unsubscribe/manual blocks, adds manual blocks, removes behind an explicit confirm. Tab strip now scrolls horizontally on narrow screens.
- **Web — `/contacts/suppressions`**: the account-wide bounce/complaint list (address, source, reason, source message, date) with a remove action whose confirm states the sender-reputation consequence. Both contacts pages share a Contacts · Suppressions view switcher.
- **CLI — `e2a suppressions list/add/remove`**: routes on `--agent` (without → account list; with → per-agent). `add` is agent-only by design and says why.
- **MCP — five admin-tier tools**: `list_suppressions`, `delete_suppression`, `list_agent_suppressions`, `create_agent_suppression`, `delete_agent_suppression` (tool count 71 → 76; frozen name baseline updated additively).
- **TS SDK (additive)**: `account.suppressions.list({ limit })` gains the page-size param its sibling resources already had.

Design doc: `docs/superpowers/specs/2026-08-01-suppressions-surfaces-design.md`.

## Client surface checklist

- [x] Go handler — no change (endpoints already exist)
- [x] OpenAPI spec — no change (verified zero diff)
- [x] TS SDK — additive `{ limit }` param + changelog entry
- [x] Python SDK — no change (already covered)
- [x] CLI — new command group + tests
- [x] MCP tool — five new tools + tests
- [x] Web dashboard — two new views + tests
- [x] Contract scenarios — n/a (no `/v1` surface change)

## Testing

- Unit/suites: SDK 233 ✓, MCP 296 ✓, CLI 273 ✓, web Jest 697 ✓, web lint + static-export build ✓ (routes `/contacts/suppressions` and `/inboxes/suppressions` present).
- Over-the-wire e2e against the local docker stack: full agent-suppression CRUD incl. cursor pagination, percent-encoded DELETE, invalid-address 400, missing-confirm 422, double-delete 404; account list + 404 remove; auth invariants (401 no/bogus key, **403 agent-scoped key on all five endpoints**); real CLI binary round-trip; real MCP HTTP server round-trip for all five tools incl. typed `not_found` error. Throwaway account deleted afterwards.


jiashuoz:
## Review pass applied (688cb64b)

Two independent reviews (standard + adversarial) ran against this branch. **No blockers**; both converged on the same class — error paths that left a *safety* list reading as "nothing is blocked". Everything real is fixed with a regression test; dispositions below.

### Fixed

| Finding | Fix |
|---|---|
| **Dot-segment path collapse** (adversarial, medium) — `encodeURIComponent` doesn't encode `.`, so an address of `..` survives into the path and the URL parser collapses `DELETE .../suppressions/..` onto the **parent** resource (`/v1/account` or `/v1/agents/{email}`) — both real DELETE endpoints taking the same `?confirm=DELETE` token. Latent only (server rejects such addresses on create; chi doesn't redirect the collapsed path) but the guard was a routing-library default. | Address validated at every client boundary before a request is built (`contacts/_lib/suppressionPath.ts`, `checkedAddress` in the CLI). Live repro `e2a suppressions remove '..'` now exits 2 with account + agent intact. |
| **Empty state under a failed load** (adversarial) — "No suppressed recipients" rendered beneath the error banner on a 401/501. | `listFailed` tracked separately from `error`; only a failed LIST suppresses the empty state (a failed remove leaves the list trustworthy). |
| **No reconverge on failed remove** (both reviewers) | Failed delete refetches, then re-sets the action error. |
| **Double-click = two DELETEs**, **duplicate React keys on append** | One delete in flight at a time; page appends deduped by address. |
| **Stale tool counts** (independent) — `mcp/README.md` inner paragraph plus the three plugin manifests and the e2a skill still said 71/51. | All corrected to 76/56. |
| **Untested tab wiring** (independent) — the design named `detectTab` ordering as the top regression risk. | Test pins the tab slug ↔ `detectTab` contract in `AgentHeader.test.tsx`. |
| `--json` missing from suppressions usage strings; stale AgentHeader "two tabs" comment | Both corrected. |

### Refuted by the adversarial pass (recorded, no change)

XSS/output escaping (payloads in `reason`/`address`/`source` render as literal text — no `dangerouslySetInnerHTML` anywhere), URL injection for all non-dot metacharacters (`/ ? # % @` and unicode all encode to a single segment), confirm bypass, MCP tier gating (double-enforced: registration-time gating plus account-scope checks in every handler), SDK/MCP pagination, CLI argument parsing across 23 malformed invocations, tab/route collisions, frozen-contract regressions, and the data policy.

### Deliberately not changed

"Load more" resetting on Refresh, `Number()` leniency in `--limit`, and the missing-`?email=` loading state — all byte-identical to the existing sibling pages/commands. Worth one pass across all pages rather than a divergence here.

### Re-verification (fresh, post-fix)

Web Jest 708 ✓ · CLI 274 ✓ · MCP 296 ✓ · SDK 233 ✓ · web lint + static export ✓ · CLI build ✓. Live re-run against the local stack: the adversarial repro blocked, normal add/list/remove round trip unaffected, server logs clean, throwaway account deleted.

jiashuoz:
Follow-up filed as #792 — the dot-segment path-collapse class this PR's guards address at two call sites, tracked for a durable fix (SDK middleware guard + a router regression test pinning the trailing-slash 404 that is currently the load-bearing defense). This PR remains scoped to the suppression surfaces.

jiashuoz:
Closed four release-hygiene gaps ahead of v1.5.0 (commits c98c75da, 7eb3c926):

1. **Python SDK parity** — `account.suppressions.list(limit=...)` now takes the same keyword-only page-size param the TS side gained here (the generated `list_suppressions` already accepted `limit`; no regeneration needed), plus a wire-level test that the param reaches every page.
2. **TS changelog** — the `Unreleased` section folded into 5.5.0 (5.5.0 is unpublished, this ships inside it).
3. **Python changelog** — matching 5.5.0 entry.
4. **CLI changelog** — the missing `e2a suppressions` entry in 2.2.0, derived from the implementation, with the #772 stability split stated (account-wide GA, agent-scoped beta).

Verified: Python suite 510 passed (94% cov, gate 90%), `mypy` clean; TS SDK build + 233 tests + type-tests green; CLI build + 276 tests green; `make spec-check` and `make generate-sdk-check` clean.

PR #806, opened 2026-08-03
fix(identity): populate agent_count in LookupDomain

### LINKED ISSUE #632: getDomain always returns agent_count: 0 (only listDomains populates it)
## Summary

`GET /v1/domains/{domain}` always returns `agent_count: 0`, regardless of how many agents are on the domain. `GET /v1/domains` returns the correct value for the same domains.

## Reproduction

Against production, same account, same moment:

```
GET /v1/domains            (list)          GET /v1/domains/{domain}  (single)
  sked.mnexa.ai   → 2                        sked.mnexa.ai   → 0
  team.mnexa.ai   → 5                        team.mnexa.ai   → 0
  inbox.mnexa.ai  → 1
  agentdrive.run  → 1
```

## Cause

`Domain.AgentCount` is only ever populated by `ListDomainsByUser`. The struct comment says so outright:

```go
// internal/identity/store.go:54
// AgentCount is computed at read time by ListDomainsByUser and is
// ...
AgentCount int `json:"agent_count"`
```

It's computed inline via a correlated subquery in the list query (`store.go:1163`, scanned at `:1203`). `LookupDomain` — the single-resource path — never selects it, so the field stays at Go's zero value. `internal/httpapi/domains.go:173` then maps `AgentCount: d.AgentCount` unconditionally, serializing that `0` as if it were a real count.

So the value isn't "missing" in a way a client could detect — it's an authoritative-looking `0`.

## Impact

`get_domain` is the **documented poll target** after `verify_domain` (per the MCP tool description: *"this is the poll target after `verify_domain`"*). So the endpoint users are explicitly directed to during domain onboarding reports a permanently wrong agent count.

Concretely misleading in at least two flows:
- Checking whether a domain still has agents before attempting `delete_domain` — which fails with `domain_has_agents`. A user reading `agent_count: 0` would reasonably conclude the domain is empty and be confused by the rejection.
- Confirming agents were created on a freshly verified domain.

## Suggested fix

Either populate `AgentCount` in `LookupDomain` with the same correlated subquery the list path uses, or — if the extra join isn't wanted on the hot single-resource path — omit the field from that response rather than emitting a false zero. Silently returning `0` is the worst of the three options.

Found while re-registering domains on a production account.

AmirF194:
## Summary

`GET /v1/domains/{domain}` (`LookupDomain`) always reports `agent_count: 0`, while `GET /v1/domains` (`ListDomainsByUser`) reports the correct value for the same domain. `LookupDomain`'s `SELECT`/`Scan` never included the correlated subquery `ListDomainsByUser` uses to compute `AgentCount`, so the field stayed at Go's zero value and was serialized as an authoritative `0` regardless of how many agents actually exist.

`get_domain` is the documented poll target after `verify_domain`, so this misled the two flows that read it: checking whether a domain still has agents before `delete_domain` (which rejects with `domain_has_agents`), and confirming agents were created on a freshly verified domain.

## Fix

`LookupDomain` now runs the same correlated subquery `ListDomainsByUser` uses (trashed agents excluded), so the single-resource and list responses agree. No API shape change: `agent_count` is an existing GA field, only its value on this path was wrong. Checked every production reader of `Domain.AgentCount`: only `domainViewWithRamp` (used by `GET /v1/domains/{domain}`) reads it, so the other two `LookupDomain` call sites are unaffected.

## Verification

- New regression test `TestLookupDomain_AgentCountMatchesList` (`internal/identity/store_test.go`): fails on `main` (`AgentCount = 0, want 1`), passes on this branch, both run in Docker against a real Postgres 16 instance.
- `go test ./...` passes in full (every package), plus `gofmt` and `go vet` clean.
- Did not check the live production behavior described in the issue: no access to a production account, verification is against a fresh local database.

Fixes #632

### DIFF
--- a/internal/identity/store.go
+++ b/internal/identity/store.go
@@ -52,11 +52,9 @@ type Domain struct {
 	// "probed and failed" which is captured by `verified=false` + a
 	// non-null LastCheckedAt.
 	LastCheckedAt *time.Time `json:"last_checked_at,omitempty"`
-	// AgentCount is computed at read time by ListDomainsByUser and is
-	// not a persisted column. Single-domain LookupDomain leaves it at
-	// the zero value — callers that need the count call the list path
-	// (this column-versus-aggregate split avoids changing every store
-	// signature to thread an agent-counter through).
+	// AgentCount is computed at read time by ListDomainsByUser and
+	// LookupDomain, via the same correlated subquery, and is not a
+	// persisted column.
 	AgentCount int `json:"agent_count"`
 	// DKIM keypair fields. The selector + public key
 	// are user-facing — the dashboard shows them so users can copy the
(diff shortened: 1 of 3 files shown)


jiashuoz (dismissed):
Approving. Diagnosis and fix are both right: `LookupDomain` now runs the same correlated subquery as `ListDomainsByUser`, trashed exclusion included, so the two paths can't disagree. Verified locally against Postgres — the new store test fails on the pre-fix code (`AgentCount = 0, want 1`) and passes with it; `internal/identity` and `internal/httpapi` are green.

I pushed one commit to the branch: `domain_agent_count` in `tests/contract/scenarios.yaml`. Changing an observable `/v1` response value needs a shared scenario proven through the three live-server runners, and I made it a new scenario rather than steps on `domain_crud` because that one is in the TS runner's `ACCOUNT_GLOBAL` skip set — a pin added there wouldn't actually run in TypeScript. It passes in Go/TS/Python and fails in all three with the store fix reverted (`get_domain_counts_agent: agent_count = 0, want 1`).

Two notes, neither blocking:

**1. Sibling path still emits the zero.** `domainViewWithRamp` has three callers — `handleListDomains` (`domains.go:429`), `handleGetDomain` (`:457`), and `handleRegisterDomain` (`:508`). The third is fed by `ClaimOrCreateDomain`, which selects the row without the count. Since a same-owner claim is idempotent (the cross-account guard is `user_id IS DISTINCT FROM $2`), re-POSTing a domain you already own returns `201` with `agent_count: 0` even when it has agents. Narrower than #632 — only re-registration, since a fresh registration's `0` is correct — so a follow-up rather than scope creep here. Filing separately.

**2. Commit message claim.** It says *"Checked every production reader of `Domain.AgentCount`: only `domainViewWithRamp` (used by `GET /v1/domains/{domain}`) reads it."* Per the above that's three call sites, not one. Worth correcting if this gets squashed on merge.

Root cause of the class, for whoever picks up the follow-up: `AgentCount` isn't a column, so every query that builds a `Domain` has to remember the subquery, and the Go zero value fills the gap indistinguishably from a real answer. Making it `*int` — absent rather than `0` — would kill the class instead of the instance.

Thanks for the clean, well-scoped fix and the thorough writeup.

jiashuoz:
Filed the sibling-path follow-up from note 1 as #811 (register path emits the false zero on re-registration of an owned domain), so this PR stays scoped to #632.

jiashuoz (approved):
Re-approving after the cross-reference fix (the Copilot nit was right — the sibling test is in `domain_trashed_agents_test.go`). No functional change since the previous approval.

AmirF194:
Thanks for keeping this scoped and filing #811 separately, that follow-up looks like a real sibling case worth a look.

jiashuoz:
@AmirF194 Thanks for the PR! Curious how you found this repo? Are you planning to deploy an e2a instance for your organization? I appreciate any feedback on the product.

AmirF194:
Thanks for merging it.

I run [FastInfer](https://fastinfer.org), where we build and ship AI products, so I keep an eye on repos in the agent-infrastructure space. e2a came up through the `ai-agents` and `human-in-the-loop` topics and it stood out: an authenticated inbound identity plus an approval gate before an agent sends anything is the piece most agent stacks are missing, and it is the piece that decides whether you can leave one running unattended. That is why I picked a fault in it to fix rather than just starring it.

Nothing to report on deployment yet. If we do put an agent behind a real inbox, this is the first thing I would try, and you will get concrete feedback then rather than guesses now.

One thing from the fix that might be worth more than the fix: #632 is a shape, not a one-off. `agent_count` is derived in the list query and was missing from the single-resource one, and #811 is that same shape on the register path. Pinning every single-resource read against its list entry, the way the contract scenario here does, would find the rest cheaply.

PR #817, opened 2026-08-03
fix(identity): populate agent_count on same-owner domain re-claim

### LINKED ISSUE #811: POST /v1/domains returns agent_count: 0 when re-registering an owned domain that has agents
## Summary

`POST /v1/domains` returns `agent_count: 0` for a domain that has agents, when the request re-registers a domain the caller already owns.

This is the residual of #632 (fixed for `GET /v1/domains/{domain}` in #806) on a sibling endpoint.

## Cause

`AgentCount` is not a column — it is computed per-query by a correlated subquery, so every query that builds an `identity.Domain` has to remember to include it. Five sites build one; three omit it:

| builder | includes the count? |
|---|---|
| `ListDomainsByUser` (`store.go:1306`) | yes |
| `LookupDomain` (`store.go:1097`) | yes, as of #806 |
| `ClaimOrCreateDomain` — SELECT existing row (`store.go:863`) | **no** |
| `ClaimOrCreateDomain` — INSERT … RETURNING (`store.go:868`) | no (correct: brand-new row, the count really is 0) |
| `AdoptSharedDomain`, `LookupCoveringDomain` | no (never serialized to a client) |

`handleRegisterDomain` (`httpapi/domains.go:508`) renders `ClaimDomain`'s result through the same `domainViewWithRamp` helper the other two domain endpoints use, so the unpopulated field is serialized as an authoritative `0`.

A same-owner claim is idempotent — the cross-account guard is `user_id IS DISTINCT FROM $2`, so re-registering your own domain does not 409, it returns the existing row. That row can have any number of agents on it.

## Reproduction

Against a local server, with `example.test` already registered and holding one agent:

```
POST /v1/domains {"domain": "example.test"}   → 201, agent_count: 0
GET  /v1/domains/example.test                 → 200, agent_count: 1
```

Confirmed at the store layer: `ClaimOrCreateDomain` returns `AgentCount=0` where `LookupDomain` returns `1` for the same domain and user.

## Impact

Narrower than #632, which was wrong on every call to the endpoint: here a first-time registration's `0` is genuinely correct, and only re-registration is wrong. But #632 was reported by someone re-registering domains, so this is a path users do hit — and the wrong value is once again indistinguishable from a true zero.

## Suggested fix

Two options, in preference order:

1. **Kill the class.** Make `Domain.AgentCount` a `*int`. An unpopulated count then serializes as absent (or explicitly `null`) rather than as `0`, so a query that forgets the subquery produces a visibly missing value instead of a plausible wrong one. This is a `/v1` response-shape change on a required GA field, so it needs the compat gate's opinion — likely a no, in which case:
2. **Populate it on the claim path**, matching what #806 did for `LookupDomain`: add the same correlated subquery to `ClaimOrCreateDomain`'s existing-row SELECT only. The INSERT branch is correct as-is.

Either way the contract scenario added in #806 (`domain_agent_count` in `tests/contract/scenarios.yaml`) is the place to extend coverage — a re-register step asserting the count survives.

hyperSuperCube:
## Summary

`POST /v1/domains` returns `agent_count: 0` when re-registering a domain the caller already owns, even when the domain has live agents. A same-owner claim is idempotent (the cross-account guard is `user_id IS DISTINCT FROM $2`), so the endpoint returns the *existing* row — but `ClaimOrCreateDomain`'s existing-row `SELECT` never computed `AgentCount`, and `handleRegisterDomain` renders the result through the same `domainViewWithRamp` as `GET /v1/domains/{domain}`, so the Go zero value was serialized as an authoritative `0`.

This is the residual of #632 (fixed for `LookupDomain` in #806) on the register path, filed as #811.

## Fix

The existing-row branch of `ClaimOrCreateDomain` now runs the same correlated subquery `LookupDomain` and `ListDomainsByUser` use (trashed agents excluded), so all three endpoints that serialize a domain agree. The `INSERT … RETURNING` branch is deliberately untouched: a brand-new row cannot have agents, so its zero is genuinely correct.

No API shape change: `agent_count` is an existing required GA field; only its value on this path was wrong. (#811's option 1 — making the field `*int` so a forgotten subquery serializes as absent — would kill the class, but it drops a required field from the GA `/v1` shape, which the compat gate exists to prevent, so this PR takes option 2.)

## Verification

All against a real Postgres 16 instance in Docker:

- New store regression test `TestClaimOrCreateDomain_ReclaimPopulatesAgentCount` (`internal/identity/store_test.go`): fails pre-fix (`re-claim AgentCount = 0, want 1`), passes with the fix. Also pins the fresh-claim zero and agreement with `LookupDomain`.
- New `reregister_counts_agent` step on the `domain_agent_count` contract scenario (`tests/contract/scenarios.yaml`) — the extension point #811 names. Fails pre-fix (`agent_count = 0 (float64), want 1`), passes with the fix, via the Go runner; the step is runner-agnostic YAML with resource-scoped assertions, matching the scenario's existing constraints.
- `internal/identity`, `internal/httpapi`, `tests/contract` pass in full, plus `internal/sendramp` / `internal/apiserver` / `internal/contactdue` (whose tests call `ClaimOrCreateDomain` as setup). `gofmt` and `go vet` clean.
- TS/Python contract runners not run locally (need a wired live server); the Go runner passes and CI runs all three.

Fixes #811

### DIFF
--- a/internal/identity/store.go
+++ b/internal/identity/store.go
@@ -52,9 +52,9 @@ type Domain struct {
 	// "probed and failed" which is captured by `verified=false` + a
 	// non-null LastCheckedAt.
 	LastCheckedAt *time.Time `json:"last_checked_at,omitempty"`
-	// AgentCount is computed at read time by ListDomainsByUser and
-	// LookupDomain, via the same correlated subquery, and is not a
-	// persisted column.
+	// AgentCount is computed at read time — ListDomainsByUser,
+	// LookupDomain, and ClaimOrCreateDomain's existing-row branch run
+	// the same correlated subquery — and is not a persisted column.
 	AgentCount int `json:"agent_count"`
 	// DKIM keypair fields. The selector + public key
 	// are user-facing — the dashboard shows them so users can copy the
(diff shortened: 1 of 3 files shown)


jiashuoz (approved):
Verified locally against real Postgres before approving:

- `TestClaimOrCreateDomain_ReclaimPopulatesAgentCount` fails pre-fix (`re-claim AgentCount = 0, want 1`) and passes with the fix.
- The contract step fails pre-fix (`step reregister_counts_agent: path "agent_count" = 0 (float64), want 1`) and passes with it.
- `internal/identity`, `internal/httpapi`, and `tests/contract` all pass in full.

The fix is minimal and the reasoning holds on both branches: the existing-row subquery is byte-identical to `LookupDomain`/`ListDomainsByUser`, and the untouched INSERT branch's zero is correct structurally, not just conventionally — `agent_identities.registered_domain` is `REFERENCES domains(domain) ON DELETE NO ACTION` and `DeleteDomainTx` maps the FK violation to `ErrDomainHasAgents`, so no agent row can outlive its domain row and greet a fresh claim. Checked the remaining domain-returning store methods too: `LookupCoveringDomain` is only read internally and `AdoptSharedDomain` is prober-only, so list/get/register really is the full set of serializing paths.

Pushed one commit on top pinning `verified: true` alongside the count on the re-register step — idempotent means the existing row returned untouched, and a future rewrite into `INSERT … ON CONFLICT DO UPDATE` would reset verification while the count assertion still passed.

Follow-ups worth their own PRs, none blocking:
1. The count subquery is now hand-written in three places; #632 and #811 are the same bug recurring per new SELECT. Hoisting the column list + subquery into shared consts would make it structurally unforgettable.
2. `handleRegisterDomain` runs `EnforceDomainCreate` before `ClaimDomain`, so an idempotent re-claim is charged against `max_domains` and 402s at cap despite creating nothing.
3. The idempotent re-claim semantics (201 + existing row rather than 409) are now pinned in contract tests but absent from the OpenAPI description.

PR #820, opened 2026-08-04
fix(web): surface contact metadata in the dashboard

### LINKED ISSUE #814: fix(web): contact metadata is stored but never displayed
## Problem

The contacts dashboard writes `metadata` and then gives you no way to read it back. The API stores and returns it correctly — the UI is the gap.

**Nothing renders it.** `metadata` is declared on the Contact type (`web/src/app/(app)/contacts/page.tsx:15`) and arrives populated from the API, but appears in no JSX anywhere in the page. The list, the detail view, and the edit form all ignore it. The edit form's PATCH body sends only `display_name` (`page.tsx:447`), so it can't be inspected or corrected there either.

**And the CSV importer fills it silently.** `mapCsvRows` sweeps every column that isn't the chosen email or name column into `metadata`:

```ts
// web/src/app/(app)/contacts/_lib/csv.ts:62-68
const metadata: Record<string, string> = {};
headers.forEach((header, index) => {
  if (header && index !== emailIndex && index !== nameIndex) {
    metadata[header] = values[index] ?? "";
  }
});
if (Object.keys(metadata).length > 0) row.metadata = metadata;
```

## Why this matters

Upload a CSV with `email,name,company,role,notes` and pick the email and name columns. The other three are accepted, stored, and become permanently invisible. The import reports success, so it looks like it worked — the data is simply unreachable from the surface that ingested it.

This is worse than not supporting metadata: a UI that rejected the extra columns, or ignored them explicitly, would at least tell the truth. Today the dashboard is a one-way door for arbitrary user data.

MCP and REST callers can still read metadata via `list_contacts` / `get_contact`, so the data isn't lost — but a user who never leaves the dashboard has no way to know that.

## Proposed fix

1. **Render metadata in the contacts table or detail view** — a key/value list on the row's expanded state is probably enough. Keys are caller-owned and unbounded, so cap what's shown inline and keep the full set behind a disclosure.
2. **Make the CSV importer's mapping visible before import.** The preview step should say which columns become metadata, so "everything else is retained" is a choice rather than a surprise.
3. **Consider allowing metadata edits** in the edit form, or state plainly that it's import-managed. Read-only is a defensible answer; invisible is not.

## Surfaces

- [ ] `web/src/app/(app)/contacts/page.tsx` — render metadata; extend the edit form or document read-only
- [ ] `web/src/app/(app)/contacts/_lib/csv.ts` — surface the column mapping in the import preview
- [ ] Web tests (`page.test.tsx`, `csv.test.ts`) — current fixtures all use `metadata: {}`, which is why this wasn't caught; add non-empty metadata cases

No API, SDK, CLI, or MCP change needed — the contract already carries `metadata` on read and write.

## Note

The web dashboard is not on `.github/pull_request_template.md`'s "Client surface checklist", which is the likely reason this drifted while the other surfaces kept parity. Worth considering as a separate follow-up.

hyj28:
## Summary

The contacts dashboard stored `metadata` and gave the user no way to read it back, and the CSV
importer filled it silently. Uploading `email,name,company,role,notes` and picking the email and
name columns accepted the other three, stored them, reported success, and made them permanently
unreachable from the surface that ingested them. This renders metadata everywhere the dashboard
reads a contact, and states the CSV column mapping before the user commits to the import.

Fixes #814. Web-only — the `/v1` contract already carries `metadata` on list, get, create and
PATCH, so no API, SDK, CLI or MCP change was needed.

**What changed**

- Metadata renders on the mobile card list, the desktop table, and read-only in the edit form.
  Inline display is capped at 3 fields and 80 characters per value; the full set sits behind a
  disclosure that reports how many more fields exist. Empty metadata renders no affordance, so
  the page stays as uncluttered as it is today for contacts that have none.
- Non-string values are formatted deliberately — `null` as `null`, numbers and booleans via
  `String`, objects and arrays as JSON, circular structures as a placeholder — instead of
  leaking `[object Object]`. The API documents metadata as flat but does not enforce scalar
  values, so this is reachable input.
- The CSV import preview names the columns that will become metadata and shows their first-row
  sample values, and says plainly when no columns will be stored as metadata rather than
  rendering an empty widget.
- `getCsvMetadataColumns` was added alongside `mapCsvRows` and shares its column resolution, so
  what the preview reports and what the import posts cannot drift — including for duplicate
  headers, blank headers, and headers differing only by whitespace.

**Metadata is read-only in the dashboard**, per the issue's "read-only is a defensible answer;
invisible is not". The edit form displays it and says so; PATCH still sends `display_name` only,
which the server treats as leaving metadata unchanged. There is a test asserting the PATCH body
has no `metadata` key, because a wholesale replace here would destroy what an import wrote.

Three pre-existing bugs surfaced while wiring this up and are fixed:

- The import panel's column auto-detection didn't trim headers, so a header like `" email "`
  set `emailColumn` to a string `mapCsvRows` could never resolve (`No " email " column found`).
- The column `<select>`s had no explicit `value`, so `HTMLOptionElement.value` fell back to
  whitespace-collapsed text — a header with an internal double space dispatched a value that
  `headers.indexOf` couldn't find.
- `preview()` bailed on a falsy email column, so re-selecting a blank header left `rows` mapped
  under the previously selected column.

## Client surface checklist

Not applicable — this is a web-dashboard-only change. The `/v1` contract already carries
`metadata` on read and write (`ContactView.Metadata`, `CreateContactRequest.Metadata`,
`UpdateContactRequest.Metadata`), and no handler, spec, SDK, CLI or MCP surface is touched. The
diff is four files, all under `web/src/app/(app)/contacts/`.

Worth noting the issue's closing point: the web dashboard isn't on this checklist, which is the
likely reason this drifted while the other surfaces kept parity. That's a separate follow-up.

## Operational risk

Low. Client-side rendering only; no new network calls, no new writes, no schema or auth change.

- No write path was added for metadata, so this cannot corrupt or erase stored metadata. PATCH
  continues to omit the field.
- The import request body is byte-identical to before — the `contacts` array is still exactly
  what `mapCsvRows` returns, asserted in a test.
- Rendering is bounded: metadata can legitimately carry the server's published maximum of 50
  keys with 4 KB values, so inline display is capped and the disclosure body is
  `max-h-64 overflow-auto`. A pathological contact degrades the row's height, not the page.
- Rollback is a revert; nothing persists.

## Test plan

Ran from `web/`. Baseline before this branch was `lint` exit 0, `jest` exit 0 with
89 suites / 731 tests, `build` exit 0.

- [x] `npm run lint` — exit 0
- [x] `npm test` — exit 0, 89 suites / **744** tests (+13)
- [x] `npm run build` — exit 0
- [x] Non-empty metadata is reachable in the list, and every key/value renders
- [x] `metadata: {}` renders no metadata heading or disclosure
- [x] Metadata beyond the inline cap is fully reachable and the UI signals more exists
- [x] Number, `null` and nested-object values render without `[object Object]`
- [x] Editor shows metadata read-only with wording saying it isn't editable there — scoped to
      the edit panel, and verified to fail when the editor's renderer is removed
- [x] The edit form's PATCH body is exactly `{display_name}` with no `metadata` key
- [x] Import preview names the metadata columns with first-row samples, and follows a change of
      the selected name column
- [x] A CSV with only email and name columns reports that nothing becomes metadata
- [x] The posted `contacts` array is unchanged by this work
- [x] A replacement CSV that fails to map disarms the previous file's staged rows, so the
      preview can never describe a different file than the one Import posts
- [x] Two uploads whose `text()` promises resolve out of order leave the second file previewed
      and imported

After merge: worth eyeballing a real contact with a wide import (many columns, long values) on
both breakpoints — the caps are judgement calls, not contract.

### DIFF
--- a/web/src/app/(app)/contacts/_lib/csv.test.ts
+++ b/web/src/app/(app)/contacts/_lib/csv.test.ts
@@ -1,4 +1,4 @@
-import { mapCsvRows, parseCsv } from "./csv";
+import { getCsvMetadataColumns, mapCsvRows, parseCsv } from "./csv";
 
 describe("contacts CSV parsing", () => {
   it("handles BOM, quoted commas, escaped quotes, and newlines", () => {
@@ -18,6 +18,49 @@ describe("contacts CSV parsing", () => {
       address: "partner@fund.vc",
       display_name: "A. Partner",
       metadata: { Fund: "Example Capital" },
+    }]);
+  });
+
+  it("reports metadata columns with first-row samples", () => {
+    const table = parseCsv(
+      "email,name,company,role,notes\npartner@example.com,A. Partner,Example Capital,GP,met at demo day\n",
+    );
+
+    expect(getCsvMetadataColumns(table, "email", "name")).toEqual([
+      { name: "company", sampleValue: "Example Capital" },
+      { name: "role", sampleValue: "GP" },
+      { name: "notes", sampleValue: "met at demo day" },
+    ]);
+  });
+
+  it("updates metadata reporting when the selected name column changes", () => {
+    const table = parseCsv("email,name,company\npartner@example.com,A. Partner,Example Capital\n");
+
+    expect(getCsvMetadataColumns(table, "email", "name")).toEqual([
+      { name: "company", sampleValue: "Example Capital" },
+    ]);
+    expect(getCsvMetadataColumns(table, "email", "company")).toEqual([
+      { name: "name", sampleValue: "A. Partner" },
+    ]);
+    expect(mapCsvRows(table, "email", "company")).toEqual([{
+      address: "partner@example.com",
+      display_name: "Example Capital",
+      metadata: { name: "A. Partner" },
+    }]);
+  });
+
+  it("reports no metadata columns and skips empty headers", () => {
+    const onlyIdentity = parseCsv("email,name\npartner@example.com,A. Partner\n");
+    const withEmptyHeader = parseCsv("email,name, ,role\npartner@example.com,A. Partner,ignored,GP\n");
+
+    expect(getCsvMetadataColumns(onlyIdentity, "email", "name")).toEqual([]);
+    expect(getCsvMetadataColumns(withEmptyHeader, "email", "name")).toEqual([
+      { name: "role", sampleValue: "GP" },
+    ]);
+    expect(mapCsvRows(withEmptyHeader, "email", "name")).toEqual([{
+      address: "partner@example.com",
+      display_name: "A. Partner",
+      metadata: { role: "GP" },
     }]);
   });
 });
--- a/web/src/app/(app)/contacts/_lib/csv.test.ts
+++ b/web/src/app/(app)/contacts/_lib/csv.test.ts
@@ -63,4 +63,17 @@ describe("contacts CSV parsing", () => {
       metadata: { role: "GP" },
     }]);
   });
+
+  it("keeps mapping and reporting aligned when the selected email header is empty", () => {
+    const table = parseCsv(",alias,company\n1,A. Partner,Example Capital\n");
+
+    expect(getCsvMetadataColumns(table, "")).toEqual([
+      { name: "alias", sampleValue: "A. Partner" },
+      { name: "company", sampleValue: "Example Capital" },
+    ]);
+    expect(mapCsvRows(table, "")).toEqual([{
+      address: "1",
+      metadata: { alias: "A. Partner", company: "Example Capital" },
+    }]);
+  });
 });
--- a/web/src/app/(app)/contacts/_lib/csv.ts
+++ b/web/src/app/(app)/contacts/_lib/csv.ts
@@ -45,25 +45,50 @@ export type ImportPreviewRow = {
   metadata?: Record<string, string>;
 };
 
-export function mapCsvRows(
-  table: string[][],
-  emailColumn: string,
-  nameColumn?: string,
-): ImportPreviewRow[] {
+export type CsvMetadataColumn = {
+  name: string;
+  sampleValue: string;
+};
+
+function resolveColumns(table: string[][], emailColumn: string, nameColumn?: string) {
   if (table.length < 2) throw new Error("CSV must include a header and at least one row");
   const headers = table[0].map((header) => header.trim());
   const emailIndex = headers.indexOf(emailColumn);
   if (emailIndex < 0) throw new Error(`No “${emailColumn}” column found`);
   const nameIndex = nameColumn ? headers.indexOf(nameColumn) : -1;
   if (nameColumn && nameIndex < 0) throw new Error(`No “${nameColumn}” column found`);
+  const metadataIndexes = headers
+    .map((header, index) => ({ header, index }))
+    .filter(({ header, index }) => header && index !== emailIndex && index !== nameIndex);
+  return { emailIndex, nameIndex, metadataIndexes };
+}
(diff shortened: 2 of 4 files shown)


jiashuoz (approved):
Approving. Verified locally against the branch: `npm run lint` exit 0, `npm test` 89 suites / 744 tests, `npm run build` exit 0 — matching the PR body. All CI checks green.

The core design is right. `resolveColumns` as the shared seam between `mapCsvRows` and `getCsvMetadataColumns` is the correct fix for the drift risk, and I couldn't get the preview and the posted body to disagree: duplicate headers, blank headers, headers differing only by whitespace, and a `__proto__` header all resolve identically in both functions. The three pre-existing bugs are real — I confirmed a `" email "` header would previously have produced ``No " email " column found``, and that an `<option>` without an explicit `value` falls back to whitespace-collapsed text content per spec.

None of the following blocks the merge; they're worth a follow-up.

**1. The disclosure renders even when it has nothing to disclose** (`page.tsx:97`)

For metadata that already fits inline, the `<details>` body is byte-identical to the list above it. `{a: "1", b: "2"}` renders "View full metadata (2 fields)" and expanding it repeats the same two rows. That's the common case for a typical import, not an edge. Suggest gating on `entries.length > metadataInlineLimit || previewEntries.some(([, v]) => formatMetadataValue(v).length > metadataInlineValueLimit)`.

**2. Which 3 fields appear inline is arbitrary** (`page.tsx:84`)

`metadata` is `JSONB` (`migrations/079_contacts.sql:23`), so Postgres reorders keys by length-then-bytes and CSV column order is gone before the UI sees it. `Object.entries` then hoists integer-like keys to the front on top of that. `{company, role, notes, "2024"}` renders inline as `2024, company, role`, hiding `notes`. Net effect is close to "the three shortest key names, unless a numeric key jumps the queue." Sorting entries alphabetically before slicing would at least make the selection predictable.

**3. Two stated behaviors are untested, and one is unreachable** (`page.tsx:63-79`)

Nothing covers the 80-character inline truncation (it does work — 79 chars plus an ellipsis). Nothing covers the circular-structure placeholder either, and that branch can't be reached: metadata comes from `response.json()`, and `JSON.parse` cannot produce a cycle, so the `catch` returning "Unable to display value" is dead. Same class: `"undefined"` in the typeof list at line 66 is unreachable from JSON. Either drop the claims or drop the code.

**4. `contact.metadata` in a dependency array is object identity** (`page.tsx:499`)

This is safe today only because `editing` is a frozen state reference that `refresh()` never replaces. If a later change keeps the editor open across a refresh and re-derives `editing` from the new list, every refresh re-runs the GET and clobbers the user's in-progress `name` edit. `contact.address` is the real key.

**5. Duplicate headers colliding after trimming stay ambiguous** (`page.tsx:645`)

`email,tag, tag` renders two "tag" options with the same `value`; picking the second resolves to the first via `indexOf`, so one column is unselectable. The React key warning at this line is fixed, but the underlying ambiguity isn't. No divergence results — both sides last-win consistently — so this is cosmetic.

**6. Search doesn't cover what's now on screen** (`page.tsx:180`)

`shown` filters on address and display_name only. Now that metadata is visible in the list, searching by company name is the obvious next expectation.

**On density:** metadata renders inline on every card and every table row, where the issue suggested the row's expanded state. For a 5-column import across 100 contacts that's roughly four extra lines per row in a table already at `min-w-[720px]`. The PR flags this for an after-merge eyeball, which is the right call — but collapsing the whole block into a single disclosure would settle it and finding #1 at the same time.

PR #821, opened 2026-08-04
test(contract): cover plan-limit enforcement in all three runners

jiashuoz:
> **Stacked on #819** — based on `fix/domain-reclaim-limit` so the diff here is just this commit. GitHub will retarget it to `main` when #819 merges.

## Summary

No limit path in the product had live-server coverage in any language. `testutil.StartContractServer` set every cap to 100000 **deliberately** — its comment read *"contract scenarios exercise contract shape, not quota enforcement (the 402 limit paths have dedicated httpapi unit tests)"* — so the 402 envelope, the field an SDK reads to tell a caller **which** quota stopped them, was proven only by handler tests against fakes.

This adds that coverage to all three contract runners, and pins the re-register-at-cap exemption from #819 — the contract-level coverage that PR could not get on its own.

## Why a second account, not a "lower the caps" step

The obvious design is a setup primitive that lowers the caps for one scenario and restores them in cleanup. That is a trap here: caps are **account-global**, and both the TS and Python setup loops **silently ignore setup keys they do not recognize**. A cap lowered for one scenario and restored by a cleanup step some runner skipped would 402 every scenario after it — surfacing as a product bug, not a harness bug.

So the contract server seeds a **second account** (`CappedLimits`, one slot per create-limited resource) and hands its key to the runners as `E2A_TEST_CAPPED_API_KEY`. Scenarios reach it through `auth_override` — an existing primitive all three runners already support and already resolve variables through. Nothing mutates, so nothing can leak. The enforcer's 60s limits cache needs no special handling either: the row is written before the server serves its first request and never changes, so there is no staleness window to race.

CI needs **no workflow change** — the env file is already sourced with `set -a`.

## The scenario pins both directions, on two resources

`account_limits_enforced` asserts refusal **at** the cap (with `resource` / `limit` / `current`) *and* success once a slot is freed, across domains and agents. A server that 402'd unconditionally — or one that wired a single quota and left the others open — cannot pass it.

## Coverage cannot silently rot back to zero

This is the part worth reviewing closely, since a quota scenario that quietly stops running is indistinguishable from one that passes:

- **Go** owns the contract server in-process, so it can never legitimately lack the key → a missing capped key is a **hard failure**, never a skip. This is the per-PR gate that cannot be skipped.
- **TS / Python** skip only when pointed at a target with no capped account (a deployed staging server, which cannot provision one). Per-PR CI always supplies the key via the env file.
- **All three** carry an **always-on shape test** — no live server needed — that fails if the scenario is deleted, stops using the capped account, or loses either direction of enforcement.

## Test plan

- [x] **Go runner, live:** `✓ account_limits_enforced`
- [x] **TypeScript runner, live:** `✓ account_limits_enforced` (against a real `cmd/e2a-contract-server`)
- [x] **Python runner, live:** `PASSED [account_limits_enforced]` (same server)
- [x] Skip path with the key unset → `SKIPPED` cleanly, so the out-of-repo staging conformance job is unaffected
- [x] The exemption step **fails** without #819's fix (`402 limits: domains cap reached (1/1)`), proving it is a real pin and not incidental
- [x] Shape test teeth: defanging one assertion (`resource: domains` → `messages`) **fails** the build
- [x] `tests/contract`, `internal/testutil`, TS SDK suite (233 tests), Python contract file all green; `gofmt` / `go vet` clean

## Not covered

Only the two **create** caps (domains, agents). `max_messages_month` and `max_storage_bytes` are seeded on the capped account but unexercised — reaching them means sending mail, which the contract server accepts but deliberately does not deliver. Worth a follow-up rather than bolting onto this PR.

### DIFF
--- a/cmd/e2a-contract-server/main.go
+++ b/cmd/e2a-contract-server/main.go
@@ -29,7 +29,14 @@ func main() {
 		log.Fatalf("contract server health check failed: %v", err)
 	}
 
-	envContent := fmt.Sprintf("E2A_TEST_BASE_URL=%s\nE2A_TEST_API_KEY=%s\n", srv.BaseURL, srv.APIKey)
+	// E2A_TEST_CAPPED_API_KEY authenticates the secondary account seeded with
+	// testutil.CappedLimits. Scenarios that assert quota enforcement run as
+	// that account; CI sources this file with `set -a`, so the runners pick it
+	// up with no workflow change.
+	envContent := fmt.Sprintf(
+		"E2A_TEST_BASE_URL=%s\nE2A_TEST_API_KEY=%s\nE2A_TEST_CAPPED_API_KEY=%s\n",
+		srv.BaseURL, srv.APIKey, srv.CappedAPIKey,
+	)
 	if envFile != "" {
 		if err := os.WriteFile(envFile, []byte(envContent), 0o600); err != nil {
 			log.Fatalf("write env file: %v", err)
--- a/internal/testutil/contract_server.go
+++ b/internal/testutil/contract_server.go
@@ -25,10 +25,38 @@ import (
 	"github.com/tokencanopy/e2a/internal/ws"
 )
 
+// CappedLimits are the plan caps of the secondary account described on
+// ContractServer.CappedAPIKey. One slot per create-limited resource is the
+// smallest cap that can express BOTH directions of enforcement: a scenario can
+// consume the slot (success), be refused at the cap (402), free it, and
+// succeed again — proving the limit is a cap and not an unconditional refusal.
+var CappedLimits = limits.Limits{
+	PlanCode:         "contract_capped",
+	MaxAgents:        1,
+	MaxDomains:       1,
+	MaxMessagesMonth: 1,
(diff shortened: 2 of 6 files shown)


jiashuoz:
Folded into #819 rather than reviewed separately — GitHub marked this merged automatically when its head commit became reachable from its base branch (`fix/domain-reclaim-limit`), which is accurate: all of this work is in #819 now.

The reason for folding: a compliance review flagged that splitting these was itself the problem. AGENTS.md requires a changed `/v1` behavior to be pinned by a scenario proven in all three runners, and keeping the fix in #819 with its proof here meant **#819's own CI never ran the scenario that proves it** — and merging in order would have left a window on `main` with the behavior changed and no coverage.

#819 now carries all three commits, the third addressing review findings against this branch: the scenario leaked its first agent on any mid-run failure (permanently parking the capped account at its cap), the `{capped_api_key}` guard was satisfied by the scenario's own description prose, and two mutations were caught by nothing at all — emptying `plan_code`/`upgrade_url`, and every cap being 1 which made `limit` and `current` mutually substitutable.

Review happens on #819.

PR #847, opened 2026-08-07
fix(mcp): let delete_agent purge permanently, matching REST

### LINKED ISSUE #633: MCP cannot permanently delete an agent, so domain re-registration is impossible over MCP alone
## Summary

The MCP server exposes `delete_agent` (soft delete → 30-day trash) and `restore_agent`, but **no way to permanently delete / purge a trashed agent**. Because a trashed agent continues to hold its address, an MCP-only user cannot free an address — and therefore cannot re-register a domain — without dropping to REST.

## Reproduction

All over MCP, on a domain with one agent:

```
delete_agent(agent@inbox.mnexa.ai, confirm: true)
  → { deleted: true }            # soft delete, agent → trash

delete_domain(inbox.mnexa.ai, confirm: true)
  → domain_has_agents: "cannot delete domain while agents exist —
     delete its agents first (including any in the trash: they hold
     the address until restored or permanently deleted)"

create_agent(agent@inbox.mnexa.ai)
  → address_in_trash: "this address belongs to an inbox in your trash —
     restore it, or delete it permanently from the trash, to reuse the address"

get_agent(agent@inbox.mnexa.ai)
  → not_found                    # trash is hidden from reads
```

Both error messages tell the user to delete permanently. **MCP offers no tool that does that.**

## Cause

`mcp/src/tools/agents.ts` — `delete_agent`'s input schema is exactly two fields, and it's a `strictInputSchema`, so a `permanent` key is rejected outright:

```ts
inputSchema: strictInputSchema({
  email:   z.string().email().optional(),
  confirm: z.literal(true),
}),
...
return client.deleteAgent(args.email);
```

The capability exists server-side — `DELETE /v1/agents/{email}?confirm=DELETE&permanent=true` — and is used internally by the prober, which permanently deletes its ephemeral agents specifically so repeated runs don't accumulate trash. It simply isn't surfaced over MCP.

## Impact

The only MCP escape from a soft-deleted agent is `restore_agent`, which undoes the user's intent. Otherwise they wait out the 30-day auto-purge.

This makes **domain re-registration impossible over MCP**, which matters because re-registration is currently the only remedy for domains registered before per-customer SES sender identities landed (server v1.0.4) — those have no DKIM key material and permanently fail sending-identity provisioning with `"no DKIM key material for domain; re-register the domain"`. So the documented fix for that condition is unreachable from MCP.

Hit this on all three legacy domains during a real re-registration; every one required a REST call to proceed.

## Suggested fix

Add an optional `permanent: true` to `delete_agent`, gated exactly as REST gates it (explicit confirmation + account scope). MCP already exposes plenty of destructive operations behind `confirm: true`, so this is consistent with the existing safety model.

Worth noting the SDK operation-coverage gate added in #625 would not catch this — it checks that every OpenAPI operation is reachable from the **SDKs**, not from MCP. An equivalent MCP-coverage check might be worth considering, since this gap is exactly the class it would find.

AmirF194:
REST's `DELETE /v1/agents/{email}` accepts `permanent=true` to skip the 30-day
trash and delete an agent right away (`AgentsApi.ts:143`, `client.ts:323`), but
the MCP `delete_agent` tool always called `client.deleteAgent(email)` with no
way to set it. `delete_domain`'s own description says a domain can only be
re-registered once every agent on it, live or trashed, is gone for good
(`domains.go:601,609`), so an MCP-only caller had no path to that state: the
only remedy was the REST/SDK layer directly.

Added an optional `permanent` boolean to `delete_agent`'s schema and threaded
it through `McpClient.deleteAgent` to the SDK's existing `agents.delete(email,
{ permanent })`. `delete_agent` is already account-scope only
(`mcp/src/tools/tiers.ts` `ADMIN_TOOLS`), the same boundary the REST handler
enforces server-side for `permanent=true` (`trash_endpoints_test.go` asserts
403 for an agent-scoped key), so this does not hand the capability to a
runtime agent credential. `delete_message` deliberately withholds `permanent`
for that reason (`client.ts:428`); `delete_agent` sits on the other side of
that boundary already, which is why the same argument does not apply here.

**Verification**
- `delete_agent forwards permanent:true to skip the trash` and the two
  existing forwarding tests (updated for the new second argument) fail on the
  pre-fix source and pass after it, run both ways in a clean container.
- Full `mcp` workspace suite (303 tests) and `test:types` pass; `tsc --build`
  and `check-mcp-tool-compat` (additive-only) both clean.
- Did not check: no live-server contract run against this endpoint, since the
  REST/SDK side of `permanent` already ships and is covered by
  `trash_endpoints_test.go`; this PR only extends what MCP exposes.

Fixes #633

### DIFF
--- a/mcp/README.md
+++ b/mcp/README.md
@@ -165,7 +165,7 @@ before it is declared stable**.
 | `reply_to_message` | Reply to a message — one the agent received (replies to its sender) or one it sent (continues the thread to the original recipients). Preserves In-Reply-To / References for thread continuity and accepts the same optional `send_at` as `send_message`. |
 | `forward_message` | Forward a message into a new thread, carrying its existing attachments by default. Accepts the same optional `send_at` as `send_message`. |
 | `list_messages` | List mail; pass `deleted:true` to list trash. Filter by `read_status` (unread / read / all) and sender with reserved-word-safe `from_`; cursor-paginated (`cursor` + `limit` in, `next_cursor` out). |
-| `delete_message` | Move a message to the trash (restorable for ~30 days). Requires `confirm: true`. Permanent deletion is deliberately not exposed over MCP — use the REST API/SDK. |
+| `delete_message` | Move a message to the trash (restorable for ~30 days). Requires `confirm: true`. Per-message permanent deletion is deliberately not exposed over MCP — use the REST API/SDK. (Purging a whole inbox is available to account scope via `delete_agent`'s `permanent: true`.) |
 | `restore_message` | Restore a soft-deleted message and resume its retention clock. |
 | `get_message` | Fetch full body, headers, and attachment metadata for one message. |
 | `get_attachment` | Get one attachment's metadata + a short-lived `download_url` (fetch the bytes out of band); `inline: true` returns base64 `data` for small files (≤256 KB). |
--- a/mcp/src/client.ts
+++ b/mcp/src/client.ts
@@ -164,9 +164,9 @@ export class McpClient {
     return this.sdk.agents.replaceProtection(this.resolveAddress(explicitAddress), config);
   }
 
-  async deleteAgent(explicitAddress?: string): Promise<DeleteAgentResult> {
+  async deleteAgent(explicitAddress?: string, permanent?: boolean): Promise<DeleteAgentResult> {
     const address = this.resolveAddress(explicitAddress);
-    return this.sdk.agents.delete(address);
+    return this.sdk.agents.delete(address, { permanent });
   }
 
   restoreAgent(explicitAddress?: string): Promise<AgentView> {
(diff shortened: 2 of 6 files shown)


jiashuoz (approved):
Verified the plumbing end to end: `permanent: undefined` produces a byte-identical pre-PR request (the generated client guards on `!== undefined`), tool catalog is unchanged at 76/76 so the compat gate is unaffected, and the new description matches the server behavior in `agents_write.go`.

Pushed a follow-up commit adding client-layer coverage — the flag was only asserted against a stubbed `McpClient`, so reverting `client.ts` to a one-argument `sdk.agents.delete(address)` kept the whole suite green while silently downgrading every purge to a trash move. Also corrected two comments that this change falsified (`tiers.ts` was using "the MCP tool doesn't expose it at all" to justify `delete_message`'s runtime tier) and the `delete_domain` dependency in the new tool description.

Thanks for the clean patch.

jiashuoz:
@AmirF194 Thanks for your contribution!

AmirF194:
Appreciate the follow-up commit, good catch on the stubbed McpClient hiding the client-layer regression. Thanks for merging.

PR #851, opened 2026-08-08
fix(web): render blog posts in the reader's theme

wchwawa:
## Summary

Blog posts were readable only in light mode. The page frame around a post —
the nav, the footer, and the post index — followed the reader's theme, but the
article itself did not: it kept light-theme colors no matter what. A visitor
whose browser or operating system defaults to dark got a dark page with
near-black text on it. Measured on the published site, headings sat at roughly
1.1:1 against the page background and body copy at 1.5:1, which is effectively
invisible; the pale background behind inline code fragments made them the only
legible part of the article. All four posts were affected, not just the one this
was reported on.

The article now takes its colors from the same shared theme values the rest of
the site already uses, so headings, body copy, lists, links, quotes, rules and
code all follow whichever theme is active. Headings and body copy now measure
about 15:1 against the background in dark mode and 17:1 in light, links 10:1 and
6:1, quotes 7:1 and 5:1. Light mode is otherwise unchanged, apart from body copy
sitting a little darker than it did before.

One smaller fix rides along: the date line above each post title was asking for
a font the site never loads and silently falling back to whatever the reader's
system offered. It now uses the site's own monospace face.

This PR is the fix only. It changes no behavior and adds nothing new — see the
follow-ups below for what it deliberately leaves out.

## Operational risk

Low. Presentation only — no user data, no side effects, no auth, billing,
notifications or external integrations touched, and nothing behind a flag to
roll back. Rollback is a plain revert.

Two things a reviewer should weigh rather than assume:

- Body copy in light mode is now slightly darker than the shade it replaced.
  There is no shared value for the in-between tone the posts used, so this takes
  the standard body color rather than introducing a new one. If the softer
  reading is preferred, that is a deliberate addition and belongs in its own
  change.
- The date line above each title now follows the same treatment as the date on
  the post index, which keeps the two pages consistent. It is still a low
  contrast line — about 2.9:1 in light and 3.3:1 in dark — so it does not meet
  AA. That is not new; it reads slightly better in light mode than before and
  slightly worse in dark. Raising it means changing the index at the same time,
  which is why it is not here.

## Test plan

- [x] Reproduced the original problem on the published post with the browser
      defaulting to dark, and recorded the contrast of every text element in the
      article before the change.
- [x] Re-measured the same elements locally after the change, in both themes.
      Everything in the article body clears AA, and headings, body copy and code
      clear AAA. The date line above the title is the one exception, noted above.
- [x] Confirmed the light theme still matches its previous appearance, value by
      value, with body copy the one intended difference.
- [x] Checked all four posts, since they share the same presentation, and the
      post index, which was already correct and is unchanged.
- [x] Dashboard test suite, lint and type check all clean. The four type errors
      that remain are in unrelated test files and predate this branch.

After merge: open any post with the operating system set to dark and confirm the
article reads normally.

## Follow-ups, not in this PR

- **A theme control on the blog.** Right now the browser or operating system
  preference is the only thing deciding how a post reads, and the site's only
  theme control lives in the dashboard sidebar, which blog readers never see —
  they arrive from search with no session. So a rendering problem in one theme
  has no workaround for exactly the people most likely to hit it, and this PR
  does not change that. Recommended as the next change: put the same three-way
  control the dashboard already uses into the blog nav, so a reader can follow
  the system preference or pin either theme. Prepared as its own PR, stacked on
  this one, on `feat/blog-theme-switcher`.
- Nothing stops the next contributor from reintroducing fixed colors in a post
  and quietly breaking one theme again. A small guard in the test suite would
  catch it; the contributor checklist does not currently mention this surface at
  all.
- Dark-mode readers still see a brief flash of the light theme on first paint,
  because the theme is applied after the page loads rather than before it paints.
  That predates this change and affects the whole site, not only the blog.
- The docs pages were not audited for the same class of problem.

### DIFF
--- a/web/src/app/blog/email-agent-with-google-adk/page.mdx
+++ b/web/src/app/blog/email-agent-with-google-adk/page.mdx
@@ -23,7 +23,7 @@ export const metadata = {
 
 <PostSchema post={post} />
 
-<div style={{ fontSize: 11, color: "#A89A8A", fontFamily: "'IBM Plex Mono', monospace", marginBottom: 6 }}>
+<div style={{ fontSize: 11, color: "var(--fg-subtle)", fontFamily: "var(--f-mono)", marginBottom: 6 }}>
   {new Date(post.date + "T00:00:00Z").toLocaleDateString("en-US", { month: "long", day: "numeric", year: "numeric", timeZone: "UTC" })} · {post.readingMinutes} min read
 </div>
 
--- a/web/src/app/blog/email-for-openclaw-agents/page.mdx
+++ b/web/src/app/blog/email-for-openclaw-agents/page.mdx
@@ -23,7 +23,7 @@ export const metadata = {
 
 <PostSchema post={post} />
 
-<div style={{ fontSize: 11, color: "#A89A8A", fontFamily: "'IBM Plex Mono', monospace", marginBottom: 6 }}>
+<div style={{ fontSize: 11, color: "var(--fg-subtle)", fontFamily: "var(--f-mono)", marginBottom: 6 }}>
   {new Date(post.date + "T00:00:00Z").toLocaleDateString("en-US", { month: "long", day: "numeric", year: "numeric", timeZone: "UTC" })} · {post.readingMinutes} min read
 </div>
 
--- a/web/src/app/blog/human-in-the-loop-for-agent-email/page.mdx
+++ b/web/src/app/blog/human-in-the-loop-for-agent-email/page.mdx
@@ -23,7 +23,7 @@ export const metadata = {
 
 <PostSchema post={post} />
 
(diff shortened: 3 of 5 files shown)


jiashuoz (approved):
Verified locally on the branch: all 13 Loft tokens used exist in both theme blocks of globals.css, ThemeProvider's .dark toggle reaches the blog via the root layout, and no hex literals remain anywhere under web/src/app/blog/. Reproduced the contrast claims independently (dark body 15.2:1, light 16.7:1, links 10.3/6.0, old h1-on-dark 1.09:1) — they match the PR description to the digit. No rehype highlighting plugin, so block code safely inherits --ink-fg. Build (static export), lint, and 773/773 Jest tests all green. The date-line font switch to var(--f-mono) is a real fix too: IBM Plex Mono was never loaded, so it was silently rendering as system monospace. Non-blocking: the sub-AA date line and the no-hex regression guard are correctly deferred to follow-ups.

PR #860, opened 2026-08-08
feat(notify): give operational email a real reply path, drop the noreply senders

jiashuoz:
## The bug

HITL approval mail has shipped with a **Reply-To that goes nowhere**. Confirmed in a live inbox:

```
from:      e2a <hitl-noreply@send.e2a.dev>
reply-to:  hitl-noreply@send.e2a.dev
```

Reply-To was hardcoded to the sender address, on the platform relay domain — whose MX points at SES's feedback endpoint, so it has no mailbox. The original intent was right (*"point replies back at the platform, not the agent"*) but the destination can't receive.

A reviewer who hits Reply gets silence. That's the worst place for it: an approval is **time-boxed**, and a reviewer who can't sign in has only the magic links.

## Changes

- **`hitlnotify` reads `notifications.from_address` / `notifications.reply_to`** — the same keys `webhooknotify` already uses. Resolution logic mirrors `webhooknotify.New` so the two can't drift.
- **Reply-To** is the configured inbox when set; unset falls back to today's exact behaviour, so unconfigured self-hosts are unchanged.
- **Sender local parts drop "noreply"**: `hitl-noreply` → `approvals`, `webhooks-noreply` → `webhooks`.

Once `reply_to` points at a real inbox, a From that reads `noreply` tells the reader the opposite — in the most visible field in their client. (`hitl` was also internal jargon that had leaked onto a customer-facing address.)

## Why distinct rather than consolidated

The two defaults stay **distinct**. An approval is time-boxed; routine webhook-health mail isn't. Keeping them separately filterable is worth more than one shared identity — and the existing comment on the `hitlnotify` constant says the fixed local part exists precisely so clients can group on it.

Setting `notifications.from_address` still collapses both into one identity, so consolidation is one line of config away if that turns out to be preferable.

## Tests

- a configured `reply_to` reaches the wire, and does **not** leave the sender in Reply-To
- From is unchanged by it — only Reply-To moves
- no `noreply` survives anywhere in a replyable message
- a configured `from_address` overrides the default, and the Message-ID domain follows it rather than claiming a relay domain the From no longer uses

Both packages green; repo-wide `go test -short` at 0 failures.

## ⚠️ Release notes

**The From address on HITL approval emails changes from `hitl-noreply@` to `approvals@`.** Any customer filter matching the old address stops matching. This is unavoidable once the mail becomes replyable, but it shouldn't arrive silently.

## Follow-on (ops repo, not here)

This adds the seam; the hosted values still need setting:

```yaml
notifications:
  reply_to: support@agents.e2a.dev
```

`config.prod.yaml` is a risky-config path, so that lands via a staging cycle. Until it does, both senders keep their current (now non-noreply) defaults and no Reply-To header.

### DIFF
--- a/cmd/e2a/main.go
+++ b/cmd/e2a/main.go
@@ -600,7 +600,7 @@ func main() {
 		// unreachable in practice — kept as a defensive guard against future drift.
 		log.Printf("[hitl] notifier disabled: notification job pipeline not registered")
 	} else {
-		notifier := hitlnotify.New(store, smtpRelay, approvalSigner, cfg.OutboundSMTP.FromDomain, cfg.Notifications.FromAddress, cfg.Notifications.ReplyTo, cfg.HTTP.PublicURL)
+		notifier := hitlnotify.New(store, smtpRelay, approvalSigner, cfg.OutboundSMTP.FromDomain, cfg.Notifications.FromAddress, cfg.Notifications.ReplyTo, cfg.HTTP.PublicURL).WithDKIM(store)
 		// Late-bind the concrete Deliverer onto the registered NotifyWorker (which
 		// has been running since jobsClient.Start; jobs enqueued before this bind
 		// simply retry) and give the hold path its accept-tx enqueuer. The HTTP
--- a/cmd/e2a/main.go
+++ b/cmd/e2a/main.go
@@ -600,7 +600,7 @@ func main() {
 		// unreachable in practice — kept as a defensive guard against future drift.
 		log.Printf("[hitl] notifier disabled: notification job pipeline not registered")
 	} else {
-		notifier := hitlnotify.New(store, smtpRelay, approvalSigner, cfg.OutboundSMTP.FromDomain, cfg.HTTP.PublicURL)
+		notifier := hitlnotify.New(store, smtpRelay, approvalSigner, cfg.OutboundSMTP.FromDomain, cfg.Notifications.FromAddress, cfg.Notifications.ReplyTo, cfg.HTTP.PublicURL)
 		// Late-bind the concrete Deliverer onto the registered NotifyWorker (which
 		// has been running since jobsClient.Start; jobs enqueued before this bind
 		// simply retry) and give the hold path its accept-tx enqueuer. The HTTP
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -116,25 +116,28 @@ outbound_smtp:
   # explicitly only for non-standard endpoints.
   # message_id_domain: "us-east-2.amazonses.com"
 
-# Platform notification emails (webhook health alerts: an early warning
-# when a webhook's deliveries start failing, and a notice when e2a
-# auto-disables one).
+# Platform notification emails: HITL approval requests, plus webhook health
+# alerts (an early warning when a webhook's deliveries start failing, and a
+# notice when e2a auto-disables one). Both senders read the settings below.
 notifications:
-  # Sender address for those emails. Optional — when empty they send from
-  # a fixed no-reply local part on outbound_smtp.from_domain, so
-  # self-hosted deployments need no configuration here. If set, it must
-  # be an address your upstream is authorized to send as. If e2a holds a
-  # DKIM key for the address's domain (a domain registered with this
-  # server), each notification is DKIM-signed in-process for that domain;
-  # with no stored key it is sent unsigned.
+  # Sender address for those emails. Optional — when empty, each kind sends
+  # from its own fixed local part on outbound_smtp.from_domain (approvals@
+  # and webhooks@, kept distinct so a time-boxed approval stays filterable
+  # apart from routine webhook mail), so self-hosted deployments need no
+  # configuration here. SETTING THIS COLLAPSES BOTH INTO ONE IDENTITY.
+  # If set, it must be an address your upstream is authorized to send as.
+  # If e2a holds a DKIM key for the address's domain (a domain registered
+  # with this server), each notification is DKIM-signed in-process for that
+  # domain; with no stored key it is sent unsigned.
   # from_address: "support@send.your-company.example"
   #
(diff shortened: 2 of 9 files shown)


jiashuoz (commented):
Reviewed the full diff, both notifiers at the head, and the config plumbing. Core fix is correct, well-tested, and backward compatible. One substantive gap and some doc drift worth addressing.

### 1. DKIM parity gap (substantive)

`hitlnotify` now honors `notifications.from_address`, but it's the only notification sender that never DKIM-signs in-process. Compare the two wirings in `cmd/e2a/main.go`:

- `webhooknotify.New(...).WithDKIM(store)` — signs for a custom from-domain when e2a holds the key
- `hitlnotify.New(...)` — no signing leg

The `webhooknotify` comments spell out why this matters: the relay never signs, and an upstream like SES only signs identities *it* manages, so a BYODKIM custom `from_address` domain is signed in-process or not at all. With this PR, an operator who sets a custom `from_address` gets signed webhook-health mail but **unsigned HITL approval mail** — the most time-sensitive email the platform sends, and unsigned-on-custom-domain is exactly the DMARC-fails-after-forward scenario the webhooknotify comment calls the worst failure mode. Either wire `WithDKIM` onto the HITL notifier too, or document why approval mail is exempt (e.g. hosted always sends it from the SES-managed relay domain).

### 2. Doc comments now describe the old world

- `internal/config/config.go` — `NotificationsConfig` says it covers "currently the webhook health warning/disabled alerts." It now also drives HITL approval mail.
- `config.example.yaml` — same scoping, plus two newly-false claims: "a fixed **no-reply** local part" (both local parts dropped noreply) and "each notification is DKIM-signed in-process for that domain" (not true for HITL mail — ties to finding 1).
- The `reply_to` doc in both places says unset means "no Reply-To header is emitted and replies follow From." True for webhooknotify; hitlnotify emits `Reply-To: <sender>` when unset. Same destination, but the header *is* emitted — worth one clause.

### Minor (no action needed)

- "Resolution mirrors webhooknotify.New so the two can't drift" — it's a verbatim copy, not a shared helper, so it can drift. Fine at this size; the comment overstates the guarantee. A tiny shared resolver in `internal/outbound` would make the claim true if you care.
- The unset-`reply_to` fallback asymmetry (hitlnotify emits `Reply-To: <sender>`, webhooknotify omits the header) is deliberate and functionally identical — replies land on the same address either way.
- Ops follow-up, not code: the production bug is only actually fixed once hosted config sets `E2A_NOTIFICATIONS_REPLY_TO`, and the new `approvals@` / `webhooks@` local parts need SES sending authorization (domain-level verification should cover it — worth confirming before deploy, along with any bounce/allowlist rules keyed on the old addresses).

### What's good

- Default-off behavior is byte-identical to today for unconfigured self-hosts.
- New tests are exactly the right ones: configured `reply_to` on the wire, no "noreply" anywhere in headers, `from_address` override, Message-ID domain tracking the configured address.
- Config validation rejects display-name forms, so header injection via config is closed.

🤖 Review by [Kimi Code](https://www.kimi.com/code)

jiashuoz:
Thanks — the DKIM finding was correct and I've fixed it. Verified each claim against the code rather than taking it as read; here's what I found and did.

## 1. DKIM parity gap — confirmed and fixed (`a1b748cb`)

Accurate. `hitlnotify` had **no signing leg at all** — no `dkim` field, no `WithDKIM`, no `SignWithDKIM` call — while `webhooknotify` is wired `.WithDKIM(store)` at `main.go:624`. Making `from_address` configurable without it created exactly the asymmetry you describe.

One clarification on blast radius: there's **no live impact today**, because hosted deliberately leaves `from_address` unset (ops PR #300), so approval mail sends from `approvals@send.e2a.dev` — an SES-managed identity SES signs itself. Confirmed against a real inbox: `signed-by: send.e2a.dev`. But the PR introduced the trap for anyone who sets the key, so it's right to close it rather than document an exemption.

Fixed by mirroring `webhooknotify` exactly: `WithDKIM` on the notifier, wired in `main.go`, and the signing call placed **after** the Message-ID prepend so the signature covers the final header set. Fail-open preserved — no lookup, no stored key, or a signing failure all send unsigned.

Tests: signed for a configured `from_address` domain (asserting `d=` matches the From domain), and unsigned both with no stored key and with no lookup wired. **Verified the signed case fails with the signing leg removed** — it isn't a vacuous assertion.

## 2. Doc drift — all three corrected

- `config.go` `NotificationsConfig` — now names both senders.
- `config.example.yaml` — dropped the "no-reply local part" claim, and the in-process-DKIM claim is now true for both senders rather than only one.
- The `reply_to` semantics are now stated per sender: `webhooknotify` omits the header when unset, `hitlnotify` emits `Reply-To: <sender>`. Same destination, but you're right that the header *is* emitted, and saying otherwise was wrong.

Also documented something the config reference never stated: **setting `from_address` collapses the two senders into one identity.** That's the consolidation lever, and it was invisible.

## 3. "Cannot drift" — fair, softened

It's a verbatim copy, not a shared helper, so it demonstrably can. Now reads "changing one means changing the other." I've left it as a copy rather than extracting a resolver into `internal/outbound` — two call sites of six lines doesn't yet earn a shared abstraction, and the comment no longer claims a guarantee the code doesn't provide.

## 4. Ops follow-up — already open

`notifications.reply_to` is set in **tokencanopy/e2a-ops#300** for both prod and staging. Two things from your note worth recording there:

- **SES authorization for the new local parts** — domain-level verification should cover `approvals@` / `webhooks@` on `send.e2a.dev`, but it's unconfirmed and I've flagged it as a pre-deploy check.
- **Bounce/allowlist rules keyed on the old addresses** — good catch, hadn't considered it. Added to the release-notes item alongside the customer-filter breakage.

Full suites green: both notifier packages, `internal/config`, and repo-wide `go test -short` at 0 failures.

PR #869, opened 2026-08-09
perf(metrics): raise work_mem for the account-metrics aggregates

jiashuoz:
## What

Raise `work_mem` to 32MB for the account-metrics read transaction only. One `SET LOCAL`, no other change.

## Why

Both aggregates in `accountMetrics` use `count(DISTINCT message_id)`, which forces a `GroupAggregate` over a sort of every lifecycle transition in the window rather than a hash aggregate. At the default 4MB that sort spills hard, and the cost lands on `/v1/metrics` — a dashboard-facing endpoint.

Measured on a locally seeded account (50 agents / 300k messages / 900k transitions), warm cache, medians of alternating runs:

| work_mem | totals | per-agent |
|---|---|---|
| 4MB (today) | ~780ms | ~1510ms |
| **32MB (this PR)** | **~480ms** | **~1230ms** |
| 64MB | ~450ms | ~1070ms |

32MB takes most of the win; past it the curve flattens while the per-node cost keeps scaling with concurrency. The sort still spills at 32MB (~60MB totals / ~70MB per-agent) — sizing `work_mem` to hold it outright would cost far more memory than the remaining ~10% is worth.

## Alternatives measured and rejected

Recorded in the code comment so they aren't re-derived later:

- **Two-phase group-then-sum rewrite** (replacing `count(DISTINCT)` with an inner per-message grouping): **~2.5x slower** on the totals query (~1350ms vs ~520ms at matched `work_mem`) and within noise on the per-agent one. Reverted.
- **Covering index** on `(reason_code, stage, outcome, retryable, message_id) INCLUDE (reconstructed)`: **not adopted by the planner at all** — identical plan, identical spill, identical timing.

## Safety

`SET LOCAL` is scoped to the transaction and reverts on commit/rollback, so it cannot leak onto a pooled connection. Verified directly, including that it is permitted inside the existing read-only repeatable-read transaction:

```
inside tx      | 32MB
after rollback | 4MB
```

## Scope

No migration, no schema change, no OpenAPI change, no SDK or client change. Existing `messagelifecycle` integration tests pass unchanged (`go test -tags=integration ./internal/messagelifecycle/...`).

## Note on the release

This does **not** on its own resolve the `account.metrics()` client-surface test timeout that blocked v1.7.2. The queries were already comfortably under 5s warm at this scale; the timeout looks environment-driven (cold cache / small staging instance), which is why the 30s budget on the grouped test still earns its place.

### DIFF
--- a/internal/messagelifecycle/account_metrics.go
+++ b/internal/messagelifecycle/account_metrics.go
@@ -65,32 +65,7 @@ func (s *Store) CountByReasonCodeForAccount(ctx context.Context, userID string, 
 	}
 	defer func() { _ = tx.Rollback(ctx) }()
 
-	// Both aggregates below use count(DISTINCT message_id), which forces a
-	// GroupAggregate over a sort of every transition in the window rather than
-	// a hash aggregate. At the default 4MB that sort spills hard. Raising it
-	// for this transaction only is the whole optimisation — measured on a
-	// seeded account (50 agents / 300k messages / 900k transitions, warm
-	// cache, medians):
-	//
-	//	work_mem   totals    per-agent
-	//	4MB        ~780ms    ~1510ms
-	//	32MB       ~480ms    ~1230ms
-	//	64MB       ~450ms    ~1070ms
-	//
-	// 32MB takes most of the win; past it the curve flattens while the
-	// per-node cost keeps scaling with concurrency. Note the sort still spills
-	// even here (~60MB totals / ~70MB per-agent) — sizing work_mem to hold it
-	// outright would cost more memory than the remaining ~10% is worth.
-	//
-	// Two rewrites were measured and REJECTED, so don't re-derive them:
-	// reshaping the aggregate into a two-phase group-then-sum was ~2.5x SLOWER
-	// on the totals query and noise on the per-agent one, and a covering index
-	// on (reason_code, stage, outcome, retryable, message_id) was not adopted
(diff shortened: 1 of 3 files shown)


jiashuoz:
Closing: this was built on a wrong root cause.

The staging metrics timeout turned out to be ~6,949 soft-deleted agents on the e2e account flipping the planner off `idx_messages_agent_created` and onto sequential scans of the whole `messages`/`message_lifecycle_transitions` tables — 1015ms with the tombstones vs 22.5ms scoped to live agents. That is fixed at the source in #870 (the e2e harness now purges fixture agents with `permanent=true` instead of trashing them) and in the query in #871 (account rollups now filter `a.deleted_at IS NULL`). `work_mem` was never the lever.

The measurements here are still accurate for what they measured — a single account with 300k messages / 900k transitions and every agent live, where 4MB→32MB buys ~780→480ms on the totals query and ~1510→1230ms per-agent. But that shape does not exist in prod today, the sort still spills at 32MB so it is only a partial mitigation, and it adds a non-obvious per-transaction memory knob to a hot read path. #871 also rewrote the doc comment on the same function, so this would need a rebase.

If an account ever gets large enough for this to matter, better to re-derive against that account's real data than to ship a value tuned on synthetic rows. Numbers and the two rejected alternatives (two-phase aggregate rewrite: ~2.5x slower; covering index: not adopted by the planner) stay recorded above.

PR #872, opened 2026-08-09
fix(availability): probe readiness off the request path, cap concurrent deletes

jiashuoz:
Closes the failure mode found while purging test fixtures on staging: six concurrent permanent agent deletes saturated the connection pool, and the instance was then **evicted from the load balancer** even though the database was merely busy.

Evidence from the staging incident:

```
03:34:04 leadership.Elector: reelection timed out after 5s
03:34:16 producer: JobGetAvailable timed out after 10s  (every queue)
         Server be_http/e2a-green is DOWN, reason: Layer7 timeout, check duration: 2000ms.
         0 active and 0 backup servers left.
         http_in be_http/<NOSRV> ... 503
```

Note `<NOSRV>`: requests never reached the app. That is why even `/api/health` — which touches nothing — returned 503.

## 1. Readiness is evaluated off the request path

`/readyz` called `pool.Ping` with a 2s deadline. Under pool exhaustion the check queued behind real work and timed out, reporting `database unreachable` for a database that was still there. So *saturation was indistinguishable from a dead database*, and HAProxy's L7 check removed the only backend.

Now a background ticker keeps the verdict fresh and the handler serves it from an atomic — it never competes for a pooled connection. A failing probe only flips readiness after a **15s grace window**, so transient pressure degrades latency instead of removing the instance.

The probe itself runs on a **connection the monitor owns, outside the pool** (added after review). Probing via `pool.Ping` would have reintroduced the bug in slower motion: it acquires a pooled connection, so a saturation outlasting the grace window would starve the probe and evict the instance anyway. Pool usability is checked separately with a 250ms acquire, where a timeout means *busy* (tolerated) and any other error means the pool is closed or broken (not ready) — otherwise a dead pool would read as healthy.

**The startup guarantee is preserved**, which matters for the deploy pipeline:
- The first probe runs **synchronously**, and the instance reports not-ready until it succeeds — a deploy whose migrations didn't apply still never joins the rotation.
- `draining` still short-circuits ahead of the dependency verdict, so blue/green drain is unchanged.

## 2. Per-account concurrency cap on destructive operations

`deleteAgent` and `deleteDomain` were in **no limiter at all** — not the poll limiter (7 read ops), not the per-IP `createAgent` limiter. Now capped at **2 in-flight per account**.

A concurrency cap rather than a rate limiter, because the hazard isn't request frequency — it's several unbounded cascades running at once, each holding a pooled connection. A rate limiter loose enough to permit legitimate bulk cleanup wouldn't have stopped this. Sequential deletion is unaffected, so the e2e harness and any customer draining an account still run at full speed.

**Deliberately not exempt for `ClassSystem`/`ClassInternal`**, unlike every other limiter here. Those exemptions exist because the rate limiters bound *user abuse*, and first-party traffic isn't abuse. This cap protects the instance from itself — and the incident that motivated it was caused by an internal-class client. Exempting the classes most likely to run bulk operations would leave the actual failure mode open.

## Verification

- `go test ./internal/httpapi/ ./cmd/e2a/` green; cap tests also pass under `-race`.
- The tolerance test is **not vacuous**: reverting the grace logic to the old always-flip behavior makes it fail with exactly the original symptom — `503 {"status":"not_ready","reason":"database unreachable"}`.
- Concurrency test asserts exactly `maxConcurrentDestructive` of 64 racing goroutines are admitted, and that refusals don't leak slots into a permanent lockout.

## Not fixed here

The root cause — `permanent=true` performing an unbounded cascade **synchronously inside the request** — remains. The standard shape is to mark and enqueue, letting a batched sweeper do the work (the webhook retention sweep in this repo already deletes in ctid-bounded chunks of 5,000). That changes the response contract, since the endpoint currently returns a real `messages_deleted` count, so it wants its own design rather than a patch.

Also worth noting independently: prod's Cloud SQL is `db-g1-small` — identical to staging — so prod had no more headroom than the environment that fell over.

### DIFF
--- a/cmd/e2a/main.go
+++ b/cmd/e2a/main.go
@@ -735,7 +735,11 @@ func main() {
 	// /readyz — instance-local readiness (DB reachable + migrations applied).
 	// Distinct from /api/health (shallow liveness); operational, not part of the
 	// /v1 contract, so it lives on this mux and never enters api/openapi.yaml.
-	router.HandleFunc("/readyz", readyzHandler(pool, &draining)).Methods(http.MethodGet)
+	// The verdict is refreshed on a background ticker, so this handler never
+	// competes for a pooled connection — see readyz.go for why that matters.
+	readiness := newReadinessMonitor(pool, &draining)
+	defer readiness.Stop()
+	router.HandleFunc("/readyz", readiness.handler()).Methods(http.MethodGet)
 	// /selftest — deep dependency diagnostics (health+json), auth-gated by the
 	// internal API secret. Operational, not part of the /v1 contract. The full
 	// SMTP→webhook round-trip lives in the external e2a-prober, not here.
--- a/cmd/e2a/readyz.go
+++ b/cmd/e2a/readyz.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"sort"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"time"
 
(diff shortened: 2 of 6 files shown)


jiashuoz (commented):
Reviewed the full diff plus surrounding context (`ratelimit.go` middleware, `account.go`, `main.go` wiring, and the HAProxy check config in e2a-ops). Solid incident-driven fix — the design holds up. Two findings, some nits.

## What checks out

- **Middleware branch is faithful to the existing pattern** — `resolvePrincipal` → `withPrincipal` handoff (no double auth downstream), auth failure falls through to the handler's canonical 401 instead of being masked as a 429.
- **429 contract is safe for clients** — Python SDK maps 429 → retryable `E2ARateLimitError` honoring `retry_after_seconds`, generated `agents_api` already declares `RateLimitedEnvelope` on these paths. Parallel bulk deletes back off instead of crashing.
- **Timing math works against the actual LB** — HAProxy runs default `inter 2s fall 3` (~6s to evict). The 15s grace means transient saturation never produces a single 503.
- **Slower eviction of a truly dead DB is harmless in this topology** — both slots share one Cloud SQL instance, so evicting a slot when the DB is down gains nothing. Trading ~6s → ~21s detection for saturation immunity is the right call.
- **Cap accounting is correct** — add-then-undo is race-safe, `defer releaseDestructive` survives panics, refusals don't leak slots (tested), keying on `p.User.ID` is right since ownership isn't resolved at middleware time.
- **Tests are non-vacuous** — `TestReadinessFlipsAfterGraceExpires` pins the other side of the tradeoff.

## Findings

**1. Medium — the probe still drinks from the poisoned well.** `probe()` calls `m.pool.Ping(ctx)`, which acquires a *pooled* connection. Under sustained saturation lasting **>15s**, the probe keeps timing out and the instance is evicted anyway — the original failure mode, deferred rather than eliminated. The delete cap makes that trigger much harder to hit, but the readiness half alone doesn't fully deliver "saturation is no longer indistinguishable from a dead database." Two cheap options: probe on a dedicated non-pooled connection held by the monitor, or check `pool.Stat()` first and treat "all conns in use" as *busy, not dead*. Not a merge blocker given the cap, but the PR body slightly oversells the readiness fix.

**2. Low — `deleteAccount` is the same shape and isn't capped.** `handleDeleteAccount` runs `DeleteUserData` synchronously with a per-table cascade and enqueues SES teardown for every owned domain inside the delete tx — the most unbounded cascade in the API. Lower risk (one account can't easily fire six meaningful concurrent account-deletes; the first removes the user row), but two racing calls before that lands both cascade. Either add it to `destructiveOps` or comment why it's excluded — right now the map's "nothing else in the API lets one call do an unbounded amount of write work" claim is false.

## Nits

- `probe()` maps a `latestMigrationApplied` *error* to `"database unreachable"` — a query failure on a reachable DB is mislabeled in logs.
- The destructive 429 omits the IETF `RateLimit-*` headers the other 429 path stamps. Defensible (not a quota), but inconsistent.
- `destructiveInFlight` never deletes keys — one `sync.Map` entry per account that has ever deleted, for process lifetime. Bounded; worth a one-line comment.
- Startup behavior shifts: the synchronous first probe runs during route setup, so with the DB down `main()` blocks up to 5s before binding the port. LB-equivalent (conn-refused vs 503), just worth knowing.

jiashuoz:
Thanks — finding 1 was right and is now fixed properly rather than papered over.

## Finding 1 (Medium) — fixed in `e3cbc4498`

You were correct that the readiness half didn't deliver what the PR body claimed. `pool.Ping` acquires a *pooled* connection, so a saturation outlasting the 15s grace starved the probe exactly as it starved the old inline check. The grace window deferred the original failure rather than eliminating it.

Took the stronger of your two options: the monitor now owns a connection built from the pool's `ConnConfig` and probes on that, so reachability is measured independently of pool pressure. I preferred it over the `pool.Stat()` heuristic because it removes the confound rather than classifying it — no need to decide what "all conns in use" means when the probe simply isn't competing for them. A stale socket is dropped and redialled next tick rather than reported as a permanent outage.

That opened a gap worth noting: a connection outside the pool **cannot see that the pool is broken**, and your point stands that such an instance can't serve. `TestReadyzHandler_DBUnreachable` (closes the pool, expects 503) caught it immediately. So pool usability is now checked separately — acquire with a 250ms deadline, where a **timeout means BUSY** (tolerated, that's the whole point) and any other error means closed/broken → not ready.

`TestReadinessProbeSurvivesPoolExhaustion` holds every pooled connection open with `grace=0`. Reverted to the pooled probe it fails with your exact predicted symptom:

```
status = 503, want 200: an exhausted pool made the probe report the database unreachable;
body={"status":"not_ready","reason":"database unreachable"}
```

**PR body corrected** — it no longer claims the grace window alone makes saturation indistinguishable from a dead database.

## Finding 2 (Low) — already fixed, 68s before your review

Race between your fetch and my push. `ce225231c` (04:08:07Z, review at 04:09:15Z) added **`deleteAccount`** and **`deleteImportBatch`**, and removed the "nothing else in the API" claim you correctly called false.

Your read of `deleteAccount` matches mine — it's the sharpest case, strictly worse than `deleteAgent`. I audited all twelve delete operations rather than trusting the incident, which is how `deleteImportBatch` (bulk-deletes a whole batch of contacts in one statement) turned up. There's now a test pinning the set in both directions, so a future unbounded delete has to opt out deliberately and bounded single-row deletes can't quietly acquire 429s.

## Nits

- **Migration error mislabelled** — fixed, now reports `"migration check failed"`.
- **`RateLimit-*` headers on the destructive 429** — leaving as-is. There's no quota to report; `RateLimit-Remaining` on a concurrency gate would be actively misleading, since the budget refills on completion rather than on a clock. `Retry-After` plus `max_concurrent` says the true thing.
- **`sync.Map` never deletes keys** — kept, comment added. Deleting on the last release would race a concurrent acquire for the same account; the entry is one pointer per account that has ever issued a capped delete, which is bounded by account count.
- **Startup blocks up to 5s with the DB down** — accepted and now documented. It's LB-equivalent (connection-refused vs 503), and the synchronous first probe is what preserves the deploy guarantee that a migration-less instance never reports ready.

Also confirming your topology point, since it's load-bearing for the 15s grace: both slots share one Cloud SQL instance (`db-g1-small` — prod and staging are the same tier), so evicting a slot when the database is down gains nothing.

PR #879, opened 2026-08-09
docs(design): async agent purge

jiashuoz:
Design for bounding permanent agent deletion, plus the review that reversed its original conclusion.

## Problem

`DELETE /v1/agents/{email}?permanent=true` ran **one transaction** ending in `DELETE FROM messages WHERE agent_id = $1`, with seven tables cascading off `messages`. Its duration, lock footprint and WAL volume all scaled with everything the inbox had ever accumulated, while holding a pooled connection.

Enough concurrent deletes exhaust the pool. A readiness probe that needs a pooled connection then stops answering, so a load balancer takes the instance out of rotation and serves its own `503` for every route — including endpoints that never touch the database — even though the database is merely busy. Reproduced in a non-production environment.

## The reversal

The document first proposed deferring oversized purges to a background job. **Adversarial review rejected that**, and the revision at the top of the doc records why.

The job's target was a reusable **name**, not a stable identity. The agent id *is* the email address, so a queued job could act on a different inbox that later occupied that address — specifically a **trashed** one, since the only guard rejected a live agent. Trash is exactly the state that promises restorability.

That's the ABA problem, and the design owns it: it named address-as-primary-key as the sharp edge, then reasoned only about reuse *latency* and dismissed a purge-state column as speculative. That column was the missing generation token.

Deferral is dropped rather than patched. Chunking is what bounds transaction duration — the thing that ends the long locks, WAL bursts and pinned xmin horizon that harm other tenants. Deferral additionally released a connection, at the cost of target identity, a discarded-job state machine that made a documented-irreversible delete reversible, and starvation of a small shared maintenance queue containing this design's own backstop.

## Two corrections recorded, not quietly fixed

- **The gutted-restore risk is not a cost of deferral.** It follows from giving up the single transaction, so it applies to chunking too. Atomicity and bounded transactions are mutually exclusive; the doc now says so.
- **"Inboxes are small enough not to need deferral" is marked unproven.** The metric that would settle it is per-inbox *retained-message stock* (max and tail percentiles). Nobody has measured it, and throughput over a window does not bound it, because messages are retained indefinitely. Review was right that the original rationale leaned on the wrong number.

Prerequisites for revisiting deferral are recorded so they are known rather than rediscovered.

## Reviewer note

Measurements in the doc come from a local fixture with only one of seven cascade children populated, and are labelled lower bounds. They support the narrow claim that committing between chunks substantially reduces transaction duration — not a production-safe threshold.

Implementation: #882.

### DIFF
--- a/docs/design/2026-08-09-async-agent-purge.md
+++ b/docs/design/2026-08-09-async-agent-purge.md
@@ -1,0 +1,183 @@
+# Bounded permanent agent purge
+
+Status: accepted
+Related: `docs/design/trash-soft-delete.md`, `docs/design/async-message-pipeline.md`
+
+## Problem
+
+Permanent agent deletion currently removes every message and its cascade
+children in one transaction. The transaction, lock set, WAL volume, River job
+cancellation work, and pooled-connection hold time therefore grow without a
+bound. The retention janitor uses the same unbounded shape.
+
+Splitting the work into committed chunks solves that resource problem, but it
+also removes the old all-or-nothing property. A safe chunked protocol must
+prevent four failures:
+
+- restoring an inbox after only part of it has been destroyed;
+- a stale purge deleting a later inbox that reused the same email address;
+- a message accepted between the last message chunk and parent deletion being
+  removed only by the foreign-key cascade, bypassing job cancellation,
+  storage accounting, and `messages_deleted`;
+- the threshold decision or retention janitor reintroducing an unbounded
+  transaction.
+
+## Goals
+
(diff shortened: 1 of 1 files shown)


jiashuoz:
Reviewed and approved.

The final design addresses the blocking concurrency and lifecycle issues: exact-incarnation purge tokens, restore refusal after claim, bounded classification and drain work, lock-holding final sealing, indexed active-send guards, and janitor resumption. The rewritten PR history is clean, all CI checks are green, and the design is ready to merge before #882.

PR #882, opened 2026-08-09
refactor(agents): chunk the oversized permanent delete instead of deferring it

jiashuoz:
Implements `docs/design/2026-08-09-async-agent-purge.md` as revised (#879 — merge first, or the doc references on this branch dangle).

> **Reworked.** This branch originally deferred oversized purges to a background job. Review found a critical data-loss defect in that approach, and the deferral was removed; only the chunking remains. The branch has since been rewritten to a single commit.

## Why the deferral was dropped

The job carried only the agent id as its target — and the agent id **is** the email address, which is reusable. The worker's only identity guard rejected a *live* agent, which cannot distinguish **the deletion the job was issued for** from **an unrelated later trashing of a different inbox at the same address**. Live inboxes were safe; **trashed** ones were not — and trash is exactly the state that promises restorability. Reproduced three ways, including across tenants.

That's the ABA problem. It was cheaply fixable, but deferral wasn't carrying its weight: chunking alone bounds transaction duration, which is what ends the long locks, WAL bursts and pinned xmin horizon that harm other tenants. Deferral additionally released one connection, while importing target identity, a discarded-job state machine, and starvation of a small shared maintenance queue containing this design's own backstop.

## What ships

`DeleteAgent` counts, then picks a shape. The threshold selects **atomic vs chunked**, not sync vs async:

- **≤ `InlinePurgeMaxMessages` (5000)** — the original single transaction, untouched. Fully atomic.
- **> threshold** — chunked in the same request, one transaction per chunk, ctid-bounded, for both `messages` and `contact_engagements` (the latter has no FK to `agent_identities`, so nothing cascades it).

## Zero contract change

```
$ git diff origin/main -- api/openapi.yaml sdks/ docs/api.md | wc -c
0
```

`cmd/e2a/main.go`, `internal/httpapi/httpapi.go`, `agents_write.go`, `delete_results.go` and `internal/apiserver/apiserver.go` are **byte-identical to main**. `messages_deleted` is the summed `RowsAffected` — exact, not a pre-count — so none of the async version's contract changes are needed. 22 files → 5.

## Reviewer attention

- **Every chunk re-reads the agent under `FOR UPDATE`** and aborts if it was restored or has a send in flight, and is **scoped by `user_id`** so a concurrently freed and re-registered address cannot be crossed into. Verified load-bearing: neutralising the predicate makes a non-owner drain delete another account's messages and remove their agent row.
- **Known residual (raised in review, not yet closed):** synchronous chunking narrows the ABA window to a single request but does not close it. If the janitor hard-deletes a past-retention row mid-drain and the address is re-registered *by the same owner* and trashed, a later chunk could act on the new incarnation. `user_id` scoping closes the cross-tenant case only, and `created_at` is not unique-constrained, so it is not a rigorous incarnation token. Narrow, but real.
- **Outbound send-job cancellation** rides the chunk's `DELETE ... RETURNING` in bounded batches rather than one unbounded prologue transaction.
- **Atomicity is deliberately lost above the threshold.** A mid-flight failure leaves a trashed, partially drained inbox; restoring it returns a gutted mailbox. Inherent to bounding transactions — you cannot have both. Shipped without a marker column, with a loud log. Review argued a marker is required; that decision is open.
- `deleteMessageChunkTx` concatenates a constant SQL fragment. Constants only, never user input, documented.

## Verification

`identity`, `httpapi`, `apiserver` green under `-tags=integration`. `make spec-check` and `make generate-sdk-check` clean. Mutation-tested: each guard confirmed red when removed, then reverted — including the restore-race guard and the cross-tenant guard.

## Known follow-up

A failed chunked delete is discoverable only via the log — nothing alerts on it. Worth a counter wired into the ops alerts.


jiashuoz:
Reviewed and approved.

I re-reviewed the final implementation at 9da6b788f, including the adversarial concurrency coverage. The prior blockers are addressed: pre-claim and post-claim ABA, late message and job writers, concurrent engagement updates, bounded River cancellation, indexed send-lease checks, restore refusal, and janitor resumption and accounting.

Local end-to-end verification also passed against an isolated server and database using synthetic data:

- An oversized permanent delete removed 5,001 messages and returned the exact `messages_deleted: 5001` receipt.
- The agent, messages, and engagement were removed while the account-level contact identity remained.
- The deleted address could be recreated.
- Restore during a claimed purge returned `409 purge_in_progress`.
- Retrying permanent deletion adopted the claim and completed successfully.

All CI checks are green. Ready to merge after #879.

PR #884, opened 2026-08-10
feat(plugin): conform plugins to Agent Plugins v1.0.0

jiashuoz:
## Summary

Makes the e2a plugin loadable by clients implementing [Agent Plugins v1.0.0](https://agent-plugins.org/) — the vendor-neutral packaging spec published 2026-08-06 by a TSC drawn from Amazon, Cursor, Microsoft, OpenAI, and Vercel. Today our four skills reach Claude Code, Codex, and Cursor through three hand-written manifests; Copilot, Kiro, and VS Code get the MCP server alone and none of the skills.

**Additive — no existing user journey changes.** Claude Code's manifest is only `.claude-plugin/plugin.json` and its MCP config only `.mcp.json` or inline, so it reads neither new file. Every legacy manifest and all three marketplaces are retained untouched. Design note: `docs/design/2026-08-09-agent-plugins-conformance.md`.

### Three changes

**1. Portable pair.** `plugin.json` + `mcp.json` at the plugin root. The v1 manifest schema is closed (`additionalProperties: false`), so `displayName`, `icon`, and `mcpServers` cannot ride along at the top level. `displayName`/`icon` are *omitted* rather than moved under `extensions` — an extension namespace is a client-owned identifier (spec §3, §8.1) and we are the plugin author, not a client, so `dev.e2a.*` would be data no client ever reads. Both keys already live in the client manifests. Transport token is `streamable-http`, where the legacy clients spell the same transport `http`.

**2. One metadata source.** Four plugin manifests plus three marketplaces repeated the same name, version, description, and license. `validate-plugin.mjs` compared them after the fact — that catches skew but not the hand-edit that caused it, and adding a fifth and sixth file would have made it worse. All nine are now generated from `plugins/e2a/plugin.meta.json`, with the validator running the generator in `--check` mode. Same freshness-gate shape as `make generate-sdk-check`.

**All seven pre-existing manifests regenerate byte-identically** — they do not appear in this diff.

**3. Skill path portability.** The `agentify`, `autopilot`, and `tether` skills located their own scripts through `${CLAUDE_PLUGIN_ROOT}` — 21 references. That variable expands only on Claude Code; everywhere else the agent is handed a literal unexpanded path, so the skill installs cleanly, validates cleanly, and fails at its first command. Each skill now names its scripts relative to its own directory, which is safe because every bundled script already self-locates via `BASH_SOURCE`.

### Review round (second commit)

An adversarial review found eight issues; each was reproduced before fixing. Four mattered:

- **The portable-pair check had a reachable hole.** Every URL rule sat behind `if (cfg.url)`, so removing `mcp.url` from the metadata source emitted a server entry with no URL — spec-invalid, silently skipped by clients, validator green. Exactly the failure the check exists to prevent. Now also rejects URL fragments, credential-bearing and case-duplicate headers, shell-string stdio commands, unanchored `cwd`, and wrong-typed metadata (spec §5.2 makes any schema violation beyond an unknown top-level field *fatal to the whole plugin*, so a number where a string belongs is an outage, not a cosmetic issue).
- **The client-variable guard only read `SKILL.md`.** Skills route the agent onward to `references/` and `scripts/`, where the same variable failed identically while passing the guard. Now scans every file in a skill, matching only interpolation forms so prose can still name the variable.
- **`extensions["dev.e2a.plugin"]` was the invented-namespace pattern** the spec's own migration guidance forbids — and the design note argued against exactly that for the legacy directories two sections earlier. Removed; the validator now rejects a namespace of ours.
- **`plugins/e2a/README.md` still instructed hand-editing `.claude-plugin/plugin.json`** as the version source of truth — a workflow the new freshness gate rejects. Rewritten.

New `scripts/validate-plugin.test.mjs` covers each hole with fixture-tree regressions, wired into the plugin CI job. Both scripts take an `E2A_PLUGIN_TREE` override so tests never touch `plugins/e2a`.

Also corrected two overstatements in the design note: it reported "23 pass, 0 fail" against the upstream checklist, but that was a local mechanization's tally, the checklist has 30 items, and the script was structurally blind to the extensions item it should have failed. Its lede asserted the standard's client list as though this plugin had been verified against it.

## Client surface checklist

Not applicable — no `/v1` handler, schema, or SDK surface is touched. This is plugin packaging plus skill documentation. `make generate` output is unaffected.

## Operational risk

Low, and deliberately bounded:

- No API, database, auth, billing, or delivery path is touched.
- No file is removed; reversible by deleting two generated files.
- The MCP endpoint is unchanged (`https://api.e2a.dev/mcp`).
- Agent Plugins v1 has no credential model, so no auth material is expressed in, or newly exposed by, the portable config.

**Two open gates, both documented rather than glossed:**

1. Conformance is validated statically. No live install on a spec-implementing client has been run, so this PR makes Copilot/Kiro/VS Code *reachable*, not *supported*.
2. Once Codex or Cursor implement the spec — both are on the TSC — the plugin will carry a native discovery path *and* the portable pair, i.e. two routes to the same MCP server. The spec is silent on coexistence. The additive analysis is verified for Claude Code only.

## Test plan

- [x] `node scripts/validate-plugin.mjs` — 4 skills, 9 manifests, portable pair OK
- [x] `node scripts/generate-plugin-manifests.mjs --check` — all 9 current
- [x] `node --test scripts/validate-plugin.test.mjs` — 9/9 guard regressions
- [x] `node --test scripts/plugin-agent-guidance.test.mjs` — 9/9
- [x] `plugins/e2a/skills/agentify/test/run.sh` — all pass
- [x] `node --test 'test/*.test.mjs'` in `skills/autopilot` — 117/117
- [x] Both generated files validated with `ajv` against the canonical v1.0.0 JSON Schemas
- [ ] Post-merge: live install on Claude Code, Codex, Cursor (no regression) then Copilot, Kiro (new reach) — and check gate 2 above at the same time

https://claude.ai/code/session_01B2a1Gi7bzCToYHUg66D8V5


jiashuoz:
**Review: Approve with nits.** I read the full diff, the Agent Plugins v1.0.0 spec (§1–§11 + Appendix A), and the plugin-authors guide, validated both generated files against the canonical schemas with ajv, and ran the verification suite on the branch (HEAD matches `255b6156`).

## Verified (all test-plan claims hold)

- `node scripts/validate-plugin.mjs` — pass, portable pair OK
- `node scripts/generate-plugin-manifests.mjs --check` — all 9 current
- Byte-identical regeneration of the 7 pre-existing manifests — write mode produces zero git diff
- `node --test scripts/validate-plugin.test.mjs` — 9/9
- `node --test scripts/plugin-agent-guidance.test.mjs` — 9/9
- ajv against the canonical `1.0.0` schemas — both valid

## Conformance (checked against the spec text, not just the checklist)

- plugin.json uses exactly the 10 closed-schema keys (§5.2); `name: "e2a"` satisfies §5.5 and the hand-rolled `AP_NAME_RE` byte-matches the canonical schema's `pattern`.
- `streamable-http` vs legacy `http` is a real gotcha, correctly handled (§7.2.1).
- URL rules (absolute HTTPS, no userinfo, no fragment, loopback exception) enforced (§7.2.1).
- plugin.json/mcp.json `$schema` version coupling check present (§10.1); pinning `1.0.0` is safe since canonical identifiers can't be reassigned.
- Omitting `extensions` is the right call — namespaces are client-owned (§3, §8) — and the own-namespace rejection is a good regression guard.
- Skills remain at the fixed `skills/<name>/SKILL.md` location (§7.1).

## On the goto patterns

1. Single metadata source + generator + CI freshness gate — mirrors the existing `generate-sdk-check`; correct choice over post-hoc drift comparison.
2. Skill path portability via self-substitution + `BASH_SOURCE` self-location — this is the only portable option, since the spec defines `PLUGIN_ROOT`/`PLUGIN_DATA` only for stdio MCP subprocess env (§9), never for skill execution. Using them in a SKILL.md would be the same bug this fixes.
3. Omitting client-only metadata rather than inventing a namespace — matches the spec's own migration guidance.

One deliberate trade worth stating: the validator re-implements a slice of the canonical schema by hand (name regex, type checks, closed author object). Justified here — no CI runtime schema dependency, and the pinned `$schema` limits drift — but it should stay pinned-and-documented, which it is.

## Nits

1. **Incorrect spec citation in a code comment** — `validate-plugin.mjs` (~L847) says "a stray top-level key is a hard reject of the whole plugin." Per §5.2 an unknown top-level field is explicitly *non-fatal*: clients "MUST report and ignore each unknown field and MUST continue loading." The stricter behavior is fine as author hygiene (the closed schema exists to catch typos), but the comment states the wrong severity. The design note and PR body got this right; only the comment is wrong.
2. **Validator crashes on a non-object server entry** — `validate-plugin.mjs:325` reads `cfg.type` before any type guard; a hand-edited `mcp.json` with `"e2a": null` throws an uncaught TypeError instead of a clean error. This is exactly the "validator green, real failure" class the test suite targets, and the fixture harness makes it a one-line regression test.
3. **Style** — `validate-plugin.test.mjs` reads files via `spawnSync("cat", …)` while `readFileSync` is already imported.
4. **Over-strict vs spec (note-only, both harmless today):** validator requires ≥1 `mcpServers` entry (spec allows empty); doesn't check stray top-level keys in `mcp.json` (only plugin.json, currently covered by the freshness gate); loopback allowlist is exactly `127.0.0.1` though the spec allows any loopback-range literal.

None block merge.

PR #888, opened 2026-08-11
fix(sender-identity): make fixture teardown converge safely

jiashuoz:
## Summary
- permanently purge custom-domain fixture agents before deleting their domains
- commit domain DELETE with a durable teardown job, then confirm provider absence best-effort before returning (async convergence is the guarantee — see the review-fix commits)
- serialize create, delete, re-registration, status polling, and BYODKIM refresh by desired-state incarnation
- retain an ownership ledger and versioned River jobs for retry exhaustion and blue/green overlap
- tag managed SES identities and refuse to mutate or delete unowned identities
- make Cloudflare fixture cleanup retryable, fixture-scoped, and honest on ambiguous failures

## Why
The conformance harness previously trashed an agent, ignored the FK-blocked domain deletion, and then removed DNS underneath a live SES identity. AWS later emitted DKIM and custom MAIL FROM failure notices. The API success boundary now means the row delete plus a durable SES teardown job are committed; with a healthy provider the SES identity is additionally confirmed absent before the response returns, so DNS cleanup still follows provider absence in practice. DNS is preserved on every failure path.

## Rollout gate
This release is paired with tokencanopy/e2a-ops#310. Existing SES identities predate the ownership tag, so operators must inventory the approved database set against SES, tag only confirmed legacy e2a identities, import/review/apply the strict IAM policy, wait out old-worker requests, and repeat inventory before deploying this OSS revision.

## Verification
- go test -count=1 -p 1 ./...
- go test -race ./internal/senderidentity ./internal/identity ./internal/apiserver ./internal/jobs
- npm run test:unit (54 passed)
- npm run typecheck
- terraform companion tests: 2 passed
- git diff --check

The max-conns=1 and delete/re-registration database regressions run when the PostgreSQL test service is available and are covered by CI.


jiashuoz:
## Review fixes (11 commits, `73423eb` … `5139d80`)

A 10-finding code review plus an independent + adversarial re-review pass landed on this branch. Summary of what changed on top of the original commits:

**Correctness**
- `fix(domains)`: the synchronous SES `Deprovision` moved out of the delete transaction. The tx now commits the row delete + durable teardown job (success boundary independent of SES); a bounded post-commit best-effort deprovision preserves the "provider absent before the 200" property with a healthy provider. Fixes: foreign/untagged identity made a domain permanently undeletable (500), deletes failing during SES outages, and a resurrection window after an irreversible provider delete.
- `fix(sender-identity)`: forced re-checks (`POST /verify`) of a healthy domain are now a read-only no-op (ledger `applied_incarnation` + one provider GET), restoring the deleted `AlreadyVerifiedNoOp` guard against verified→pending flaps; a transient GET error retries instead of converging blind.
- Events: reap/audit workers now fire `domain.sending_*` webhooks; terminal events survive a failed ledger `Forget`; and firing is gated on genuine status *transitions* — the adversarial pass proved the naive fix emitted one duplicate webhook per River retry and per hourly sweep of a stuck-failed domain.
- Post-drain audit dedupe is scoped by schedule window (River can't drop `completed` from unique state sets), so a re-registered-then-deleted domain gets its 15-minute audit again.
- The pre-ledger orphan ALERT is restored (alert-only) for provider identities that are neither ledgered nor backed by a domain row; orphan-check DB errors log instead of failing the sweep.
- The process-wide mutation gate honors context cancellation (capacity-1 channel instead of `sync.Mutex`).
- `ReapV2` runs on start, making the ledger sweep the version-agnostic backstop for v2 jobs stranded by a rollback (contract documented in the design doc).
- e2e-prod cleanup: the ten-minute 409-conflict budget is now monotonic — a 429/5xx blip mid-wait no longer abandons the fixture.

**Contract/docs**
- `deleteDomain` description corrected (the old text was unreachable/overclaiming); `api/openapi.yaml` + generated TS/Python SDK layers regenerated.
- Design doc updated: delete semantics, rollback contract, healthy-recheck caveat (the gate must be revisited by any future in-place DKIM rotation or MAIL FROM change).

**Verified**: package tests green (`senderidentity`, `apiserver`, `httpapi`, `identity`, harness 56/56), spec golden in sync, over-the-wire e2e against a locally booted binary (register/verify/delete lifecycle, auth/confirm guards, live spec). Known pre-existing, unrelated: a time-of-day flake family (engagement/outreach/inbound-activity tests) fails identically at the base commit and on `origin/main`.

**Deferred** (pre-existing or follow-up): domain-case normalization on the delete path feeding provider calls verbatim; double provider GET on the audit path; ops-side check of the SES account's identity set before wiring log alerts on the restored `ALERT orphan sending identity` line (it fires hourly for any non-e2a identity in the account); the time-of-day test flake above.

jiashuoz:
## Second review round — dispositions (4 new commits, `7db605b` … `6fd253f`)

Verdict was "needs revision" with 4 blockers, 1 should-fix, 1 CI blocker. Disposition of each:

**1. Early tombstone finalization — FIXED, and more thoroughly than the bucket framing.** The hole exists even without dedupe: an *earlier* mutation's audit (or the hourly sweep) can finalize a *later* mutation's tombstone inside its drain window. The guard is now persisted in the ledger itself: `FinalizeSendingIdentityTombstone` deletes the row only when `updated_at` is older than the drain window, and **both delete paths** (per-domain DELETE and the account cascade) stamp `updated_at` in the delete tx — without that stamp a long-lived domain's row is already "old" at delete time and the guard would never engage. A younger tombstone survives as a no-op; that mutation's audit or the next sweep finalizes it. SQL-level, worker-level, and race-scenario tests added.

**2. Rollback convergence — DOCUMENTED AS ROLL-FORWARD-ONLY, consumers-first rejected with rationale.** Producing legacy kinds from the new binary would let the old slot claim new-style mutations with its blind delete semantics mid-overlap — strictly worse than bounded stranding. The stale-legacy-deprovision ABA hazard on rollback is the *pre-existing* prod behavior returning, not new exposure. The design doc now states the operator requirement plainly (auto-rollback of this release = re-deploy a fixed candidate promptly) and it's step 6 of the upgrade runbook. If you'd rather have a flag-gated two-phase producer switch instead, say so — it's a scope decision I didn't take unilaterally.

**3. Stuck-pending invisible to the reaper — FIXED.** Confirmed real: applied+present+pending with a lost reconcile budget was permanently unpollable. The non-forced sweep now treats exactly that state as its one read-only inspection: a provider `Status` GET resolves it to verified/failed (committing + firing the genuine transition), still-pending waits for the next sweep, absent/foreign falls through to full convergence. Verified/failed steady states keep the zero-provider-call fast path — forcing *every* candidate (the suggested alternative) would have re-fired `sending_failed` hourly for stuck-failed domains via the reconcile transition.

**4. DNS removal not gated on teardown state — FIXED end-to-end.** `deleteDomain` now returns `sending_teardown: "confirmed" | "pending"` (open set per the response-enum policy; the tagged-enum version failed `TestResponseEnumsAreOpenSets`). The harness threads the terminal delete body through `cleanupFixtures`; on `"pending"` it retains every DNS record, reports them, and keeps the fixture tracked. Spec + generated SDKs refreshed; absent field (older server, bodiless 204/404) keeps prior behavior. Verified over the wire against a locally booted binary.

**5. No provider-operation deadline — REFUTED on the premise, metrics deferred.** River's default `JobTimeout` (1 minute) applies to every sender-identity worker — `internal/jobs` doesn't override it — so worker-held gate/connection/AWS time is already ctx-bounded; the HTTP path is bounded at 10s. Gate-wait/provider-duration metrics are a real gap but belong with the metrics registry work — deferred with that scope, not silently.

**6. Coverage gate timeout — FIXED.** `cover` now runs with `-timeout 15m` (justified in the Makefile comment, under the job's 18-minute cap). The failure was the default 10-minute per-package timeout under atomic instrumentation + shared-DB contention, matching your read.

Verification at head: `senderidentity` / `apiserver` / `httpapi` / `jobs` green; full `identity` green modulo one shared-instance contention artifact that passes ×2 in isolation (my concurrent e2e boot was the aggressor) plus the previously-declared time-of-day flake family (fails identically on `origin/main`); harness 58/58 + typecheck clean; spec golden in sync; over-the-wire register→delete shows the new field with clean logs.

Agreed on ops PR #310 as a production prerequisite (tag inventory + IAM before activation) — that gate is unchanged by these commits.

jiashuoz:
## Third review round (Codex #2) — dispositions (4 commits, `f1cc4d8` … `d443995`)

**Blocker 1 (teardown confirmation not authoritative) — FIXED, with the reproduced sequence as the regression test.** DNS removal now requires an explicit `sending_teardown:"confirmed"`; pending, malformed bodies, missing fields, and unknown open-set values all fail closed. A per-process pending memory makes the exact repro (pass 1 pending → pass 2 bodiless 404) retain DNS — the 404 proves nothing once the domain has been seen pending. One deliberate exception: a 404 with **no** pending history proceeds, because that's the registration-never-succeeded case where no identity can exist and requiring `confirmed` would leak DNS for every such fixture. On the server side, `confirmed` is now honestly documented as "e2a-managed teardown complete (removed, or never e2a's to remove)". The *durable cross-process receipt* is deliberately deferred: cross-process retries now fail closed (retain + report — consistent with this module's honest-reporter philosophy), and a pollable receipt needs an API-shape decision (idempotent re-DELETE receipt vs. a teardown endpoint on a deleted resource, with anti-enumeration constraints) that shouldn't be improvised inside a fix round — flagged for Josh.

**Blocker 2 (rollback mechanically unsafe) — FIXED with the consumers-first equivalent you asked for.** New `sender_identity.legacy_job_compat` splits the producer flip from the binary deploy: phase 1 deploys this release with the flag on — it produces the **legacy** job kinds (mutations *and* reconcile budgets), consumable by the previous release, so the bake gate's automatic rollback strands no teardown work; the old slot claiming a legacy job mid-overlap is today's production behavior, not new risk. Phase 2 flips the flag off in a config-only deploy whose rollback target is the phase-1 binary, which consumes v2. Unit tests pin the kind/queue selection per mode. Residual (documented): v2-only post-drain audit *accelerators* can strand on a phase-1 rollback — bounded by the ledger + orphan ALERT + RunOnStart sweep. Your IAM point is also recorded as a hard ordering rule: apply #310's tag-conditioned policy no earlier than phase 1 going stable, since a rolled-back pre-tagging binary cannot create identities under it.

**Should-fixes:**
- *Verified drift invisible* — FIXED: the hourly sweep's read-only GET now covers `verified` too; a definitive provider `failed` (tag removed, DNS pulled) commits with axes and fires within the hour; a provider mid-recheck `pending` is deliberately not committed (no flap).
- *Unbounded hourly pending polls* — FIXED with a persisted absolute deadline, as suggested: the ledger row's last-mutation time anchors a 24h backstop TTL (well beyond River's exhausted-budget horizon); older stuck-pendings time out as `verification timed out` with the event.
- *Required field breaks older-server compat* — FIXED: `sending_teardown` is schema-optional (`omitempty`); the regenerated Python model is `Optional[...] = None`.
- *now() is transaction-start time* — FIXED: the tombstone stamp uses `clock_timestamp()` and runs after the job insert in both delete paths, so a long advisory-lock wait can no longer commit a pre-aged stamp.
- *gofmt* — FIXED for the three files this branch touched (`internal/httpapi/webhooks.go` is pre-existing drift on `origin/main`, left alone).

Verification at head: `senderidentity` / `apiserver` / `httpapi` / `config` / `jobs` / targeted `identity` green; harness 61/61 + typecheck; spec golden in sync; SDKs regenerated. Agreed #310 + the legacy tag inventory remain activation prerequisites, now with the explicit phase-1-first ordering.

PR #901, opened 2026-08-16
fix(identity): close the check-then-act race on max_domains

### LINKED ISSUE #822: limits: max_domains is not enforced under concurrency (check-then-act, and the endpoint has no rate limiter)
## What

`POST /v1/domains` enforces `max_domains` with an unsynchronized check-then-act: `EnforceDomainCreate` counts rows (`internal/limits/enforcer.go:200-218`), then `ClaimDomain` inserts (`internal/httpapi/domains.go:506-514`). There is no transaction spanning the two, no lock, and no constraint expressing the cap, so concurrent requests all observe the pre-insert count and all pass.

## Evidence

Measured during a review of #819 against a real Postgres: an account with `max_domains: 1` issuing **12 concurrent `POST /v1/domains`** with distinct names ends up with **12 rows**. Identical before and after #819 — that PR neither introduced nor widened this (it only skips the check for a domain the caller already owns, which creates nothing).

Compounding it, the endpoint has **no rate limiter**: `internal/httpapi/ratelimit.go:80` covers only `createAgent`. The overshoot is bounded only by client parallelism.

## Why it matters

`max_domains` is a billing boundary. An account can currently exceed it by an arbitrary factor with no special tooling, and nothing after the fact reconciles the count down.

The same shape almost certainly applies to `EnforceAgentCreate` (`internal/httpapi/agents_write.go:370`) — worth checking as part of any fix, though agent creation at least sits behind the rate limiter.

## Options

1. Enforce inside the claim transaction — the count and the insert become atomic. Needs the enforcer to accept a `pgx.Tx`, which crosses a package boundary today.
2. A partial unique index or a per-account counter row updated in the same transaction.
3. Advisory lock per account for the duration of the claim — cheapest, and `ClaimOrCreateDomain` already takes advisory locks for the domain hierarchy.

Option 3 looks closest to the existing grain.

## Not covered by tests today

Nothing exercises concurrent cap enforcement in any suite. A regression test would need N goroutines against one account and an assertion on the final row count.

AmirF194:
## Summary

`CheckDomainCreate` (the limits pre-check) and `ClaimOrCreateDomain` (the insert) ran as
two independent transactions with nothing serializing them: concurrent `POST /v1/domains`
requests could all read the same pre-insert count and all pass, so `max_domains` could be
exceeded by an arbitrary factor. Reported with a live reproduction against real Postgres
(12 concurrent requests all succeeded against a cap of 1).

`ClaimOrCreateDomainWithLimit` now takes a per-user advisory lock and re-checks the count
inside the same transaction as the `INSERT`, so the check and the write can no longer
interleave across requests. `EnforceDomainCreate` stays as a fast, richly-detailed
pre-check (plan code, upgrade URL); the transactional check inside `ClaimDomain` is the
race-proof source of truth for the cap itself.

## Operational risk

No schema change, no behavior change on the non-concurrent path. A request that
previously slipped past the cap under a race now gets `402 limit_exceeded`, the same
response the endpoint already returns for a non-racy over-cap request.

## Test plan

- [x] New e2e test (`internal/e2e`, `-tags integration`): 8 concurrent `POST /v1/domains`
      against `max_domains=1`: fails on `main` (8/8 succeed, cap blown 8x), passes on this
      branch (1 created, 7 `402 limit_exceeded`), both against a real Postgres 16 container
- [x] `go test ./internal/identity/...` and `./internal/httpapi/...` green, including two
      new unit tests for the `ClaimDomain`/`GetLimits` error-mapping paths
- [x] `go test -p 4 -covermode=atomic ./internal/...` (the `Go coverage gate` job's own
      command) green
- [x] Not verified: contention beyond 8 concurrent requests, or across more than one
      connection pool / process

Fixes #822

### DIFF
--- a/internal/apiserver/apiserver.go
+++ b/internal/apiserver/apiserver.go
@@ -189,7 +189,7 @@ func BuildDeps(p Params) httpapi.Deps {
 
 		ListDomains:                  p.Store.ListDomainsByUser,
 		SendingRampSnapshot:          rampSnapshot,
-		ClaimDomain:                  p.Store.ClaimOrCreateDomain,
+		ClaimDomain:                  p.Store.ClaimOrCreateDomainWithLimit,
 		EnforceDomainCreate:          p.Enforcer.CheckDomainCreate,
 		DeleteDomain:                 deleteDomainFunc(p),
 		LookupDomainTeardownSnapshot: p.Store.LookupDomainTeardownSnapshot,
--- a/internal/e2e/domains_create_race_e2e_test.go
+++ b/internal/e2e/domains_create_race_e2e_test.go
@@ -1,0 +1,96 @@
+//go:build integration
+
+package e2e_test
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"net/http"
+	"sync"
+	"testing"
+
+	"github.com/tokencanopy/e2a/internal/limits"
+	"github.com/tokencanopy/e2a/internal/testutil"
+)
+
+// max_domains must be enforced under concurrency: N simultaneous POST
+// /v1/domains for distinct new domains on one account, capped at 1, must
+// leave exactly one domain claimed (#822). Before the fix, EnforceDomainCreate's
+// count read and ClaimDomain's insert were two independent DB calls with
+// nothing serializing them, so every concurrent request could read the same
+// pre-insert count and all pass the cap.
+func TestRegisterDomainConcurrentRequestsRespectMaxDomainsE2E(t *testing.T) {
+	pool := testutil.TestDB(t)
+	ts := testutil.TestServer(t, pool)
+	ctx := context.Background()
+
+	user, err := ts.Store.CreateOrGetUser(ctx, "race-owner@example.com", "Race Owner", "google-race-owner")
+	if err != nil {
+		t.Fatalf("CreateOrGetUser: %v", err)
+	}
+	apiKey, err := ts.Store.CreateAPIKey(ctx, user.ID, "race-key", nil)
+	if err != nil {
+		t.Fatalf("CreateAPIKey: %v", err)
+	}
+	if err := limits.NewStore(pool).Upsert(ctx, user.ID, limits.Limits{
+		PlanCode: "test", MaxAgents: 100000, MaxDomains: 1,
+		MaxMessagesMonth: 100000, MaxStorageBytes: 1 << 40,
+	}); err != nil {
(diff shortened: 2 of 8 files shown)


jiashuoz (approved):
Solid fix — the race is real, the advisory-lock approach is sound, and the test coverage (including a fails-on-main e2e repro) is exactly what this needed. Approving.

Verified while reviewing:

- Lock design: `hashtextextended(userID, 1)` is a distinct keyspace from the seed-0 domain locks and the `sender-identity:` session lock; ordering (domain locks → user lock) is consistent and the user lock is taken in exactly one place, so no deadlock cycle is possible. Under READ COMMITTED the post-lock `count(*)` sees every insert committed before lock acquisition — properly serialized across connections, pools, and processes on the same DB.
- The in-tx count is byte-identical to `usage.CountDomainsByUser`, so the backstop and `CheckDomainCreate` can't disagree about what consumes a slot.
- Cap is only charged on genuine inserts; same-owner re-claims return before the count check, preserving idempotent re-register at cap.
- `GetLimits` failure → 500 before `ClaimDomain` is called (fail-safe, and pinned by the new unit test), and `DomainLimitExceededError` maps to the same `402 limit_exceeded` shape clients already get from `limitEnvelope`. No API-shape change, so no contract-scenario/spec obligations.

Two non-blocking suggestions:

1. Consider fetching `GetLimits` unconditionally in `handleRegisterDomain` (while keeping the `EnforceDomainCreate` pre-check skipped on the `alreadyOwned` path). Today `alreadyOwned` passes `maxDomains=0`, so a `DELETE` racing a re-claim of the same domain would fall through to the insert branch with the cap check disabled. Contrived, but the store only charges on genuine inserts anyway, so passing the real limit costs nothing.
2. The `maxDomains <= 0` = "unlimited" sentinel inverts the pre-check's semantics (where `MaxDomains=0` blocks everything). Unreachable today since a zero-cap plan 402s in the pre-check first, but worth a one-line comment at the lock site so a future zero-cap plan doesn't silently disable the backstop.

Nit: the 500 message for a `GetLimits` failure is identical to the `EnforceDomainCreate` failure message ("limits check failed") — distinguishing them would ease log triage.

AmirF194:
Thanks for the thorough review, Josh, and for catching the lock-ordering and idempotent-reclaim details. Good calls on both suggestions, I will follow up with a small PR for the unconditional `GetLimits` fetch and a comment at the lock site on the `maxDomains <= 0` sentinel.

PR #903, opened 2026-08-16
fix(identity): check domain ownership before the plan cap on register

### LINKED ISSUE #825: api: 402 precedes 409 on POST /v1/domains — an at-cap account is told to upgrade for a domain it can never have
## What

`handleRegisterDomain` checks the plan cap before attempting the claim, so an account at `max_domains` that registers a domain **another account owns** gets `402 limit_exceeded` rather than `409 domain_taken`.

402 tells the caller "upgrade and retry." That is false here: upgrading will never get them someone else's domain.

Pre-existing — the cap already ran first before #819, and that PR did not move it. It is *more* visible now, because after #819 the 402 branch on an owned-looking domain means exclusively "you do not own this."

## Why conflict should win

The handler's own comment states the principle (`internal/httpapi/agents_write.go:368-369`): a 402 must mean "valid request, out of capacity" and must "never mask a 400/401". A 409 belongs in that same set — it is a permanent property of the request, not a capacity condition.

The codebase already gets this right for contacts: both the store and the handler prefer `ErrContactExists` over the cap (`internal/identity/contacts.go:173-185`, `internal/httpapi/contacts.go:387-393`).

## Cost

Fixing it needs an existence probe that is **not** user-scoped, before the cap check — `LookupDomain` filters on `user_id` by design, so it cannot see another account's row. `DomainOwner` (`internal/identity/store.go:1076`) or `DomainExists` (`:1095`) would serve. That is an extra query on every registration, which is why this is an issue rather than a drive-by change.

Note the ordering also has to stay compatible with the parent/child cross-account guard inside `ClaimOrCreateDomain`, which rejects a claim that *overlaps* another account's domain, not just an exact match.

AmirF194:
## Summary

`handleRegisterDomain` checked the caller's plan cap before checking domain
ownership. `EnforceDomainCreate` ran first and rejected any at-cap caller
with 402 `limit_exceeded`, even when the domain they were asking for was
already owned by someone else: `ClaimDomain` (`claimOrCreateDomain`), the
call that actually detects that conflict, never ran. The caller got a false,
unactionable "upgrade your plan" error for a request no plan upgrade would
fix; the real answer is 409 `domain_taken`.

`claimOrCreateDomain` already checks the cross-account conflict before the
cap internally, so this is a matter of not pre-empting it: `EnforceDomainCreate`
now runs only after `ClaimDomain` reports a genuine `DomainLimitExceededError`,
where it enriches the 402 with plan/upgrade-URL details instead of gating the
request up front.

## Operational risk

None. No schema, endpoint, or client-facing contract change: same two error
codes (`limit_exceeded`, `domain_taken`) as before, just attributed to the
right condition. No migration.

## Test plan

- `TestRegisterDomainAtCapStillReturnsConflictForAnotherAccountsDomain`
  (new): an at-cap caller registering a domain owned by another account now
  gets 409 `domain_taken`; fails on `main` (got 402 `limit_exceeded`), passes
  on this branch.
- `go test ./internal/httpapi/... ./internal/identity/... ./internal/apiserver/...`
  and the full `go test ./...` both green.
- `go tool cover -func` confirms every line touched in `handleRegisterDomain`
  is exercised by the suite (coverage for the package: 87%, above the 73%
  CI floor).
- Not verified: contention between `ClaimDomain`'s own cap check and a
  concurrent `EnforceDomainCreate` call under a real race (the two now run
  sequentially against the same store, same as before this PR).

Fixes #825

### DIFF
--- a/internal/httpapi/domains.go
+++ b/internal/httpapi/domains.go
@@ -511,25 +511,14 @@ func (s *Server) handleRegisterDomain(ctx context.Context, in *registerDomainInp
 		}
 	}
 	maxDomains := 0
-	if !alreadyOwned {
-		// A richly-detailed pre-check (plan code, upgrade URL); ClaimDomain
-		// below is the race-proof source of truth for the cap itself.
-		if s.deps.EnforceDomainCreate != nil {
-			if err := s.deps.EnforceDomainCreate(ctx, user.ID); err != nil {
-				if env, ok := limitEnvelope(err); ok {
-					return nil, env
-				}
-				return nil, NewError(http.StatusInternalServerError, "internal_error", "limits check failed")
-			}
+	if !alreadyOwned && s.deps.GetLimits != nil {
+		lim, err := s.deps.GetLimits(ctx, user.ID)
+		if err != nil {
+			return nil, NewError(http.StatusInternalServerError, "internal_error", "limits check failed")
 		}
-		if s.deps.GetLimits != nil {
-			lim, err := s.deps.GetLimits(ctx, user.ID)
-			if err != nil {
-				return nil, NewError(http.StatusInternalServerError, "internal_error", "limits check failed")
-			}
-			maxDomains = lim.MaxDomains
-		}
+		maxDomains = lim.MaxDomains
 	}
+	// ClaimDomain checks a cross-account conflict before the cap (#825).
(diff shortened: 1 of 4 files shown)


jiashuoz (approved):
Thanks for the fix!!

AmirF194:
Thanks for the quick review and merge!

PR #906, opened 2026-08-17
fix(identity): skip DKIM keygen on a domain reclaim

### LINKED ISSUE #826: identity: ClaimOrCreateDomain generates an RSA-2048 DKIM keypair before checking whether the row already exists
## What

`ClaimOrCreateDomain` calls `dkim.GenerateKeypair()` (`internal/dkim/dkim.go:73-74`, `rsa.GenerateKey(rand.Reader, 2048)`) plus an AEAD seal on **every** invocation, before the transaction opens. On the existing-row branch — a same-owner re-register, which returns the existing row and creates nothing — the keypair is discarded.

## Why now

Pre-existing, but #819 changes who can reach it. Before, an account at `max_domains` was rejected by two indexed queries and never got as far as the keygen. After, a re-register by an at-cap account proceeds — and `POST /v1/domains` has **no rate limiter** (`internal/httpapi/ratelimit.go:80` covers only `createAgent`), so that caller can loop a 2048-bit keygen per request.

Largely academic in practice: the default `MaxDomains` is `1_000_000` (`internal/config/config.go:356`), so under-cap accounts have had this path available all along. It does remove the one incidental cost gate that class of caller had.

## Fix

Move the keygen inside the `pgx.ErrNoRows` branch, so it only runs when a row is actually being inserted. The existing-row `SELECT` does not consume it. Worth confirming no caller depends on a keypair being generated as a side effect of a re-claim.

AmirF194:
## Summary

`claimOrCreateDomain` generated a fresh RSA-2048 DKIM keypair (plus the AEAD
seal over its private key, #144) before checking whether the domain row
already existed. On a reclaim, the common idempotent path the function's own
docstring describes ("the DKIM keypair are minted on first INSERT and remain
stable across re-claims"), the SELECT branch returns the existing row and the
freshly generated keypair is discarded, unused. Every re-register of an
already-owned domain paid for a full RSA-2048 keygen for nothing.

The fix moves the keygen/seal block into the `pgx.ErrNoRows` branch, right
before the INSERT that is its only consumer, so it runs once per domain
instead of once per call. Stored data is byte-for-byte identical either way;
this only removes the wasted work.

## Operational risk

None. No stored data changes, no API response changes. The keygen/seal now
runs while holding the per-domain advisory lock (previously it ran before the
transaction started), which only extends lock hold time on the once-per-domain
new-row path; the existing `TestClaimOrCreateDomain_ConcurrentParentChildClaimsDoNotSplitOwnership`
and `TestClaimOrCreateDomain_HierarchicalClaimsAreExclusiveAcrossAccounts`
tests still pass unmodified.

## Test plan

- [x] New regression test `TestClaimOrCreateDomain_ReclaimSkipsDKIMKeygen`
  (`internal/identity/store_test.go`): swaps `crypto/rand.Reader` for a
  byte-counter around a reclaim call only. Fails on unmodified `main`
  (120977 bytes read, a real RSA-2048 keygen) and passes on the branch
  (well under the 256-byte threshold, which is comfortably above
  `generateID`'s own unrelated 16-byte read on every call).
- [x] `go vet ./internal/identity/...` and `gofmt -l` on both changed files:
  clean.
- [x] Full `go test ./internal/identity/... ./internal/dkim/...`: all
  passing (505s/1s), including the pre-existing
  `TestClaimOrCreateDomain_StableOnReclaim` (DKIM key/verification token
  stability across reclaim, unaffected) and the two concurrency tests named
  above.
- [x] Docker, Go 1.26 (matching CI's `go-version: "1.26"`) against a
  `postgres:16-alpine` service on `:5433` with the CI's own credentials, the
  same shape `go-coverage`/`go-e2e` use.
- Not run: `make test-e2e` (the separately tagged integration suite) and the
  web/TS/Python/CLI/MCP jobs, none of which touch this file.

Fixes #826

### DIFF
--- a/internal/identity/store.go
+++ b/internal/identity/store.go
@@ -819,31 +819,6 @@ func (s *Store) claimOrCreateDomain(ctx context.Context, domain, userID string, 
 
 	verificationToken := "e2a-verify=" + generateID()
 
-	// Generate a DKIM keypair for this domain. Failures here are
-	// non-fatal — the columns are nullable and the outbound signer
-	// treats a missing key as "skip DKIM". We still log because key gen
-	// failing is a hard signal (entropy exhaustion or an OS-level
-	// CSPRNG bug) that ops should see.
-	var dkimSelector string
-	var dkimPubKey string
-	var dkimPrivKey []byte
-	if kp, kerr := dkim.GenerateKeypair(); kerr == nil {
-		// Encrypt the private key at rest (#144). On seal failure (catastrophic
-		// RNG) drop ALL three DKIM columns so we never publish a public key /
-		// selector without a usable private key — non-fatal, same posture as a
-		// keygen failure (the signer treats a missing key as "skip DKIM").
-		sealed, serr := s.sealDKIM(kp.PrivateKeyDER, domain)
-		if serr != nil {
-			log.Printf("[identity] dkim key seal failed for %s: %v", domain, serr)
-		} else {
-			dkimSelector = kp.Selector
-			dkimPubKey = kp.PublicKeyDNS
-			dkimPrivKey = sealed
-		}
-	} else {
-		log.Printf("[identity] dkim keygen failed for %s: %v", domain, kerr)
-	}
-
 	tx, err := s.pool.BeginTx(ctx, pgx.TxOptions{})
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


jiashuoz (approved):

AmirF194:
Thanks for merging!

Issue #792, opened 2026-08-01
Harden client URL building against dot-segment path collapse (latent destructive-request retarget)


jiashuoz:
## Summary

Client code that interpolates an identifier into a URL **path segment** can be silently retargeted at a different endpoint. `encodeURIComponent` does not encode `.`, so a value of `..` survives escaping, and the URL parser then resolves it as "go up one level" — **before the request leaves the client**. The server never sees the `..`; it receives an ordinary-looking request to the parent resource.

Where the parent is itself a destructive endpoint, a routine delete becomes a much larger one, and it carries a valid confirmation token with it because `?confirm=DELETE` is the same literal on every delete op.

**This is latent, not live** — see "Why nothing is broken today". Filing it because the only thing currently preventing it is a routing-library default that nothing in the codebase asserts or tests.

## Mechanism

```
encodeURIComponent("..")                    -> ".."            (dots are unreserved)
new URL("/v1/account/suppressions/..").pathname -> "/v1/account/"
```

Since `..` also consumes the segment before it, every two-segment delete resolves one level up. Mapping the collapse target of each `DELETE` route against the set of `DELETE` routes:

| Route | Collapses to | Parent is a DELETE endpoint? |
|---|---|---|
| `/v1/account/suppressions/{address}` | `/v1/account` | **yes — deletes the account** |
| `/v1/account/api-keys/{id}` | `/v1/account` | **yes — deletes the account** |
| `/v1/agents/{email}/suppressions/{address}` | `/v1/agents/{email}` | **yes — deletes the agent** |
| `/v1/agents/{email}/contacts/{address}` | `/v1/agents/{email}` | **yes — deletes the agent** |
| `/v1/contacts/{address}`, `/v1/domains/{domain}`, `/v1/templates/{id}`, `/v1/webhooks/{id}` | `/v1` | no |
| `/v1/contacts/imports/{batch_id}` | `/v1/contacts` | no |

The confirmation token does not help: `DeleteConfirm` (`internal/httpapi/delete_confirm.go`) is one shared type with `enum:"DELETE"` used by all 12 delete ops, so a token minted to authorize "remove one suppression" is byte-identical to one authorizing "delete this account". The caller's own safety check satisfies the escalated operation's guard.

## Why nothing is broken today

Two independent accidents, both verified against a local server:

1. **Values can't reach the list.** `handleCreateAgentSuppression` validates with `mail.ParseAddress`, so no stored address is all-dots. Same for contacts.
[...]

PR #909, opened 2026-08-19
fix(sdk): reject dot-segment path params that collapse a DELETE onto a bigger resource

AmirF194:
## Summary

Five DELETE routes take a path parameter (`address` / `id`) that, when it is
exactly `.` or `..`, collapses the built URL onto the resource *above* the
one being deleted. Neither `encodeURIComponent()` (TS) nor
`urllib.parse.quote()` (Python) escapes `.`, so the literal value reaches the
client's URL parser, which then removes the dot segment together with the
segment before it (verified: `new URL('/v1/account/suppressions/..', origin).pathname === '/v1/account/'`,
and `httpx.URL('.../account/suppressions/..').path == '/v1/account'`). Two
accidents currently keep this latent (address-format validation rejects an
all-dots email locally; the router has no trailing-slash redirect, so the
collapsed path 404s), but neither is asserted anywhere in the codebase.
Refs #792.

Enumerated every DELETE route in `api/openapi.yaml` programmatically (not by
reading the issue) and kept the ones whose 2-segment collapse lands on
another DELETE-capable route:

- `agents.deleteSuppression` (`/v1/agents/{email}/suppressions/{address}`) → `DELETE /v1/agents/{email}` (deletes the agent)
- `contacts.deleteOutreach` (`/v1/agents/{email}/contacts/{address}`) → same target
- `messages.delete` (`/v1/agents/{email}/messages/{id}`) → same target, **not named in the issue**
- `account.suppressions.delete` (`/v1/account/suppressions/{address}`) → `DELETE /v1/account` (deletes the account)
- `account.apiKeys.delete` (`/v1/account/api-keys/{id}`) → same target

Both SDKs now reject a value of exactly `.` or `..` for these five
parameters before a request is built (`E2AError` in TS / `E2AValidationError`
in Python, code `unsafe_path_segment`, no HTTP call made). The CLI and MCP
server both build on `@e2a/sdk`'s `E2AClient`, so this closes the same gap
there with no change to either package.

## Scope

This is priority-1 only, per the issue's own stated order: the client-side
guard in both SDKs. Left as a fast-follow:

- Router-level hardening (a `chi` trailing-slash redirect would turn the
  current silent 404 into a live collapse) plus a regression test pinning it
- The unguarded raw-`fetch` call sites in the web dashboard
  (`contacts/page.tsx`, `templates/page.tsx`, `webhooks/page.tsx`,
  `components/onboarding/api.ts`), which do not go through either SDK
- The audit script the issue proposes

## Client surface checklist

- [x] TypeScript SDK: `assertNotDotSegment` in `sdks/typescript/src/v1/client.ts`, called at the five delete sites
- [x] Python SDK: `_assert_not_dot_segment` in `sdks/python/src/e2a/v1/client.py`, same five sites
- [x] Tests at each surface (reject-and-no-request-sent + one ordinary-value pass-through)
- Go handler / OpenAPI spec / CLI / MCP: not touched; no wire-format change, and CLI/MCP inherit the fix through `@e2a/sdk`

## Test plan

- [x] `npm run test:coverage --workspace @e2a/sdk` (typecheck, 251 unit tests, type tests) in Docker on Node 22; new lines confirmed hit in the lcov report
- [x] `pytest sdks/python/tests/` in Docker on Python 3.9 and 3.13 (568 passed on both) plus `mypy` clean; new lines confirmed hit in the coverage report
- [x] Both new test files fail against pre-fix `client.ts` / `client.py` (the reject assertions resolve instead of throwing) and pass against this branch, run side by side in separate containers
- [x] Not verified: the router 404 behavior itself and the web-dashboard call sites named above; both are out of scope per "Scope"

### DIFF
--- a/sdks/python/CHANGELOG.md
+++ b/sdks/python/CHANGELOG.md
@@ -5,18 +5,21 @@
 ### Changed
 - **Dot-segment path parameters (``.`` / ``..``) are now rejected
   client-side** on ``agents.delete_suppression``, ``contacts.delete_outreach``
-  (both ``email`` and ``address``), ``messages.delete``, ``messages.restore``,
-  ``messages.update_labels``, ``account.suppressions.delete``, and
-  ``account.api_keys.delete``, raising ``E2AValidationError`` (code
-  ``unsafe_path_segment``) before any request is built. This was LIVE on
-  Python, not merely latent: ``httpx`` drops the collapsed path's trailing
-  slash (unlike the TS client, which keeps it unless the input string itself
-  ends in ``/``), so several of these values reached a real, matching route
-  rather than 404ing, e.g. ``account.suppressions.delete("..")`` reached
-  ``DELETE /v1/account`` directly. This is a behavior change for any caller
-  that was, deliberately or not, passing one of these values: it now raises
-  instead of sending the (misdirected) request. Available on both
-  ``AsyncE2AClient`` and the synchronous ``E2AClient``.
+  (both ``email`` and ``address``), ``contacts.delete_import``,
+  ``contacts.set_outreach`` (``email``), ``messages.delete``,
+  ``messages.restore``, ``messages.update_labels``,
+  ``account.suppressions.delete``, and ``account.api_keys.delete``, raising
+  ``E2AValidationError`` (code ``unsafe_path_segment``) before any request is
+  built. This was LIVE, not merely latent: ``httpx`` normalizes the dot
+  segments away and drops the collapsed path's trailing slash (the TS
+  client's URL builder does the same for every mid-path collapse, and for a
+  collapsed trailing slash whenever no query string follows), so several of
+  these values reached a real, matching route rather than 404ing, e.g.
+  ``account.suppressions.delete("..")`` reached ``DELETE /v1/account``
+  directly. This is a behavior change for any caller that was, deliberately
+  or not, passing one of these values: it now raises instead of sending the
+  (misdirected) request. Available on both ``AsyncE2AClient`` and the
+  synchronous ``E2AClient``.
 
 ## 5.7.0
 
--- a/sdks/python/CHANGELOG.md
+++ b/sdks/python/CHANGELOG.md
@@ -12,7 +12,7 @@
   Python, not merely latent: ``httpx`` drops the collapsed path's trailing
   slash (unlike the TS client, which keeps it unless the input string itself
   ends in ``/``), so several of these values reached a real, matching route
-  rather than 404ing — e.g. ``account.suppressions.delete("..")`` reached
+  rather than 404ing, e.g. ``account.suppressions.delete("..")`` reached
   ``DELETE /v1/account`` directly. This is a behavior change for any caller
   that was, deliberately or not, passing one of these values: it now raises
   instead of sending the (misdirected) request. Available on both
--- a/sdks/python/CHANGELOG.md
+++ b/sdks/python/CHANGELOG.md
@@ -1,5 +1,23 @@
 # Changelog
 
+## Unreleased
+
+### Changed
+- **Dot-segment path parameters (``.`` / ``..``) are now rejected
+  client-side** on ``agents.delete_suppression``, ``contacts.delete_outreach``
+  (both ``email`` and ``address``), ``messages.delete``, ``messages.restore``,
+  ``messages.update_labels``, ``account.suppressions.delete``, and
+  ``account.api_keys.delete``, raising ``E2AValidationError`` (code
+  ``unsafe_path_segment``) before any request is built. This was LIVE on
+  Python, not merely latent: ``httpx`` drops the collapsed path's trailing
+  slash (unlike the TS client, which keeps it unless the input string itself
+  ends in ``/``), so several of these values reached a real, matching route
+  rather than 404ing — e.g. ``account.suppressions.delete("..")`` reached
+  ``DELETE /v1/account`` directly. This is a behavior change for any caller
+  that was, deliberately or not, passing one of these values: it now raises
+  instead of sending the (misdirected) request. Available on both
+  ``AsyncE2AClient`` and the synchronous ``E2AClient``.
+
 ## 5.7.0
 
 Additive only. Every 5.6.0 call site keeps working identically. Available on
--- a/sdks/python/src/e2a/v1/client.py
+++ b/sdks/python/src/e2a/v1/client.py
@@ -210,6 +210,20 @@ def _coerce(model_cls: Type[T], body: Optional[Body]) -> T:
         ) from e
 
 
+def _assert_not_dot_segment(value: str, param: str) -> str:
+    # A value of exactly "." or ".." collapses the built URL onto the
+    # PRECEDING path segment (urllib.parse.quote leaves "." unescaped),
+    # retargeting a DELETE at a bigger resource. Reject before the wire.
+    if value in (".", ".."):
+        raise E2AValidationError(
+            code="unsafe_path_segment",
+            message=f'{param} must not be "." or ".."; it would collapse the request path onto a different, larger resource',
+            status=0,
+            retryable=False,
+        )
+    return value
+
+
 class _TypedApiClient(ApiClient):
     """Map malformed successful responses before the retry boundary sees them."""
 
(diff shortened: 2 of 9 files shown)


jiashuoz:
## Review

Real bug, right mechanism, tests pass as claimed (TS 251, Python 568 + `mypy` clean). Rejecting rather than encoding is correct — percent-encoding wouldn't help, since WHATWG URL treats `%2e`/`%2E` as dot segments too, and both SDKs already escape `/` (`encodeURIComponent`; Python `safe_chars_for_path_param = ''`). So exact `.`/`..` really is the complete dangerous set for these params: no `%2e%2e` or `a/../b` bypass. No false-positive risk either — ids are `{type}_{random}`, addresses are emails. Generated `*Api` classes aren't exported from `src/v1/index.ts`, so the ergonomic layer is the only public path, and the CLI/MCP inheritance claim checks out.

Findings below were verified by building the branch and probing both SDKs with a recording transport.

### Blockers

**1. Two live collapse positions are still unguarded.** Both are on methods this PR already touches, and both produce **no trailing slash**, so neither of the two "accidents" cited in the description applies. Measured on this branch, both SDKs:

```
c.contacts.deleteOutreach("..", "alice@example.test")
  -> DELETE /v1/contacts/alice%40example.test?confirm=DELETE

c.messages.restore("bot@example.test", "..")
  -> POST /v1/agents/bot%40example.test/restore
```

The first hits `deleteContact`, which requires exactly `?confirm=DELETE` — already supplied — so it destroys the contact record, directly contradicting the method's own docstring ("The contact itself survives … this is not consent"). The PR guards `address` on this call but not `email`. The second hits `restoreAgent`: neither op takes a body, so it is accepted, returns an `AgentView` typed as `MessageView`, and un-does a deliberate agent deletion.

Worth noting for threat model: in the MCP server both arguments come straight from LLM-supplied tool args, and `resolveAddress` does no format validation.

Counting only state-changing positions whose collapse lands on a live same-method route, this guards 5 of 7.

**2. The "latent, not live" framing holds for TypeScript but not for Python.** Measured pre-fix:

|---|---|---|
| `account.apiKeys.delete("..")` | `/v1/account/` — 404s | `/v1/account` — **live** |
| `agents.deleteSuppression(email, "..")` | `/v1/agents/{email}/` — 404s | `/v1/agents/{email}` — **live** |

`httpx` drops the collapsed trailing slash outright, so the chi trailing-slash 404 never protected the Python SDK — `account.suppressions.delete("..")` was a live account deletion. (`messages.delete(email, "..", permanent=True)` would have been a *hard* delete of the agent, since `permanent` rides along on the collapsed URL.)

On the TS side the slash survives only because `RequestContext.getUrl()` strips a trailing slash just when the whole URL string ends in `/`, and `?confirm=DELETE` prevents that. Requests with no query string do get it stripped: `messages.get(email, "..")` → `GET /v1/agents/{email}`, live.

This decides whether the fix needs an advisory, a patch release of both SDKs, and a backport, so it's worth getting into the description before merge.

### Requests

**3. A denominator test rather than more point tests.** Enumerate the path parameters (from `api/openapi.yaml`, or from the ergonomic methods) and assert each rejects `..` with no request sent, with an explicit allowlist for deliberate exclusions — same shape as the existing `contract-coverage.test.ts` ratchet. Given finding 1, this is the mechanism that actually prevents the next resource method from reopening the hole.

**4. TS should throw `E2AValidationError`, not bare `E2AError`** (`errors.ts:65` — "input validation"). It subclasses `E2AError`, so the change is strictly compatible, and it closes the parity gap with the Python side.

**5. Python tests should assert no request was sent** — `assert httpx_mock.get_requests() == []`, matching what the TS tests already assert. It's currently only implied by `pytest-httpx` erroring on an unmatched request.

**6. The explanatory comment is wrong about `.`.** Both copies say a value of exactly `"."` or `".."` "collapses the built URL onto the PRECEDING path segment". Only `..` removes the preceding segment; `.` removes only itself, landing on the collection (`/v1/account/suppressions/.` → `/v1/account/suppressions/`). For the five guarded DELETEs no collection has a DELETE, so `.` is harmless defense-in-depth rather than a live path — worth saying so in a comment whose whole job is explaining why the check exists.

**7. CHANGELOG entries** for both SDKs — this is a behavior change for callers (a new throw).

### Out of scope, noted

The WS URL builders (`ws.ts`, `websocket.py`) interpolate the agent address by hand and won't be covered by any HTTP-layer guard, but they collapse to non-routes (there is no `/v1/ws`), so they fail closed — a one-line comment, not a fix.

The generic all-params guard, and the corrections it needs relative to #792's proposed plan, are in the follow-up issue.

AmirF194:
Thanks for the thorough review, both blockers were real and confirmed. Pushed a fix.

**Blocker 1 (two live unguarded call sites).** You were right on both:

- `contacts.deleteOutreach`: the `email` param had no guard. Reproduced against the pre-fix branch (Docker, mocked fetch/httpx): `deleteOutreach("..", address)` sent `DELETE /v1/contacts/{address}?confirm=DELETE`, exactly `deleteContact`, permanently deleting the contact the docstring says survives this call. Added `assertNotDotSegment`/`_assert_not_dot_segment` on `email` in both SDKs.
- `messages.restore`: no guard at all. Reproduced the same way: `restore(email, "..")` sent `POST /v1/agents/{email}/restore`, exactly `restoreAgent`. Guarded `id` in both SDKs.

While re-deriving the enumeration to catch these, I re-ran it against every HTTP method instead of just DELETE and found one more: `messages.updateLabels` is a PATCH, and collapsing `id` retargets it onto `PATCH /v1/agents/{email}` (`updateAgent`). I did not confirm this one live against a real server; `updateAgent`'s body schema (`additionalProperties: false`) would likely reject the labels fields, so it may just 422 rather than silently patch the agent. Guarded it anyway on the same principle as the WS builders you flagged: the client shouldn't send a state-changing request at the wrong resource regardless of what the server does with it.

**Blocker 2 (Python framing).** Confirmed with `httpx.URL()` in Docker: `httpx.URL(".../account/suppressions/..").path == "/v1/account"`, no trailing slash, an exact match for the real route. `new URL()` on the TS side keeps a trailing slash (`/v1/account/`), which is what kept it 404ing there. So on Python this was live before the guard existed, not merely latent the way the PR body originally framed it. Fixed the code comment on `_assert_not_dot_segment` and wrote the CHANGELOG entries around the live/latent distinction rather than the general "two accidents" framing.

**Requests picked up:**
- #4: TS now throws `E2AValidationError`, not the bare `E2AError`.
- #5: Python tests assert `httpx_mock.get_requests() == []` instead of relying on pytest-httpx's own unmatched-request behavior.
- #6: fixed the comment, only `..` collapses onto the preceding segment; `.` collapses onto its own parent collection.
- #7: added CHANGELOG entries to both SDKs under an Unreleased section.
- #3: added `sdks/python/tests/test_dot_segment_enumeration.py`. It computes the denominator straight from `api/openapi.yaml` (every path param whose `..` collapse lands on a different route sharing the same HTTP method), classifies each hit into an explicit GUARDED or ALLOWLIST map, and fails if a new one shows up unclassified. Current sweep: 15 collapses, 8 guarded (matches every DELETE/PATCH/POST-restore call site above), 7 allowlisted, all GET-only, read-only info disclosure rather than a destructive action, out of scope for this pass on the same reasoning as the WS builders. Didn't duplicate the same parser in TS; the TS test file pins the same 8 call sites directly and points at the Python file as the source of the enumeration.
- The out-of-scope WS builders: added a one-line comment to both `ws.ts` and `websocket.py` explaining why they're safe without a guard, no functional change there.

**Verified this session, in Docker on this VPS:**
- `npm run test:coverage --workspace @e2a/sdk` on `node:22-slim`: typecheck, 255 unit tests (was 251, +4 new), type tests, all pass.
- `pytest tests/ -q` + `mypy` on `python:3.9-slim` and `python:3.13-slim`: 607 passed, 48 skipped on both, mypy clean.
- `scripts/check-repository-text-integrity.sh`: passes.
- Both blockers reproduced against the pre-fix branch state before patching (temp test files, deleted after), and re-run after the fix to confirm they now raise `unsafe_path_segment` with no request sent.

Not independently verified: whether `updateAgent` actually rejects the labels body server-side (no live server available here), so I called that one out as unconfirmed above rather than claiming it.

jiashuoz (approved):
Reviewed with a deep multi-agent pass (10 finder angles + adversarial cross-verification). The two blocking gaps found — an unguarded `delete_import` dot-segment collapse and the enumeration gate's blindness to `"."` collapses (which also hid an unguarded `set_outreach` retarget) — are fixed in 18b21a3d2, along with two release-note corrections. Both SDK suites are green and the gate now proves all 10 mutating collapses in the current spec are guarded. Remaining findings are non-blocking follow-ups (TS-side gate parity, choke-point refactor, web raw-fetch inventory, WS builder comments, test hygiene) and will be tracked separately.

AmirF194:
Thanks for the deep review, both blockers you found were real. Appreciate the merge.

Issue #823, opened 2026-08-04
tests/contract: scenarios.yaml has no strict decode — a typo'd assertion key is silently dropped by all three runners


jiashuoz:
## What

`scenarios.yaml` is decoded permissively in every runner, so a misspelled key is not an error — it is a **silently missing assertion**, and the scenario passes with less coverage than it appears to have.

- Go: `yaml.Unmarshal` in `loadScenarios` (`tests/contract/contract_test.go`) ignores unknown fields.
- TypeScript: `execRequest` keys off `ex?.body_match`.
- Python: `"body_match" in ex`.

## Evidence

Demonstrated during a review of #819/#821. Misspelling the assertion **block key** — `body_matchs:` instead of `body_match:` — on a step of `account_limits_enforced` leaves `TestScenarios/account_limits_enforced` **passing** in all three runners, with nothing asserted about that response beyond its status.

(Misspelling a *path inside* `body_match`, e.g. `error.details.resourse`, is caught — the runner fails on a missing path. It is the block key itself that vanishes.)

This is the more dangerous direction: the scenario still runs, still passes, and reads as if it proves something.

## Scope

Every scenario in the file, not just the new ones. The three hand-written shape tests added in #819 pin only two steps of one scenario; the same typo anywhere else is invisible everywhere.

## Proposed fix

`yaml.Decoder` + `KnownFields(true)` in the Go loader. That makes the Go contract job the gate for the whole file in every language, since all three read the same YAML, and it is far less code than hand-written shape tests per scenario.

Worth expecting a first run to surface pre-existing typos — those are the point.

The same reasoning applies to unknown *setup* keys, which the TS and Python runners silently ignore (a setup primitive one runner does not implement currently degrades to "scenario runs without its preconditions" rather than an error). Strict decode addresses the shape; the runners' `if key in step` dispatch would still need a matching "unknown primitive" error.

PR #910, opened 2026-08-19
fix(contract): strict-decode scenarios.yaml so a typo'd key is a decode error

AmirF194:
## Summary

`loadScenarios` (`tests/contract/contract_test.go`) parses `scenarios.yaml` with plain `yaml.Unmarshal`, which silently drops any key with no matching struct field. A scenario author who misspells the assertion block key (`body_matchs:` instead of `body_match:`) gets a scenario that still runs and still reports PASS, having asserted nothing about the response body.

`decodeScenarios` now decodes with `yaml.NewDecoder(...).KnownFields(true)`, so an unknown key fails the decode with the field name in the error instead of vanishing. `loadScenarios` calls it and fails the test with that error, same as it already fails on a malformed YAML document.

Refs #823. This covers the issue's proposed fix (the Go loader) and, since all three runners read the same `scenarios.yaml`, the Go contract job now gates the file for TypeScript and Python too. It does not touch the runners' own unknown-*setup*-key handling that the issue separately flags in its last paragraph (`if key in step` dispatch in the TS/Python runners), which is a different code path than the assertion-key decode this PR fixes.

## Test plan

- New `TestDecodeScenariosRejectsUnknownFields` exercises `decodeScenarios` directly with a `body_matchs` typo and asserts the decode error names the field. It cannot compile against `origin/main` (`decodeScenarios` does not exist there); it passes on this branch.
- Full `TestScenarios` suite (32 scenarios) against a real Postgres 16 container: all pass, so the strict decode does not flag any pre-existing key in the real `scenarios.yaml`.
- `go vet -tags integration ./tests/contract/...` and `gofmt -l` clean on the changed file.
- Not verified: the TypeScript and Python contract runners' own setup-key handling (out of scope for this PR, see above).

### DIFF
--- a/tests/contract/contract_test.go
+++ b/tests/contract/contract_test.go
@@ -919,17 +919,55 @@ func extractEmailFromWSPath(path string) string {
 
 // ── Entry point ────────────────────────────────────────────────────
 
+// decodeScenarios parses scenario YAML with strict field checking, so an
+// unknown key (a typo) is a decode error instead of a silently dropped field.
+func decodeScenarios(data []byte) (scenarioFile, error) {
+	var sf scenarioFile
+	dec := yaml.NewDecoder(bytes.NewReader(data))
+	dec.KnownFields(true)
+	if err := dec.Decode(&sf); err != nil {
+		return scenarioFile{}, err
+	}
+	return sf, nil
+}
+
 func loadScenarios(t *testing.T) []scenario {
 	t.Helper()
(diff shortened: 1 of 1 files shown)


jiashuoz (approved):
Reviewed the strict YAML decode path and regression coverage. The refreshed branch is clean and all CI checks pass.

AmirF194:
Thanks for reviewing and merging.

PR #919, opened 2026-08-20
docs(design): SOC 2 / HIPAA compliance plan

hyperSuperCube:
## Summary

Docs-only — no code, API, or client-surface changes. (Client surface checklist removed per template guidance.)

Highlights reviewers may want to weigh in on:
- Treating e2a as a HIPAA **business associate** (persistent message storage rules out the conduit exception) and targeting the 2025 Security Rule NPRM bar (mandatory encryption + MFA).
- The Gemini AI Studio screening endpoint as the headline HIPAA blocker, with the proposed resolution options (Vertex/BAA-covered provider vs hard-off in a HIPAA profile).
- The proposed `audit_events` shape and the envelope-encryption migration strategy (dual-read + River re-encryption, no destructive DDL on `messages`).

## Operational risk

None from this PR itself (documentation only). The document does *propose* future changes with operational weight (field encryption migration, retention sweeps, production `sslmode` enforcement); each would land as its own PR train with tests and migration review.

## Test plan

- [x] `scripts/check-repository-text-integrity.sh` passes
- [x] `scripts/check-no-committed-credentials.sh` passes
- [x] Public-repo data boundary reviewed: doc references only in-tree code paths and already-public material, no customer data

PR #920, opened 2026-08-20
feat(piguard): BAA-coverable Vertex AI endpoint support for the Gemini detector

hyperSuperCube:
## Summary

Implements finding **G1** from the SOC 2 / HIPAA compliance plan (#919): the Gemini LLM-as-detector layer hardcoded the Google AI Studio consumer endpoint, which cannot sit under a Google Cloud BAA/DPA — a blocker for any deployment with confidentiality obligations, since the detector sends the message subject, sender, and up to 4,000 chars of body text.

This makes the endpoint and auth mode first-class configuration so the detector can run against **project-scoped Vertex AI under the operator's Google Cloud agreement**:

- `GEMINI_BASE_URL` (or `GeminiConfig.BaseURL`) — the appended `/models/{model}:generateContent` path fits both the AI Studio default and Vertex bases ending in `.../publishers/google`.
- `GEMINI_AUTH=adc` (or `GeminiConfig.Auth`) — OAuth bearer via Application Default Credentials, which Vertex requires; `api_key` stays the default and the AI Studio behavior is byte-for-byte unchanged. An unknown mode fails construction (no silent fallback to the non-contracted endpoint).
- Bearer-token fetch failures are classified transient (retried within the existing 5s/10s budget, then `StatusError` so heuristics carries) and never propagate token or token-source error text.
- `inboundscreen.BuildEngine` now logs when Gemini is configured but the detector can't be constructed, instead of silently degrading to heuristics-only.
- Deployment docs + `config.example.yaml` now state exactly what content leaves the system and show the compliant Vertex configuration.

No new module dependencies: `golang.org/x/oauth2` was already a direct dependency and `go mod tidy` is a no-op.

## Client surface checklist

Not applicable — env/config-only change to the screening engine; no `/v1` handler, schema, migration, or SDK/CLI/MCP surface is touched (verified: `make test-unit` green, no OpenAPI diff).

## Operational risk

- Default behavior (API-key mode against AI Studio) is unchanged; existing deployments need no action.
- `GEMINI_AUTH=adc` without resolvable ADC fails detector construction at startup — screening continues heuristics-only and the condition is now logged loudly. Rollback = unset the two new env vars.
- Fail-open semantics on scan errors are unchanged.

## Test plan

- [x] New unit tests: bearer header attached / api-key header absent in `adc` mode (and vice versa), `GEMINI_BASE_URL` env routing, unknown `GEMINI_AUTH` rejected, token-fetch failure transient + non-leaking
- [x] `go test ./internal/piguard/ ./internal/inboundscreen/ ./internal/logredact/` green
- [x] `make test-unit`, `go build ./...`, `go vet`, `gofmt` clean
- [x] Repo text-integrity and committed-credentials gates pass
- [ ] Staging: point at a Vertex `publishers/google` base with workload identity and confirm a scanned message round-trips


internal/inboundscreen/inboundscreen.go
```diff
@@ -232,6 +232,13 @@ func BuildEngine() *piguard.Engine {
 			detectors = append(detectors, d)
 			cfg.Timeout = GeminiDetectorTimeout
 			log.Printf("[piguard] Gemini detector enabled (model: %s)", d.Model())
+		} else if os.Getenv("GEMINI_API_KEY") != "" || os.Getenv("GOOGLE_API_KEY") != "" || os.Getenv("GEMINI_AUTH") != "" {
```
AmirF194, line 235: Ran a test against `f54f3a97` in a clean container that sets only `GEMINI_BASE_URL` (no API key, no `GEMINI_AUTH`) and captures `BuildEngine`'s log output.

The new else-if at line 235 checks `GEMINI_API_KEY`, `GOOGLE_API_KEY`, and `GEMINI_AUTH`, but not `GEMINI_BASE_URL`. An operator who sets only `GEMINI_BASE_URL` while pointing at Vertex, and forgets `GEMINI_AUTH=adc`, gets `NewGeminiDetector` failing on the default api_key path with no API key set. All three checked env vars are empty, so the else-if is skipped and `BuildEngine` falls through to heuristics-only with zero log output, exactly the silent degradation this PR's own comment says should never happen.

```
$ go test ./internal/inboundscreen/ -run TestBuildEngine_BaseURLOnlySilentlyDegrades -v
=== RUN   TestBuildEngine_BaseURLOnlySilentlyDegrades
    zz_repro_test.go:26: captured log output: ""
    zz_repro_test.go:28: expected an operator-visible log line when GEMINI_BASE_URL is set but construction fails; got: ""
--- FAIL: TestBuildEngine_BaseURLOnlySilentlyDegrades (0.00s)
```

Fix: add `|| os.Getenv("GEMINI_BASE_URL") != ""` to the else-if condition.

Issue #771, opened 2026-07-31
outbound: scale rate-limit deferral snooze by elapsed deferral time (re-drive cost is quadratic in backlog)


jiashuoz:
Follow-up from PR #770 review (fire-time per-agent rate limit, internal/sendrate).

**Observation** (cost curve, not a correctness defect): each rate-limit deferral costs ClaimSend + Reserve + ReleaseSend — 3 transactions, 2 of them UPDATEs on messages — and every still-deferred job repeats that each window. Draining a backlog of N costs ~N²/(2·limit) re-drives. A realistic burst (60 accepts/min for an hour, all scheduled for the same instant → N ≈ 3,600) is ~108k wake-ups and ~216k messages row updates over the drain; the 72h-horizon ceiling (259,200/agent) works out to ~560M. Those wake-ups also occupy outbound queue workers, so one agent's backlog adds latency to other tenants' sends.

The per-message jitter added in #770 fixes lockstep but not volume.

**Proposal**: the worker already holds AcceptedAt/ScheduledAt, so scale the deferral snooze by how long the message has already been deferring (e.g. exponential backoff capped well under the 72h horizon — minutes, not hours), making re-drives logarithmic. No schema change. Alternatively track a per-job deferral count.

Context: PR #770 (worker.go clampRateSnooze/rateJitter, internal/sendrate).

PR #921, opened 2026-08-20
fix(outbound): back off rate-limit deferral by elapsed wait (#771)

cyj-git-0825:
## Problem

Follow-up from PR #770 review. The fire-time per-agent rate limiter
(`internal/sendrate`) defers an over-budget send with a snooze bounded at one
window, so a deferred job re-drives ~every window until it wins a slot. Draining
a backlog of N costs ~N²/(2·limit) re-drives, each ClaimSend + Reserve +
ReleaseSend (3 txs, 2 of them `messages` UPDATEs) and each occupying a send
worker — so one agent's backlog also adds latency to other tenants' sends. The
per-message jitter added in #770 fixes lockstep but not volume.

## Fix

Scale the deferral snooze by how long the message has **already** been waiting to
fire (elapsed since `max(accept, scheduled)`, which the worker already loads).
Sleeping at least "as long as we've already waited" doubles the total wait on
each pass, so re-drives become **logarithmic** instead of quadratic. Bounded by
`rateMaxSnooze` so a job still drains well under the 72h retry horizon.

No schema change; `pastRetryHorizon` (which runs before the snooze) is refactored
to share the same `fireTime()` helper with identical behavior.

## The cap value

`rateMaxSnooze = 3m` (3× the 1m production window). The exponential doubling
already does the heavy lifting; the cap only bounds the deep tail. 3m keeps
essentially all of the wake-up savings while bounding worst-case re-check latency
and the fairness skew. It's a tunable knob — raise it to trade lower wake-up
volume for higher tail latency.

## Fairness trade-off (documented in `rateSnooze`)

Because a longer-waiting message sleeps longer, a draining backlog skews
newer-first (LIFO-ish) — an old message can cede a freed slot to a fresher one.
This is inherent to any elapsed/attempt-scaled backoff and is bounded: the cap
keeps an old message re-driving on a fixed cadence (hundreds of retries before
72h), and the pre-existing `pastRetryHorizon` still terminates a genuinely stuck
job.

## Jitter

`rateJitter` now scales its spread to the actual snooze instead of the fixed
window, so a capped, minutes-long backlog still fans out proportionally rather
than re-herding in a window-sized band — preserving #770's anti-herd property
that this change would otherwise erode.

## Scope

Pure send-worker behavior: **no migration, no `/v1` API / OpenAPI / SDK / CLI /
MCP / dashboard change.**

## Testing

- White-box (`rate_snooze_internal_test.go`): doubling, cap, zero-fireTime
  fallback, floor, `fireTime` max/zero, and jitter scaling with spread.
- Black-box (`rate_test.go`): a job with an old fire time snoozes beyond the
  window through the real `Work()` deferral path; existing deferral/clamp/jitter
  tests unchanged (their zero-timestamp jobs keep the base wait).
- `go vet` + `go test -race ./internal/outboundsend` green.


internal/outboundsend/worker.go
```diff
@@ -624,21 +645,60 @@ func clampRateSnooze(d, window time.Duration) time.Duration {
...
+// skews toward newer-first (LIFO-ish) — an old message can cede a freed slot to
+// a fresher one. This is inherent to any elapsed/attempt-scaled backoff and is
+// bounded: the rateMaxSnooze cap keeps an old message re-driving on a fixed
+// cadence (hundreds of retries before the 72h horizon), and pastRetryHorizon —
+// which runs BEFORE this in the deferral branch — still terminates a genuinely
+// stuck job rather than letting it churn forever.
+//
+// A zero fireTime (unknown timestamps) disables the backoff and falls back to
+// the base wait, mirroring pastRetryHorizon's zero-handling.
+func rateSnooze(fireTime time.Time, retryDelay, window time.Duration) time.Duration {
```
AmirF194: I read this against 4763015b (worker.go, rate_test.go).

fireTime() at worker.go:141 is max(AcceptedAt, ScheduledAt), and AcceptedAt is messages.created_at, not the time of a message's first rate-limit defer. rateSnooze (worker.go:671) uses that same fireTime as its elapsed-since clock.

So a message that sat a while before its first send attempt for a reason unrelated to rate limiting, most concretely one held for HITL review and then approved (ReviewedAt is deliberately excluded from fireTime, per the comment at line 122), can hit the rateMaxSnooze cap on the very first rate-limit defer it ever gets, since elapsed is already large relative to the base window clamp. The PR body's fairness section describes the older-waits-longer trade-off as a property of a backlog draining through repeated defers, and this is a different case: a message with zero prior defers already at the cap.

Is that intended, or should the backoff anchor to the time of the message's own first rate-limit defer rather than fireTime?

jiashuoz (dismissed):
Approved. The backoff is a worthwhile reduction in repeated limiter work, and the targeted worker coverage is solid.

One non-blocking follow-up: fireTime uses max(AcceptedAt, ScheduledAt) but excludes ReviewedAt. A message held for human review can therefore reach its first rate-limit deferral with an already-old clock and immediately receive the 3-minute cap, even though it only just became eligible to submit. Please consider anchoring the rate backoff to submission eligibility (including ReviewedAt) or to the first rate deferral, with a held-message regression test. This affects avoidable latency rather than delivery correctness, so I am not blocking the PR on it.

cyj-git-0825:
@AmirF194 — thanks for catching this. You're right that using `fireTime` could make a held message hit the cap on its first rate-limit defer, since the elapsed clock doesn't account for time spent waiting for HITL review.

I changed the backoff anchor to `submissionAnchor = max(acceptedAt, scheduledAt, reviewedAt)`, so the clock starts when the message actually becomes eligible to submit rather than from the older `fireTime`.

@jiashuoz — thanks for laying out the two solutions, that was inspiring and instructive. I went with the submission-eligibility anchor since it already matches the clock used for the latency SLI and avoids adding another persisted timestamp.

The trade-off is that it may include some time between becoming eligible and the first rate-limited attempt. I think that's a reasonable trade-off for now, but I'm happy to switch to a first-deferral anchor if you think that's the better model.

Also added `TestSendWorker_RateLimitedHeldMessageAnchorsAtReview` to cover the held-message case and verify that the first deferral doesn't immediately hit the 3m cap.

internal/outboundsend/worker.go
AmirF194: Anchoring at submissionAnchor instead of fireTime is the right fix, and it lines up with the latency SLI clock already in use. The held-message test covers the case I was asking about.

internal/outboundsend/worker.go
```diff
@@ -109,9 +118,11 @@ type SendJob struct {
 	ScheduledAt time.Time
 	// ReviewedAt is messages.reviewed_at — when a HITL hold was resolved into the
 	// send pipeline (human approve or TTL auto-approve), zero for a message that
-	// was never held. Consumed ONLY by submissionAnchor for the latency SLI; the
-	// retry horizon deliberately still measures from AcceptedAt, so the F2
-	// limitation in docs/design/hitl-ttl-async-send.md is unchanged by this field.
+	// was never held. Consumed via submissionAnchor — the latency SLI baseline and
+	// the rate-deferral backoff clock — so a reviewer's dwell time counts as
+	// neither error budget nor "already waiting to send". The retry horizon
+	// deliberately still measures from AcceptedAt, so the F2 limitation in
```
AmirF194, line 124: I read this against 84b1f918 (the PR head).

The comment on `ReviewedAt` at worker.go:124 says "The retry horizon deliberately still measures from AcceptedAt", but `pastRetryHorizon()` (worker.go:156-162) measures from `fireTime()` (worker.go:145-151), which is `max(AcceptedAt, ScheduledAt)`, not `AcceptedAt` alone. For a scheduled send where `ScheduledAt` is later than `AcceptedAt`, the retry horizon starts at `ScheduledAt`, contradicting this line.

This predates the PR (the pre-PR inline code at the same spot already computed `max(accept, scheduled)`), but the PR rewrites the rest of this exact comment block (the `ReviewedAt` paragraph, worker.go:119-126) to describe the new `submissionAnchor` wiring, and carries the inaccurate claim forward unchanged.

Suggested fix, same line:

```
// deliberately still measures from fireTime (max of accept and scheduled), not
// from submissionAnchor, so the F2 limitation in
```

PR #924, opened 2026-08-21
feat(cli): add --cc and --bcc to e2a send and e2a reply

### LINKED ISSUE #816: feat(cli): e2a send has no --cc or --bcc flag
## Problem

`e2a send` cannot set Cc or Bcc. Every other client surface can.

| Surface | `cc` / `bcc` on send |
|---|---|
| REST / OpenAPI | yes |
| TypeScript SDK | yes |
| Python SDK | yes |
| MCP `send_message` | yes |
| **CLI `e2a send`** | **no** |

Current flags (`e2a send --help`, v2.1.0):

```
--to <email>               Recipient (repeatable)
--subject <s>              Subject line
--body <text>              Plain-text body (or --body-file <f>)
--html-file <f>            HTML body; text fallback derived if no --body
--attach <file>            Attach a file
--conversation-id <id>     Thread id
--reply-to <email>         Reply-To header
--idempotency-key <k>      Stable key so a retried invocation can't double-send
--agent <email>            Sending inbox
--json                     Print the full send result as JSON
```

`--reply-to` is there; `--cc` and `--bcc` are not. Nothing in the help text says Cc is unsupported, so the omission reads as an oversight rather than a decision.

## Impact

The CLI is documented as the scriptable surface — "Scriptable primitives for agent harnesses (send/reply/messages/whoami, with a stable exit-code contract)". Any script that needs to copy a supervisor, an audit mailbox, or a shared inbox on outbound mail cannot use it, and has to drop to raw REST with a hand-managed API key. That is a meaningful step down in ergonomics for exactly the automation use case the CLI exists to serve.

Concretely: a batch-send script that had to Cc a fixed address on every message could not use `e2a send` at all, despite the CLI being otherwise a perfect fit — it already has `--idempotency-key`, `--reply-to`, `--body-file`, `--html-file`, and a stable exit-code contract.

## Proposed change

Add `--cc <email>` and `--bcc <email>`, both repeatable, matching `--to`'s existing shape and mapping to the `cc` / `bcc` request fields.

The API caps recipients at 50 combined across `to` + `cc` + `bcc`; the CLI should surface an over-cap rejection as the existing usage error (exit 2) rather than a request error where practical.

Also applies to `e2a reply` and `e2a forward` if they accept recipient overrides — worth checking the same gap there.

## Surfaces

- [ ] `cli/src` — flag parsing, help text, request mapping
- [ ] CLI tests (vitest) covering repeated `--cc`, `--bcc`, and combined-cap behavior
- [ ] `cli/README.md` flag reference
- [ ] Check `reply` / `forward` for the same omission

No API, SDK, or MCP change needed — the contract already carries `cc` and `bcc` on send.

## Note

Exit codes are a frozen contract (`cli/src/exit.ts`); this change should not need a new one. Reusing exit 2 for a recipient-count usage error keeps it inside the existing contract.

AmirF194:
## Summary

`e2a send`/`e2a reply` had no `--cc`/`--bcc`, even though the REST API, both
SDKs, and the MCP `send_message` tool already carry `cc`/`bcc` on send. Adds
the flags to both commands, wired straight onto the existing request fields.

## Client surface checklist

- [x] OpenAPI spec + generated types refreshed (`make generate` is clean): not touched, `cc`/`bcc` were already on `SendEmailRequest`/`ReplyRequest`
- [x] TypeScript SDK: not touched, for the same reason
- [x] Python SDK: not touched, for the same reason
- [x] CLI command or flag in `cli/src/commands/` + wired into `cli/src/bin/e2a.ts`
- [x] MCP tool: not touched, `send_message`/`reply_message` in `mcp/src/tools/legacy.ts` already accept `cc`/`bcc`
- [x] Tests at each surface above (positive + at least one negative-path / regression case)

**Skipped rows:** Go handler, migration. No server-side change, the fields already exist.

## Operational risk

None. Client-only change, no new server behavior.

## Test plan

- `npm run test:coverage --workspace @e2a/cli`: 321/321 passed (5 new). The 5 new
  cases fail on unmodified `main` (`tsc` rejects `cc`/`bcc` as unknown `SendOptions`/
  `ReplyOptions` properties) and pass on this branch.
- `npm run build --workspace @e2a/cli`: clean.
- Manual: built CLI, ran `e2a send` with 51 combined `--to`/`--cc`/`--bcc` recipients,
  confirmed it exits 2 (usage) before any request is sent, per the issue's ask.
- Not checked: a live end-to-end send through a running server (this environment
  has no e2a instance to send through); the SDK call shape is covered by the unit
  tests above instead.

The over-cap check only applies to `send`; `reply`'s primary recipients come
from the thread being replied to, so the CLI cannot compute the true combined
total the way it can for `send`.

`e2a forward` does not exist as a CLI command, so the issue's suggestion to
check it there does not apply.

Fixes #816

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -208,7 +208,9 @@ e2a reply msg_abc123 --body "On it." --agent bot@acme.com
 
 Common `send`/`reply` flags: `--body` / `--body-file`, `--html-file` (text
 fallback derived if no `--body`), `--attach` (repeatable; max 10 files, 10 MB
-each, 25 MB total), `--reply-to` (repeatable; max 5 addresses), `--send-at`
+each, 25 MB total), `--reply-to` (repeatable; max 5 addresses), `--cc` /
+`--bcc` (each repeatable; `--to` + `--cc` + `--bcc` combined are capped at
+50 recipients on `send`), `--send-at`
 (RFC 3339 with an explicit UTC offset, at most 90 days ahead),
 `--idempotency-key`, `--agent`, `--json`
 (print the full send result). `send`-only: `--to` (repeatable), `--subject`,
--- a/cli/src/__tests__/send.test.ts
+++ b/cli/src/__tests__/send.test.ts
@@ -81,6 +81,51 @@ describe("send/reply commands", () => {
     expect(mockSend.mock.calls[0][1].replyTo).toBeUndefined();
   });
 
+  it("passes repeated --cc and --bcc through as arrays", async () => {
+    mockSend.mockResolvedValue({ messageId: "msg_cc", status: "sent" });
+    const { send } = await import("../commands/send.js");
+    await send({
+      to: ["you@example.com"],
+      subject: "hi",
+      body: "hello",
+      cc: ["cc1@example.com", "cc2@example.com"],
+      bcc: ["audit@example.com"],
+    });
+    const call = mockSend.mock.calls[0][1];
+    expect(call.cc).toEqual(["cc1@example.com", "cc2@example.com"]);
+    expect(call.bcc).toEqual(["audit@example.com"]);
+  });
+
+  it("omits cc and bcc when their lists are empty", async () => {
+    mockSend.mockResolvedValue({ messageId: "msg_nocc", status: "sent" });
+    const { send } = await import("../commands/send.js");
+    await send({ to: ["you@example.com"], subject: "hi", body: "hello", cc: [], bcc: [] });
+    const call = mockSend.mock.calls[0][1];
+    expect(call.cc).toBeUndefined();
+    expect(call.bcc).toBeUndefined();
+  });
+
+  it("accepts exactly 50 recipients combined across --to, --cc, and --bcc", async () => {
+    mockSend.mockResolvedValue({ messageId: "msg_50", status: "sent" });
+    const { send } = await import("../commands/send.js");
+    const cc = Array.from({ length: 24 }, (_, i) => `cc${i}@example.com`);
+    const bcc = Array.from({ length: 25 }, (_, i) => `bcc${i}@example.com`);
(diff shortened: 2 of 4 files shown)


jiashuoz (approved):
LGTM — approving.

Verified against the code before approving:
- Generated `SendEmailRequest`/`ReplyRequest` already carry optional `cc`/`bcc`, and `SendEmailInput`/`ReplyInput` inherit them via `Omit`, so the pass-through is type-valid with no SDK changes.
- The server cap is 50 combined across to+cc+bcc (`internal/httpapi/outbound.go`, `maxRecipients = 50`, `too_many_recipients`), so the CLI pre-check matches server semantics and fails with exit 2 before any request — exactly what #816 asked for.
- Skipping the pre-check on `reply` is correctly reasoned (thread recipients aren't knowable client-side; server stays the backstop).
- 5 new tests cover passthrough, empty-list omission, the 50/51 boundary, and reply — good negative-path and boundary coverage.
- No CHANGELOG entry needed per-PR; entries are batched at release time per repo convention.

Two nits, non-blocking:

1. **Help-text placement** (`cli/src/bin/e2a.ts`): the cap note is a continuation line under `--json`, so it reads as part of the `--json` description. Consider attaching it to the `--bcc` line or making it a standalone note under the `send` block.

2. **Duplicated constant**: `MAX_COMBINED_RECIPIENTS = 50` mirrors the server's `maxRecipients`. If the server value ever changes, the CLI message drifts (harmlessly — the server error carries `max_recipients` in details). A one-line comment pointing at the server source would help future maintainers.

AmirF194:
Thanks for the quick review and merge. Glad the SendEmailInput/ReplyInput types lined up cleanly with the CLI wiring.

PR #929, opened 2026-08-22
fix(sdk): guard the generated-layer chokepoint against dot-segment path params

### LINKED ISSUE #915: Rewrite #792's client-side dot-segment guard plan: the proposed middleware hook can't see the dots
Follow-up to #792. While reviewing #909 I verified the client-side guard plan in #792 against the actual generated code. **Priority 1 as written cannot work**, and priority 2 does not cover the whole bug class. Filing the corrected plan so nobody implements the original.

## Priority 1 as written can't fire

#792 proposes:

> Every request passes through the hand-written `pre(context)` hook in `sdks/typescript/src/v1/generated/middleware.ts`'s middleware chain **after the path is built**. Reject any outgoing request whose path contains a `.`/`..` segment.

The dots are already gone by then. `RequestContext`'s constructor normalizes eagerly:

```ts
// sdks/typescript/src/v1/generated/http/http.ts:60
public constructor(url: string, private httpMethod: HttpMethod) {
    this.url = new URL(ensureAbsoluteUrl(url));
}
```

That runs inside the generated `*Api.ts` method, at construction — before `Middleware.pre()`, and before `RetryHttpLibrary.send()` (which was my first suggestion on #909, and is wrong for the same reason). Both hooks receive a `RequestContext` whose URL is already collapsed, so a dot-segment check there never fires.

## Corrected plan

**TypeScript — codegen post-processing.** The only place that still sees the raw value is inside the generated code, so the guard has to be injected by the pipeline. There's already a precedent: `sdks/typescript/scripts/generate-oag.sh:50` runs `scripts/guard-optional-header-params.py` over `$OUT/apis`. Either extend that pass to emit a dot-segment check per path parameter, or add a second pass over `generated/http/http.ts` that guards `RequestContext`'s constructor before the `new URL(...)`. Hand-editing `http.ts` is not an option — it isn't in `.openapi-generator-ignore`, so `make generate` overwrites it.

Whichever pass lands, it needs a static audit pinning the emitted shape per PR, for the same reason the optional-header-param guards have one: the freshness gate alone would re-bless whatever a new generator version emits.

**Python — `param_serialize`.** This one is straightforward. `ApiClient.param_serialize` (`sdks/python/src/e2a/v1/generated/api_client.py:145`) returns the URL as a plain string with path params substituted and dots intact; `httpx` normalizes later, in the rest client. `_TypedApiClient` (`sdks/python/src/e2a/v1/client.py:213`) already subclasses `ApiClient`, so overriding `param_serialize` there is a clean single chokepoint. `sync_client.py` delegates to `AsyncE2AClient`, so one override covers both clients.

**Matching rule, both languages.** Split the *path* on `/` and reject a segment equal to `.` or `..`. Not a substring match — emails and ids are full of dots — and the query string must be excluded.

## Priority 2 needs rewriting too

#792 proposes pinning that the trailing-slash form of every destructive route 404s. That is worth having, but it does not cover the bug class, because not every collapse produces a trailing slash:

- `DELETE /v1/agents/{email}/contacts/{address}` with `email=".."` → `DELETE /v1/contacts/{address}?confirm=DELETE`
- `POST /v1/agents/{email}/messages/{id}/restore` with `id=".."` → `POST /v1/agents/{email}/restore`

Both land squarely on live, legitimately routable endpoints with a well-formed `confirm` token. The server cannot distinguish them from a genuine request — there is nothing to reject. A router-level test can't close this, which is what makes the client-side guard load-bearing rather than defense-in-depth.

`#792`'s table only enumerated collapses of the **last** path parameter of each DELETE route, which is why these two aren't in it. The corrected enumeration should cover every path parameter position and every method.

## Also: the trailing-slash defense was never load-bearing for Python

#792 calls chi's no-redirect default the load-bearing defense. Measured pre-#909 against both SDKs:

|---|---|---|

`httpx` drops the collapsed trailing slash outright. On the TS side the slash survives only because `RequestContext.getUrl()` strips a trailing slash just when the whole URL string ends in `/`, which `?confirm=DELETE` prevents — TS requests without a query string do get it stripped and are live (`messages.get(email, "..")` → `GET /v1/agents/{email}`).

So the exposure was live for Python SDK callers, not latent. Priority 4 (scaling confirmation to blast radius, so a misrouted request carrying `confirm=DELETE` is rejected on arrival) remains the only option that fails closed regardless of client correctness, and is worth revisiting on that basis.

## Scope

#909 lands per-call-site guards in the hand-written ergonomic layer of both SDKs, covering the five highest-consequence DELETE positions (plus two more requested in review). That is not erased by `make generate`, so it stands on its own. This issue covers the generic all-parameters guard that #792's priority 1 was meant to be, plus the corrected priority 2.

AmirF194:
## Summary

#915 (filed during review of #909) shows that the per-call-site dot-segment
guard in the ergonomic wrapper classes (`client.agents`, `client.contacts`, ...)
never runs if a caller instead uses a generated `*Api` class directly against
the SDK's own client. This PR adds the guard at the one place every generated
method actually routes through: `_TypedApiClient.param_serialize`
(`sdks/python/src/e2a/v1/client.py`).

`ApiClient.param_serialize` (the generated base) substitutes path parameters
via `quote()` with no dot-segment check and is marked "Do not edit the class
manually" (overwritten by `make generate-sdk`), so the fix overrides
`param_serialize` in `_TypedApiClient`, the hand-written subclass every
`AsyncE2AClient`/`E2AClient` request already constructs its `*Api` objects
against. It checks every path parameter for a bare `.` or `..` before
delegating to the base implementation, matching #915's own proposed fix.

**Scope:** this is the Python half of #915's corrected plan. The TypeScript
half needs a codegen post-processing pass over `generated/http.ts` (a hand
edit there is overwritten by `make generate`), which is a separate, larger
change, not attempted here, so this closes only the Python gap (`Refs #915`,
not `Fixes #915`).

Deleted the client-surface checklist below: this changes no API surface, no
generated types, and no other client. It is a Python-SDK-only hardening fix.

## Test plan

- New `tests/test_generated_layer_dot_segment_guard.py`: constructs the
  generated `AgentsApi` directly against a real `AsyncE2AClient`'s
  `_api_client`, bypassing the ergonomic wrapper. Confirmed against
  unmodified `main`: the request is built and sent (httpx_mock reports no
  matching mock, i.e. a real request would reach the network) rather than
  raising. On this branch it raises `E2AValidationError(code="unsafe_path_segment")`
  before any request is built; an ordinary (non-dot) value still passes
  through unchanged.
- Full suite (`pytest tests/ -v`) and `mypy`, both from this repo's own
  `python-sdk` CI job recipe, in a `python:3.12-slim` container matching that
  job's version: 611 passed, 48 skipped, 94.12% coverage (repo gate is 90%),
  mypy clean.
- Not independently verified: the TypeScript SDK's equivalent gap, since this
  PR does not touch it.

### DIFF
--- a/sdks/python/src/e2a/v1/client.py
+++ b/sdks/python/src/e2a/v1/client.py
@@ -29,7 +29,7 @@ from .generated.api.reviews_api import ReviewsApi
 from .generated.api.templates_api import TemplatesApi
 from .generated.api.contacts_api import ContactsApi
 from .generated.api.webhooks_api import WebhooksApi
-from .generated.api_client import ApiClient
+from .generated.api_client import ApiClient, RequestSerialized
 from .generated.configuration import Configuration
 from .generated.models import (
     AgentView,
@@ -233,6 +233,43 @@ def _assert_not_dot_segment(value: str, param: str) -> str:
 
 class _TypedApiClient(ApiClient):
     """Map malformed successful responses before the retry boundary sees them."""
+
+    def param_serialize(
+        self,
+        method: str,
+        resource_path: str,
+        path_params: Optional[Mapping[str, Any]] = None,
+        query_params: Optional[Any] = None,
+        header_params: Optional[Any] = None,
+        body: Optional[Any] = None,
+        post_params: Optional[Any] = None,
+        files: Optional[Any] = None,
+        auth_settings: Optional[Any] = None,
+        collection_formats: Optional[Any] = None,
+        _host: Optional[str] = None,
+        _request_auth: Optional[Any] = None,
+    ) -> RequestSerialized:
+        # Every generated *Api method builds its URL through this one call, so
+        # checking here covers every path parameter of every resource, not just
+        # the 10 sites the ergonomic wrapper classes above happen to name.
+        if path_params:
+            for name, value in path_params.items():
+                if isinstance(value, str):
+                    _assert_not_dot_segment(value, name)
+        return super().param_serialize(
+            method,
+            resource_path,
+            path_params,
+            query_params,
+            header_params,
+            body,
+            post_params,
+            files,
+            auth_settings,
+            collection_formats,
+            _host,
+            _request_auth,
+        )
 
     def response_deserialize(self, response_data: Any, response_types_map: Any = None) -> Any:
         try:
--- a/sdks/python/tests/test_generated_layer_dot_segment_guard.py
+++ b/sdks/python/tests/test_generated_layer_dot_segment_guard.py
@@ -1,0 +1,49 @@
+"""Regression test for e2a#915: the dot-segment guard must cover the
+generated-layer chokepoint, not just the 10 ergonomic call sites pinned in
+``test_dot_segment_enumeration.py``.
+
+That file already proves every ergonomic method (``client.agents.delete``,
+``client.contacts.delete_outreach``, ...) rejects a dot-segment value before
+any request is built. It cannot prove anything about a caller who imports a
+generated ``*Api`` class directly and calls it against the SDK's own
+``_TypedApiClient`` instance, skipping the ergonomic wrapper entirely,
+which is exactly the gap #915 reports: the per-call-site
+``_assert_not_dot_segment`` checks live in the ergonomic layer, not in the
+one place (``ApiClient.param_serialize``) every generated method routes
+through.
+"""
+
+from __future__ import annotations
+
+import pytest
+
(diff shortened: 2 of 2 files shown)


jiashuoz (approved):

AmirF194:
Thanks for merging this one. Happy to follow up if the dot-segment guard needs adjusting for any other generated-layer path.

PR #931, opened 2026-08-23
fix(emailauth): IDNA-encode domains before DMARC DNS lookups

### LINKED ISSUE #608: fix(emailauth): close remaining RFC 9989 DMARC gaps
## Summary

Track the remaining RFC 9989 conformance gaps in inbound DMARC evaluation identified while reviewing #607.

The current implementation correctly requires at least one aligned, passing SPF or DKIM identifier before returning DMARC `pass`. These gaps are therefore expected to cause false negatives or inaccurate effective-policy metadata, not a known spoofed-mail authentication bypass.

## Confirmed gaps

- Treat an otherwise valid DMARC record without `p=` as `p=none` instead of `permerror`.
- Support effective-policy behavior for `np=` and testing mode `t=y`, or avoid surfacing an effective `policy` value until it is calculated correctly.
- Convert a Unicode RFC 5322 From domain from U-label to A-label before DMARC lookup and alignment.
- For a null SMTP `MAIL FROM`, use the HELO identity for SPF evaluation as required by the SPF/DMARC model.

## Acceptance criteria

- [ ] Missing-`p` and malformed/unsupported tag handling follows RFC 9989 sections 4.7, 4.8, and 4.10.1.
- [ ] The surfaced DMARC policy reflects applicable `p`, `sp`, `np`, and `t` semantics.
- [ ] U-label From domains are converted to A-labels before policy discovery and identifier alignment.
- [ ] Null reverse-path messages can use the HELO identity for SPF-based DMARC alignment.
- [ ] Unit tests cover each case and preserve the existing rule that DMARC passes only with an aligned SPF or DKIM pass.
- [ ] Developer-facing documentation describes the supported DMARC result semantics.

## References

- RFC 9989 §3.2.4: https://www.rfc-editor.org/rfc/rfc9989.html#section-3.2.4
- RFC 9989 §4.7: https://www.rfc-editor.org/rfc/rfc9989.html#section-4.7
- RFC 9989 §4.8: https://www.rfc-editor.org/rfc/rfc9989.html#section-4.8
- RFC 9989 §4.10.1: https://www.rfc-editor.org/rfc/rfc9989.html#section-4.10.1
- RFC 9989 §5.3.1: https://www.rfc-editor.org/rfc/rfc9989.html#section-5.3.1
- Related PR: #607

AmirF194:
## Summary

`normDomain` lowercased and trimmed a domain but never converted it to
its ASCII/punycode form. The DMARC tree walk builds every
`_dmarc.<domain>` TXT lookup from that value (`discoverDMARCRecordWithEvaluator`,
`walkDMARCTree`), so a message with an internationalized From domain
queried the raw Unicode label instead of the label the DNS zone
actually publishes records under, and evaluation silently degraded to
`policy=none`. `internal/identity.normalizeDomain` already does
lowercase + `idna.Lookup.ToASCII` for stored domains; this applies the
same pattern to the inbound auth path, with the same fallback to the
unmodified string on an encoding error.

## Operational risk

None outside the emailauth package. `normDomain` also feeds the
displayed SPF/DKIM domain fields; those are already-ASCII values in
practice (SMTP HELO and DKIM `d=` tags), and the full existing suite
for this package (including those cases) is unchanged by this diff.

## Test plan

- [x] `TestEvaluateDMARCEncodesUnicodeFromDomainForLookup` (new):
      fails on `main` (queries `_dmarc.münchen.example`, the record
      under the punycode name is never found), passes on this branch.
- [x] `go test ./internal/emailauth/...`: full package suite green.
- [x] `go build ./...` and `go test ./...`: unaffected elsewhere.
- Not checked: real DNS resolution against a live internationalized
  domain (`fakeTXTResolver` only); the query-name construction is what
  this fix changes and what the test asserts.

Fixes #608

### DIFF
--- a/internal/emailauth/checker.go
+++ b/internal/emailauth/checker.go
@@ -11,6 +11,7 @@ import (
 
 	"blitiri.com.ar/go/spf"
 	"github.com/emersion/go-msgauth/dkim"
+	"golang.org/x/net/idna"
 )
 
 var defaultDMARCEvaluator = newDMARCEvaluator(netTXTResolver{resolver: net.DefaultResolver})
@@ -120,8 +121,15 @@ func aggregateLegacyDKIM(results []DKIMResult) CheckResult {
 	return CheckResult{Status: results[0].Status, Detail: results[0].Detail}
 }
 
+// normDomain lowercases and IDNA-normalizes a domain (same pattern as
+// internal/identity.normalizeDomain), so the "_dmarc."+domain DNS queries
+// below use the label the resolver actually serves records under.
 func normDomain(d string) string {
-	return strings.TrimSuffix(strings.ToLower(strings.TrimSpace(d)), ".")
+	d = strings.TrimSuffix(strings.ToLower(strings.TrimSpace(d)), ".")
+	if ascii, err := idna.Lookup.ToASCII(d); err == nil {
+		return ascii
+	}
+	return d
 }
 
 type AuthorIdentity struct {
--- a/internal/emailauth/dmarc_test.go
+++ b/internal/emailauth/dmarc_test.go
@@ -42,6 +42,23 @@ func TestEvaluateDMARCAlignedDKIM(t *testing.T) {
 	}
 }
 
+func TestEvaluateDMARCEncodesUnicodeFromDomainForLookup(t *testing.T) {
+	r := &fakeTXTResolver{records: map[string][]string{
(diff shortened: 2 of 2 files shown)


jiashuoz (approved):
Thanks for your contribution!!

AmirF194:
Thanks for the quick review and merge. Glad the IDNA fix lines up cleanly with the pattern already in identity/store.go.

PR #938, opened 2026-08-25
fix(contract): raise on an unrecognized setup step

### LINKED ISSUE #823: tests/contract: scenarios.yaml has no strict decode — a typo'd assertion key is silently dropped by all three runners

AmirF194:
## Summary

Issue #823 covers two gaps in the contract-test harness's permissive decoding: a strict-decode gap in the Go loader (fixed by #910) and this one, the TS/Python contract runners' setup-step dispatch.

`Runner.execute_setup` (Python, `sdks/python/tests/test_contract.py`) and `ScenarioRunner.executeSetup` (TypeScript, `sdks/typescript/test/v1/contract.test.ts`) each dispatch a setup step through a chain of checks for the four known primitives (`register_domain`, `register_agent`, `verify_domain`, `inject_message`) with no fallback branch. A step whose only key matches none of the four falls through the whole chain silently: no HTTP call, no exception, and the scenario still reports PASS with its precondition missing. `execute_steps`/`executeSteps` (the STEPS dispatch, a separate method) already raise/throw on an unrecognized `action`; this brings the setup dispatch to the same shape.

## Root cause

The setup dispatch chain never had an else/default branch, so an unrecognized key (a typo, or a new primitive added to `scenarios.yaml` before a runner implements it) degrades to "scenario runs without its precondition" instead of an error.

## Fix

Python's chain becomes `if/elif/else raise ValueError`. TypeScript keeps its existing `if (...) { ...; continue; }` branches and adds a `throw Error` after the last one. Checked every scenario in `tests/contract/scenarios.yaml`: `register_agent`, `register_domain`, `verify_domain`, `inject_message` are the only four setup keys used anywhere in the file, so no existing scenario reaches the new branch.

## Test plan

- Python: `pytest tests/ -v` (3.12, clean container): 613 passed, 48 skipped (need `E2A_TEST_BASE_URL`/`E2A_TEST_API_KEY`, a live contract server), coverage 94.12%. The new `test_execute_setup_rejects_unrecognized_key` fails on main (`DID NOT RAISE`) and passes on this branch; `test_execute_setup_registers_domain_then_agent` pins the known-key dispatch is unchanged.
- Python: `mypy` clean.
- TypeScript: `tsc --project tsconfig.test.json` clean (clean container).
- Not run: the live-server `python-contract`/`ts-contract` jobs (Postgres + a built Go contract server), so the TypeScript-side fix is verified by type-check and by mirroring the Python dispatch exactly, not executed against a live server.

Fixes #823

### DIFF
--- a/sdks/python/tests/test_contract.py
+++ b/sdks/python/tests/test_contract.py
@@ -257,14 +257,15 @@ class Runner:
                 resp = self._raw("POST", "/v1/domains", {"domain": domain})
                 if resp.status_code >= 400 and resp.status_code != 409:
                     resp.raise_for_status()
-
-            if "register_agent" in s:
+            elif "register_agent" in s:
                 agent = s["register_agent"]
                 email = self.resolve(agent["email"])
                 resp = self._raw("POST", "/v1/agents", {"email": email})
                 if resp.status_code >= 400 and resp.status_code != 409:
                     resp.raise_for_status()
                 self.vars["agent_email"] = email
+            else:
+                raise ValueError(f"unknown setup step: {s!r}")
 
         return False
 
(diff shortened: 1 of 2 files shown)


jiashuoz (approved):

AmirF194:

PR #942, opened 2026-08-27
fix(identity): close the check-then-act race on max_agents

### LINKED ISSUE #940: limits: max_agents is not enforced under concurrency (check-then-act)
## What

`POST /v1/agents` enforces `max_agents` with an unsynchronized check-then-act: `EnforceAgentCreate` counts rows (`internal/limits/enforcer.go:181-199`), then `handleCreateAgent` calls `CreateAgent` to insert (`internal/httpapi/agents_write.go:374-388`, backed by `createAgent` in `internal/identity/store.go:2291`). No transaction spans the two and no lock serializes them, so concurrent requests can all read the same pre-insert count and all pass.

#822 (fixed for `max_domains` in #901) named this exact path as a likely follow-up: "The same shape almost certainly applies to `EnforceAgentCreate` ... worth checking as part of any fix, though agent creation at least sits behind the rate limiter." It was closed with only the domains side fixed.

## Evidence

Reproduced against a real Postgres 16 container at HEAD (`e1f8b1839b3817029bfc3b43f606f91582ff69f4`): an account with `max_agents: 1` issuing 8 concurrent `POST /v1/agents` requests, each for a distinct new address, all 8 succeeded (0 rejected). `agent_identities` ended up with 8 rows for that account.

The per-operation rate limiter does apply to `createAgent` (`internal/httpapi/ratelimit.go:137`), unlike `max_domains` at the time #822 was filed, so the overshoot is bounded by the limiter's window rather than unbounded client parallelism. It still lets an account exceed its plan cap within that window.

## Why it matters

`max_agents` is a billing boundary, same as `max_domains`. An account can exceed it with ordinary client-side parallelism, no special tooling needed.

## A second, separate gap

The OAuth auto-create flow (`issueOAuthCodeWithNewAgent` in `internal/agent/oauth_handlers.go`) creates an agent via `CreateAgentTx` directly and never calls `EnforceAgentCreate` at all, race or no race. Noting it here since it's the same invariant, but it's a different mechanism (no check at all, vs. an unsynchronized one) and probably wants its own fix.

## Not covered by tests today

Nothing exercises concurrent `max_agents` enforcement in any suite, the same gap #822 noted for `max_domains` before #901.

AmirF194:
## Summary

`EnforceAgentCreate` (the `max_agents` pre-check) and `CreateAgent` (the insert) ran as two independent operations with nothing serializing them: concurrent `POST /v1/agents` requests could all read the same pre-insert count and all pass. Reported with a live reproduction against real Postgres: 8 concurrent requests against a cap of 1, all 8 succeeded.

`CreateAgentWithLimit` now takes a per-user advisory lock (keyspace 2, distinct from `claimOrCreateDomain`'s keyspace 1) and re-checks the count inside the same transaction as the `INSERT`, mirroring #901's fix for `max_domains`. `handleCreateAgent` calls it directly instead of `EnforceAgentCreate` + `CreateAgent`. `EnforceAgentCreate` stays wired, called only on the limit-exceeded path, to attach the plan/upgrade-URL details a bare `AgentLimitExceededError` doesn't carry; if it's unset or doesn't corroborate, the endpoint still answers 402 from the bare error.

## Operational risk

No schema change, no behavior change on the non-concurrent path or in the 402 response shape (still `limit_exceeded` / resource `agents`, same fields). A request that previously slipped past the cap under a race now gets `402 limit_exceeded`, the same response the endpoint already returns for a non-racy over-cap request.

`internal/agent/oauth_handlers.go`'s `issueOAuthCodeWithNewAgent` (the OAuth auto-provisioning path) also creates an agent, via `CreateAgentTx`, and does not check `max_agents` at all, before or after this change. Left alone here (different mechanism, and I don't have context on the intended cap behavior for auto-provisioning); flagged in the issue.

## Client surface checklist

Not applicable: internal concurrency fix, the request/response shape of `POST /v1/agents` is unchanged.

## Test plan

- New e2e test (`internal/e2e/agents_create_race_e2e_test.go`, `-tags integration`): 8 concurrent `POST /v1/agents` against `max_agents=1`. Fails on `main` (8/8 succeed, cap blown 8x), passes on this branch (1 created, 7 `402 limit_exceeded`), both against a real Postgres 16 container.
- `go test ./internal/identity/...` and `./internal/httpapi/...` green, including two new unit tests for the `GetLimits`-failure and enforcer-unwired error-mapping paths.
- `go test ./...` (the `Go tests` job's own command) green.
- `internal/httpapi` coverage 87.0%, above the package's 73% floor.
- Not verified: contention beyond 8 concurrent requests, or across more than one connection pool / process.

Fixes #940

### DIFF
--- a/internal/apiserver/apiserver.go
+++ b/internal/apiserver/apiserver.go
@@ -157,6 +157,7 @@ func BuildDeps(p Params) httpapi.Deps {
 		GetConversation:   p.Store.GetConversationByID,
 
 		CreateAgent:          p.Store.CreateAgent,
+		CreateAgentWithLimit: p.Store.CreateAgentWithLimit,
 		LookupDomain:         p.Store.LookupDomain,
 		LookupCoveringDomain: p.Store.LookupCoveringDomain,
 		// Exact-domain MX probe for the create-time two-way-inbox gate.
--- a/internal/e2e/agents_create_race_e2e_test.go
+++ b/internal/e2e/agents_create_race_e2e_test.go
@@ -1,0 +1,101 @@
+//go:build integration
+
+package e2e_test
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"net/http"
+	"sync"
+	"testing"
+
+	"github.com/tokencanopy/e2a/internal/limits"
+	"github.com/tokencanopy/e2a/internal/testutil"
+)
+
+// max_agents must be enforced under concurrency: N simultaneous POST
+// /v1/agents for distinct new addresses on one account, capped at 1, must
+// leave exactly one agent created (same class as #822/#901's max_domains
+// race). Before the fix, CheckAgentCreate's count read and CreateAgent's
+// insert were two independent DB calls with nothing serializing them, so
+// every concurrent request could read the same pre-insert count and all pass
+// the cap.
+func TestCreateAgentConcurrentRequestsRespectMaxAgentsE2E(t *testing.T) {
+	pool := testutil.TestDB(t)
+	ts := testutil.TestServer(t, pool)
+	ctx := context.Background()
+
+	user, err := ts.Store.CreateOrGetUser(ctx, "agent-race-owner@example.com", "Agent Race Owner", "google-agent-race-owner")
+	if err != nil {
+		t.Fatalf("CreateOrGetUser: %v", err)
+	}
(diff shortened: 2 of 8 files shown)


jiashuoz (approved):

AmirF194:
Thanks for reviewing and merging this one.

PR #944, opened 2026-08-27
fix(webhook): thread scheduled_at into the review-hold event payloads

### LINKED ISSUE #836: webhook events don't carry scheduled_at for held/scheduled sends (#815 follow-up)
## What

Since #833 ("preserve send_at across a review hold"), a held message that carried a future `send_at` is re-armed on approval: it returns `status=scheduled` with `scheduled_at` on the HTTP surfaces (send response, approve response, review queue, message detail) and actually submits at that instant.

But the async webhook events are not extended:

- `email.review_requested` (`buildPendingApprovalEvent` in `internal/agent/webhooks_api.go`) — no `scheduled_at`, even though a future `send_at` was stamped on the held draft.
- `email.review_approved` (`buildApprovedEvent`) — no `scheduled_at`, so a subscriber cannot tell whether approval re-armed the schedule or is sending immediately.

The events themselves fire correctly and on time (`review_approved` at approval, `email.sent` at the scheduled instant) — this is purely a payload-surface gap.

## Why it is worth closing

A reviewer that drives review asynchronously via webhooks (an automated approver, a queue consumer, a compliance workflow) receives `email.review_approved` with no way to know the message was deferred to `scheduled_at` vs. delivered now. Today they would have to go back and `GET /v1/messages/{id}` to discover it. Both features are beta, so this is an additive (optional field) change with no compatibility cost.

## Fix

Add optional `scheduled_at` (RFC 3339, `x-stability-level: beta`) to the `email.review_requested` and `email.review_approved` event payloads when the message carries a schedule:

1. `buildPendingApprovalEvent`: include `msg.ScheduledAt` (the row already has it after the hold stamps it).
2. `buildApprovedEvent`: include `sent.ScheduledAt` — present when approval re-armed the schedule; note it stays set on the row after the instant passes, so document "set when the message was scheduled; retained afterwards".
3. Regenerate / update the OpenAPI + generated SDK event models, and cover with a webhook-payload test (e.g. in `internal/agent/webhooks_events_test.go`).

Follow-up to #815 / #833.

AmirF194:
## Summary

`buildPendingApprovalEvent` and `buildApprovedEvent` (`internal/agent/webhooks_api.go`) never read `msg.ScheduledAt` / `sent.ScheduledAt`, so a webhook-only subscriber of `email.review_requested` / `email.review_approved` cannot tell a deferred send from an immediate one without a follow-up `GET` on the message. Adds `scheduled_at` to both payloads when the row carries a schedule, the same optional-field pattern the same builders already use for `provider_message_id` and `lifecycle_transitions`.

On the issue's third fix step (regenerate the OpenAPI + generated SDK event models): `email.review_requested` and `email.review_approved` are deliberately kept as `map[string]any` at their trigger sites and are not registered as OpenAPI component schemas (`internal/eventpayload/payloads.go`'s package doc: these two stay untyped "until their shape settles"). There is no generated model or golden fixture for either event, so there is nothing to regenerate; `make generate-sdk-check` and the OpenAPI contract gates don't cover this payload.

## Client surface checklist

Not applicable: this is an additive field on an already-untyped, beta webhook payload, not a typed API or client-SDK surface (see note above).

## Test plan

- `TestBuildPendingApprovalEvent_CarriesScheduledAt` / `TestBuildApprovedEvent_CarriesScheduledAt`: fail on `main`, pass on this branch.
- `TestBuildPendingApprovalEvent_OmitsScheduledAtForImmediateSend` / `TestBuildApprovedEvent_OmitsScheduledAtForImmediateSend`: cover the unset case.
- `go test ./...` (full suite, go1.26 per CI) and `make fmt-check` both pass.
- Did not check: the coverage-gate job (`make cover`) needs Postgres and doesn't gate `internal/agent`, so I didn't run it.

Fixes #836

### DIFF
--- a/internal/agent/webhooks_api.go
+++ b/internal/agent/webhooks_api.go
@@ -97,6 +97,9 @@ func (a *API) buildPendingApprovalEvent(
 		"conversation_id":     req.ConversationID,
 		"approval_expires_at": msg.ApprovalExpiresAt,
 	}
+	if msg.ScheduledAt != nil {
+		data["scheduled_at"] = msg.ScheduledAt
+	}
 	attachReviewLifecycle(data, msg.LifecycleTransitions)
 	return webhookpub.Event{
 		ID:             generateEventIDForAgent(),
@@ -130,6 +133,9 @@ func (a *API) buildApprovedEvent(
 	}
 	if sent.ProviderMessageID != "" && sent.Method != "loopback" {
 		data["provider_message_id"] = sent.ProviderMessageID
+	}
+	if sent.ScheduledAt != nil {
+		data["scheduled_at"] = sent.ScheduledAt
 	}
 	attachReviewLifecycle(data, sent.LifecycleTransitions)
 	return webhookpub.Event{
--- a/internal/agent/webhooks_events_test.go
+++ b/internal/agent/webhooks_events_test.go
@@ -219,6 +219,31 @@ func TestBuildPendingApprovalEvent(t *testing.T) {
 	}
 }
 
+func TestBuildPendingApprovalEvent_CarriesScheduledAt(t *testing.T) {
+	a := &API{}
+	agent := &identity.AgentIdentity{ID: "bot@x.example.com", UserID: "u_1"}
+	future := time.Now().Add(2 * time.Hour)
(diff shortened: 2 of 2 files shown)


jiashuoz (approved):

AmirF194:

Issue #815, opened 2026-08-03
feat(api): preserve send_at across a review hold instead of discarding it


jiashuoz:
## Problem

A scheduled send caught by an HITL review hold loses its schedule entirely. On approval it sends **immediately**, not at the requested time. `send_at` is discarded at accept time and never persisted on the held message.

Documented as intended (`api/openapi.yaml`, on send/reply/forward):

> Scheduling does NOT survive a review hold: if held, `send_at` is dropped and the message sends on approval (the hold takes precedence over the loopback check).

## Observed behavior

Sent from an agent whose outbound gate is `policy=allowlist, allowlist=[], action=review` (the documented "require human review for every outbound email" configuration), with `send_at` fifteen minutes in the future.

Response — `pending_review`, and no `scheduled_at`:

```json
{
  "message_id": "msg_…",
  "status": "pending_review",
  "approval_expires_at": "…"
}
```

The stored message has **no `scheduled_at` field at all**. For contrast, the same send from an agent with an open outbound gate returns `status: scheduled` and the message carries `"scheduled_at": "…"` through to the record, firing 1.1s after the requested instant.

So the schedule is not stored-and-ignored — it is dropped at accept. It cannot be recovered later even if the approval path wanted to honor it.

The API also accepts `send_at` **silently**: no warning, no echoed field, no error. A caller has no way to learn their schedule was discarded except by reading the field docs or noticing the mail arrived early.

## Why this is wrong

`send_at` and the review gate express different things. `send_at` is the caller's business intent — deliver at a particular moment, in the recipient's morning, at the start of a campaign window. The review gate is authorization — a human confirming the message may go out at all. Approving a message answers "may this send," not "send it now."

Collapsing them means a reviewer's click silently overrides the sender's timing:

1. An agent schedules a message for Tuesday 09:00 for a recipient in another timezone.
2. The outbound gate holds it for review.
3. A human approves Monday at 16:00.
4. The message goes out Monday at 16:00 — many hours early, with no warning at any point.

It also makes the two features mutually exclusive in practice. Any agent configured for review-everything cannot use scheduled sending at all. Both are advertised beta capabilities and they do not compose.

## Proposed behavior

Persist `send_at` through the hold, and re-arm it on approval:

- **`send_at` still in the future at approval** → return to `scheduled` and submit at `send_at`.
- **`send_at` already passed** → submit immediately. "Not before" is already the documented contract, and approval was the only remaining blocker.

[...]

PR #974, opened 2026-08-30
feat(api): scheduled-send queue (/v1/scheduled) + dashboard Scheduled tab

cyj-git-0825:
## Requirement
Operators need a single, account-wide view of **what's going to send later**.
Scheduled sending (`send_at`, #815 / migration 084) is modeled and shown *per
message* — `scheduled_at` rides on message/review views and renders as a chip —
but there is no way to ask the server for the scheduled set as a whole. The
dashboard's Pending page is driven by `GET /v1/reviews`, which returns only
`pending_review` holds; a scheduled-but-not-held message is `delivery_status =
accepted`, so it never appears there and was invisible in the dashboard.

## Solution
Add a first-class, account-scoped **`GET /v1/scheduled`** endpoint: outbound
messages that are `accepted` with a future `scheduled_at`, soonest-first,
keyset-paginated, read-only. Surface it as a **"Scheduled" tab** on the Pending
page; clicking a row lazy-loads the full message (recipients + body) read-only —
no approve/reject, because a scheduled send is not a hold. The endpoint reuses
the `/v1/reviews` account-scoped pattern (cursor infra, scope guard, page shape)
and is marked **beta**.

## Approach & the main tradeoff (why a new endpoint, not a filter)
Two options were considered:

1. **Add a `scheduled` filter to the existing message list**
   (`GET /v1/agents/{email}/messages`, or its AIP-160 filter registry).
2. **A dedicated account-scoped `GET /v1/scheduled`** (chosen).

The existing message list is **agent-scoped by its path** — it lists one inbox.
The view needed here is **account-scoped** (every inbox at once), ordered by send
time. Reusing the per-agent list would force the client to fan out over every
agent and merge/paginate in the browser, with no way to order or bound by
`scheduled_at` server-side. There is no account-scoped message list to hang a
filter on, so option (1) would require inventing one anyway.

`GET /v1/reviews` set the precedent: a dedicated account-scoped endpoint (not a
filter on `/messages`) for exactly this operator-wide view. `/v1/scheduled` is
its direct analog. Being beta, it can later fold into a general account-scoped
filterable message list without a breaking change.

At the store layer, `ListReviews` couldn't be reused — its query hardcodes
`WHERE status = 'pending_review'`. `ListScheduled` is a sibling query
(`scheduled_at IS NOT NULL AND scheduled_at > now() AND delivery_status =
'accepted'`, `ORDER BY scheduled_at ASC`), keeping the two sets disjoint by
construction rather than by a runtime flag.

## Impact on other components
Purely additive — nothing existing changes behavior:
- No new migration (reads the existing `messages.scheduled_at` from #815).
- No change to any existing endpoint, and **no change to when/whether a message
  sends** — this is a read projection; it never stamps `scheduled_at`, enqueues
  a job, or cancels one. The send/schedule path (`internal/outboundsend`,
  `internal/httpapi/outbound.go`) is untouched.
- Shared-file edits are one-liners: a cursor constant, a deps field, one wiring
  line, one route registration.
- Dashboard: default tab is still "Held", so existing deep links (`?id=`) and
  the sidebar Pending badge (holds-only count) are unchanged. Held-and-scheduled
  drafts still show under Held with their existing chip; once approved they move
  to Scheduled — no duplication.

## From a user's perspective (verified in a running stack)
Ran the server (this branch) + dashboard + Postgres locally, created an agent,
sent two messages with a future `send_at`:
- `POST …/messages` with `send_at` → `202 {status: scheduled}`.
- Pending page shows a **Held / Scheduled** tab switch; the Scheduled tab lists
  both, soonest-first, each with a "Sends <local time>" chip and
  `inbox → recipient`.
- Clicking a row expands the full email body read-only (no approve/reject).
- `GET /v1/scheduled` returns them ordered by send time; once a message sends or
  is trashed it drops off.

## Testing
- **Go**: unit (scope guard, pagination, not-implemented) + DB (`ListScheduled`
  returns only accepted+future, soonest-first, account-scoped; excludes past /
  unscheduled / already-sent) + the shared **contract scenario proven live by
  the Go runner**.
- **TS SDK**: build + 262 unit + contract-coverage audit (`scheduled.list`
  exercised).
- **Python SDK**: 235 unit + **mypy** clean (async + sync).
- **MCP**: 332 tests (new `list_scheduled` tool, tier map, freeze list).
- **Web**: tsc + 44 Jest (tab render, empty state, click-to-expand read-only) +
  eslint.
- **Gate**: SDK operation coverage 75/75.
- CI-only live jobs (`ts-contract` / `python-contract`) consume the same
  `scenarios.yaml` the Go runner already passed.

## Client surface checklist
- [x] Go handler + tests / no migration (reuses `messages.scheduled_at`)
- [x] OpenAPI spec + generated TS/Python types
- [x] TS SDK (`client.scheduled.list`) + Python SDK (async+sync)
- [x] MCP tool (`list_scheduled`)
- [ ] CLI — N/A (the sibling review queue has no CLI command; kept consistent)
- [x] Web dashboard — Pending "Scheduled" tab
- [x] Contract scenario in `scenarios.yaml`, proven via the Go runner


jiashuoz:
Create a fresh branch/PR off latest main?

cyj-git-0825:
Reopened as a fresh branch/PR off latest `main` per review: #977 (branch `feat/v1-scheduled-queue`, cut from current `main` HEAD, same content — identical tree). Closing this one in favor of #977.

PR #1016, opened 2026-09-11
test(contract): point shared test domain at agents.localhost

### LINKED ISSUE #829: testutil: contract server uses agents.e2a.dev as its shared domain, against the public data boundary
## What

`internal/testutil/contract_server.go` configures `SharedDomain: "agents.e2a.dev"` (lines 96 and 115 — `agent.NewAPI(...)` and `apiserver.Params`). Agent creation without domain verification only works on the shared domain, so every scenario that needs an agent writes an `@agents.e2a.dev` address into `tests/contract/scenarios.yaml`. There are currently five such scenarios.

AGENTS.md says verbatim:

> An address on `agents.e2a.dev` is a customer identifier, not a safe example value.

The committed values are token-randomized and not production-derived, so nothing here leaks real data. But the file is public, the rule is stated without exception, and the shape means every future agent scenario adds another one.

## Fix

Point the contract server's `SharedDomain` at `agents.localhost` (already on AGENTS.md's approved list), then update the affected scenario addresses. Contained to test infrastructure — no production configuration reads this.

Worth checking `internal/testutil/server.go` and the prober seed for the same constant while in there.

AmirF194:
Swapped `agents.e2a.dev` for `agents.localhost` as the contract harness's SharedDomain in `contract_server.go` and `server.go`, and updated the 17 scenario addresses in `scenarios.yaml` that depend on it. Also caught one more literal in `contract_server.go`'s over-cap fixture (`overcap-bot-%d@agents.e2a.dev`) that uses the same domain outside the SharedDomain wiring.

Checked `server.go` and the prober seed like the issue asked. `server.go` only had the same two SharedDomain lines, already covered above. The prober's agent email comes from `E2A_PROBE_AGENT_EMAIL` with no hardcoded domain, so nothing there needed a change.

Added a test that reads the three files and fails if `agents.e2a.dev` shows up again. It fails without the change.

Fixes #829

### DIFF
--- a/internal/e2e/agents_create_race_e2e_test.go
+++ b/internal/e2e/agents_create_race_e2e_test.go
@@ -50,9 +50,9 @@ func TestCreateAgentConcurrentRequestsRespectMaxAgentsE2E(t *testing.T) {
 		go func(i int) {
 			defer wg.Done()
 			// A shared-domain address (testutil's default SharedDomain,
-			// agents.e2a.dev) needs no domain ownership/verification, so the
+			// agents.localhost) needs no domain ownership/verification, so the
 			// only thing under test is the max_agents race.
-			body := []byte(fmt.Sprintf(`{"email":"race-agent-%d@agents.e2a.dev","name":"race agent"}`, i))
+			body := []byte(fmt.Sprintf(`{"email":"race-agent-%d@agents.localhost","name":"race agent"}`, i))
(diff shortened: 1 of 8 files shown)


AmirF194:
CI's clean at head 105d0ae7 (28/28 legs) and this has sat a week with no review. Let me know if anything's blocking a look.

